PR #9, opened 2020-08-29
Update PowerVirtualServers::CloudManager descripton

jaywcarman:
Match the Cloud Manager description to the product website:
https://www.ibm.com/cloud/power-virtual-server

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
@@ -55,6 +55,6 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
   end
 
   def self.description
-    @description ||= "IBM Cloud".freeze
+    @description ||= "IBM Power Systems Virtual Servers".freeze
   end
 end


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -55,6 +55,6 @@ def self.ems_type
   end
 
   def self.description
-    @description ||= "IBM Cloud".freeze
+    @description ||= "IBM Power Systems Virtual Servers".freeze
```
jaywcarman, line 58: This is a little bit confusing since it is different from our namespace `IbmCloud::PowerVirtualServers`. I think we need to match the branding exactly, and adding 'Systems' to the namespace would make it even longer.

agrare: Yeah the description doesn't have to match the namespace exactly (branding tends to change more often than the backend :wink:) so I think this is fine

agrare: Can you also put a PR in to https://github.com/ManageIQ/manageiq/blob/master/spec/models/ext_management_system_spec.rb#L69 to match?

jaywcarman: https://github.com/ManageIQ/manageiq/pull/20499

agrare:
Just noticed this branch is on the main repo not a fork, double check your local clone's remotes to make sure you're not pushing branches to the main ManageIQ repo

jaywcarman:
> Just noticed this branch is on the main repo not a fork, double check your local clone's remotes to make sure you're not pushing branches to the main ManageIQ repo

Oops! Looks like I pasted the wrong url when I switched to the public address. Fixed it (and added no_push to my upstream remote). Thanks @agrare

PR #11, opened 2020-09-01
Use "ibm" image_name for all managers

jaywcarman:
All of the IBM Cloud managers use the same icon image. In our recently
submitted PR to manageiq-decorators
(https://github.com/ManageIQ/manageiq-decorators/pull/31) the
CloudManager symlink name isn't correct.

A more straightforward approach is to set each manager's 'image_name'
method to return common image name: "ibm".


iv10 (approved):

iv1111 (approved):

iv1111:
As discussed in Slack. My commit to the decorators repository will follow now.

PR #13, opened 2020-09-01
Fix volume size

jaywcarman:
The Power Cloud API returns volume size in units of gigabytes but
ManageIQ expects bytes.


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -120,7 +120,7 @@ def volumes
         :creation_time     => vol['creationDate'],
         :description       => 'IBM Cloud Block-Storage Volume',
         :volume_type       => vol['diskType'],
-        :size              => vol['size'],
+        :size              => vol['size'] * (1024 ** 3),
```
agrare, line 123: You can do `vol['size']&.gigabytes`

PR #14, opened 2020-09-02
Fix instance hardware disk values

jaywcarman:
The instance hardware disk values can be set to "real" values using
'collector.volume' calls.


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -44,12 +44,12 @@ def instances
       instance["volumeIDs"].each do |vol_id|
         persister.disks.build(
```
jaywcarman, line 45: @agrare The disks we're populating here are the same block storage volumes added to our StorageManager. Is this for some other type of disk, or is it meant as a quick informational reference?

iv1111:
There is another thing that should be fixed regarding the volumes. Volume size is currently propagated to the UI as 'bytes'. We should multiply the value by Numeric::GIGABYTE to get Gigabytes there as that is what our API gives us back.

`:size => vol['size'].to_i * Numeric::GIGABYTE,`

or similar, not sure if "vol['size']" is already integer.

jaywcarman:
WIP - I need to record a new refresher VCR to account for the additional volume calls.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
agrare: Yeah we should be linking up the VM to the cloud_volumes rather than creating new disk records

iv1111: `collector.volume(vol_id)["size"]&.gigabytes`
One question. Here 'size' would evaluate to 'nil' if  "collector.volume(vol_id)["size"]" has the "nil" value, right? Will MIQ handle that?

agrare: Yes, generally we treat `nil` as not available and only limited cases absolutely require the value to be present.  Most of those cases have model-level constraints validating the presence of the column.

jaywcarman:
> There is another thing that should be fixed regarding the volumes. Volume size is currently propagated to the UI as 'bytes'. We should multiply the value by Numeric::GIGABYTE to get Gigabytes there as that is what our API gives us back.

@iv1111  6ac0105 was recently merged to to fix that one.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -44,12 +44,12 @@ def instances
...
           :hardware        => ps_hw,
-          :device_name     => vol_id,
-          :device_type     => "block",
+          :device_name     => collector.volume(vol_id)["name"],
```
agrare, line 47: If we are not able to find the volume by id for whatever reason (volume was deleted after we collected the VM) this will fail trying to dereference `nil["name"]` and the entire refresh will fail.

Also `collector.volume()` doesn't cache the result so this looks like it is actually making 3 API calls.

Would recommend doing `volume = collector.volume(vol_id)` at the top then you can do `name, disk_type, size = volume&.values_at("name", "diskType", "size")`

PR #15, opened 2020-09-02
Add an IBM Cloud Provider

agrare:
Add an IBM Cloud Provider class which can group many child cloud_managers (power virtual servers and others) with the same credentials

TODO:
- [ ] Only require API Key if not selecting an existing provider


app/models/manageiq/providers/ibm_cloud/provider.rb
```diff
@@ -0,0 +1,11 @@
+class ManageIQ::Providers::IbmCloud::Provider < ::Provider
```
jaywcarman, line 1: We'd like to define a common "IBM Cloud" provider registration form, with a drop down for resource selection (initially PowerVS or VPC). I made put together an prototype DDF [here](https://github.com/jaywcarman/manageiq-providers-ibm_cloud/blob/ibm_cloud_registration/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L32-L105).

Can we do this through a ` class ManageIQ::Providers::IbmCloud::CloudManager` that defines the registration form such that the type field selection sets class used for the actual CloudManager instantiate?

agrare: We can just delegate params_for_create to the provider class instead of creating a new unused cloud_manager.

That's awesome though I'll definitely be stealing that DDF form :smile:

blancett (approved):
This looks great - we would like to get it merged start playing with it.

jaywcarman (approved):

agrare:
Still WIP because the default credentials are shown as None on the UI once added so I need to delegate one or two more methods to the provider.  I'll take it out of WIP once I get that fixed up though.

agrare:
Alright I moved what would be shared by VirtualServers into the Provider class, taking out of WIP

app/models/manageiq/providers/ibm_cloud/provider.rb
```diff
@@ -0,0 +1,72 @@
...
+          :component  => "password-field",
+          :name       => "authentications.default.auth_key",
+          :id         => "authentications.default.auth_key",
+          :label      => _("IBM Cloud API Key (if not using an existing provider)"),
+          :type       => "password",
+          :isRequired => true,
+          :condition  => {
+            :when    => "provider_id",
+            :isEmpty => true
+          },
```
agrare, line 34: Hey @skateman this condition worked great but I realized we can't edit the authentication with this hidden after creation then.

Can we change the "required" to be contingent on if the provider is selected or not?

skateman: You can add the field twice with mutually exclusive conditions and having/not having the validation...

agrare: Hm this didn't work, they both show up and I get an error about duplicate keys

agrare:
```
                      {
                        :component  => "password-field",
                        :name       => "authentications.default.auth_key",
                        :id         => "authentications.default.auth_key",
                        :label      => _("IBM Cloud API Key (Required)"),
                        :type       => "password",
                        :isRequired => true,
                        :condition  => {
                          :when    => "provider_id",
                          :isEmpty => true
                        },
                        :validate   => [{:type => "required"}]
                      },
                      {
                        :component  => "password-field",
                        :name       => "authentications.default.auth_key",
                        :id         => "authentications.default.auth_key-optional",
                        :label      => _("IBM Cloud API Key (Optional)"),
                        :type       => "password",
                        :isRequired => false,
                        :condition  => {
                          :when    => "provider_id",
                          :isEmpty => false
                        }
                      }
```

skateman: Yeah, we might need a custom validator, that only gets triggered if a condition is set. I'm not sure what should I do with the `*` marking yet but I'll figure that out too.

agrare: Okay for now I'll just leave the authentications.default.auth_key required

skateman: So it seems you can't have the field with the same name twice :confused: and we cannot do `resolveProps` as we send the schema as pure JSON. I will make a decorator around all the validators and enhance it by allowing a condition in the validator's body. This way the validation would only run if the condition is met. I actually had this as a proposed feature in the upstream DDF but it has been removed from v2 :disappointed: 

The `isRequired` is a little bit trickier, I have a few options, but all of them seem crazily hacky for now. I will keep thinking while I'm implementing the conditional validators. In the meantime I have to ask :trollface: is it necessary to hide the `*` from the field?

skateman: https://github.com/data-driven-forms/react-forms/issues/802

skateman:
Something like this should help you. For now the `skipSubmit` is causing a warning as it is an illegal prop, but it works properly. I'm gonna figure out how to suppress that warning.
```diff
diff --git a/app/models/manageiq/providers/ibm_cloud/provider.rb b/app/models/manageiq/providers/ibm_cloud/provider.rb
index fd7c99c..3657619 100644
--- a/app/models/manageiq/providers/ibm_cloud/provider.rb
+++ b/app/models/manageiq/providers/ibm_cloud/provider.rb
@@ -21,11 +21,33 @@ class ManageIQ::Providers::IbmCloud::Provider < ::Provider
             }
           end
         },
+        { # If this field has a value, it means that we are editing an existing provider
+          :component  => 'text-field',
+          :type       => 'hidden',
+          :name       => 'id',
+          :skipSubmit => true, # we don't want to submit this field
+        },
         {
           :component => 'sub-form',
           :id        => 'endpoints-subform',
           :name      => 'endpoints-subform',
           :title     => _("Endpoints"),
+          :condition => { # show this field only if
+            :or => [
+              { # we are editing an existing provider
+                :when       => 'id',
+                :isNotEmpty => true,
+              },
+              { # we are adding a new proivider AND the provider_id is unset
+                :and => [
+                  {
+                    :when    => 'provider_id',
+                    :isEmpty => true,
+                  }
+                ]
+              }
+            ],
+          },
           :fields    => [
             :component => 'tabs',
             :name      => 'tabs',
```

iv1111 (approved):

chessbyte:
@agrare is this still an active PR?

agrare:
Still something we should do but this is old enough it probably isn't helpful to build off of

PR #17, opened 2020-09-03
Cleanup Template#provider_object method

chessbyte:
`connection` variable was being passed in, but never used

Found via [CodeClimate Issutes](https://codeclimate.com/github/ManageIQ/manageiq-providers-ibm_cloud/issues)


jaywcarman:
> Found via [CodeClimate Issutes](https://codeclimate.com/github/ManageIQ/manageiq-providers-ibm_cloud/issues)

@chessbyte  Thanks for including the reference - I hadn't checked CodeClimate out yet!

jaywcarman (approved):

PR #21, opened 2020-09-09
Add APP_PATH to allow using Rails directly

agrare:
Allow calling e.g. `rails c` directly from the plugin by adding APP_PATH to the bin/rails script

This was blasted out to all the plugins but this one seems to have been missed, see https://github.com/ManageIQ/manageiq-providers-amazon/pull/641

cc @Fryguy


Fryguy:
My plugin blaster outer should catch this. 😁

PR #22, opened 2020-09-11
Add PowerVS CloudVolume Spec

rspradeep:
While creating cloudvolume spec I am getting this error

```
Randomized with seed 2381
FF

Failures:

  1) ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume cloud volume operations #delete_volume deletes the cloud volume
     Failure/Error:
       with_aws_stubbed(stubbed_responses) do
         expect(cloud_volume.delete_volume).to be_truthy
       end
     
     NoMethodError:
       undefined method `with_aws_stubbed' for #<RSpec::ExampleGroups::ManageIQProvidersIbmCloudPowerVirtualServersStorageManagerCloudVolume::CloudVolumeOperations::DeleteVolume:0x000000000a7d6ba8>
     # ./spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb:17:in `block (4 levels) in <top (required)>'

  2) ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume cloud volume operations #delete_volume catches error from the provider
     Failure/Error:
       with_aws_stubbed(stubbed_responses) do
         expect do
           cloud_volume.delete_volume
         end.to raise_error(MiqException::MiqVolumeDeleteError)
       end
     
     NoMethodError:
       undefined method `with_aws_stubbed' for #<RSpec::ExampleGroups::ManageIQProvidersIbmCloudPowerVirtualServersStorageManagerCloudVolume::CloudVolumeOperations::DeleteVolume:0x000000000a31fec0>
     # ./spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb:29:in `block (4 levels) in <top (required)>'

Finished in 0.0284 seconds (files took 8.88 seconds to load)
2 examples, 2 failures

Failed examples:

rspec ./spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb:10 # ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume cloud volume operations #delete_volume deletes the cloud volume
rspec ./spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb:22 # ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume cloud volume operations #delete_volume catches error from the provider

Randomized with seed 2381

```
```
bundle exec rspec spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb 
```


spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
+describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume do
+  let(:ems) do
+    FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_cloud, :provider_region => "us-south")
+  end
+  let(:StorageManager) { FactoryBot.create(:ems_IbmCloud_StorageManager, :parent_ems_id => ems_cloud.id) }
```
jaywcarman, line 5: This factory needs to be defined in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/spec/factories/ext_management_system.rb

Let's be consistent with the cloud manager, naming it:
`factory :ems_ibm_cloud_power_virtual_servers_storage`.

agrare:
Hey it looks like you copied this from the Amazon provider (ec2 and aws) without also bringing in the dependent methods that you're calling which is why it is failing.  You need to write the specs for this repo's API calls not Amazon's.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:cloud_volume) { FactoryBot.create(:cloud_volume_IbmCloud, :ext_management_system => StorageManager ) }
```
jaywcarman, line 6: Also need to define a cloud_volume factory (similar to `:vm_ibm_cloud_power_vritual_servers`, https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/spec/factories/vm.rb).

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "cloud volume operations" do
+    context "#delete_volume" do
+      it "deletes the cloud volume" do
+        stubbed_responses = {
+          :ec2 => {
```
jaywcarman, line 12: `:ec2` is an Amazon reference and should be modified

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+            :delete_volume => {}
+          }
+        }
+
+        with_aws_stubbed(stubbed_responses) do
```
jaywcarman, line 17: The 'with_aws_stubbed' call is specific to aws.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      end
+
+      it "catches error from the provider" do
+        stubbed_responses = {
+          :ec2 => {
+            :delete_volume => "UnauthorizedOperation"
+          }
+        }
+
+        with_aws_stubbed(stubbed_responses) do
```
jaywcarman, line 29: The 'with_aws_stubbed' call is specific to aws.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+          expect(cloud_volume.delete_volume).to be_truthy
+        end
+      end
+
+      it "catches error from the provider" do
+        stubbed_responses = {
+          :ec2 => {
```
jaywcarman, line 24: :ec2 is an Amazon reference and should be modified

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        with_aws_stubbed(stubbed_responses) do
+          expect do
+            cloud_volume.delete_volume
+          end.to raise_error(MiqException::MiqVolumeDeleteError)
+        end
+      end
+    end
+  end
+end
```
jaywcarman, line 37: Needs a newline at the end of the file.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
rspradeep: Hi jay in this ec2 place what i want to do modify, I am trying to Ibmcloud,powervc like that,
even i am getting same error so what i need to change here?

agrare:
@rspradeep I wouldn't start out by copying the aws stubbed responses, I would start by looking at what you're trying to test in this repo, start writing tests, and return mocks for API calls that are made from those tests.

rspradeep:
Hi @agrare  in this ec2 place what i want to do modify, I am trying to Ibmcloud,powervc like that,
even i am getting same error so what i need to change here?
and here also 'with_aws_stubbed' ?

jaywcarman:
@rspradeep Here's the class we're wanting to test:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb

So high level we can test two deletion paths:
1. Delete Volume that is _not_ in use -> response: :+1: 
2. Delete Volume that _is_ in use -> response: "Can't delete volume that is in use."

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,27 @@
+describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume do
+
+  let(:ems_cloud) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_cloud) }
+  let(:ems_storage) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_storage, :parent_ems_id => ems_cloud.id) }
+  let(:cloud_volume) { FactoryBot.create(:cloud_volume_ibm_cloud_power_virtual_servers, :ext_management_system => ems_storage ) }
+
+
+  context "#validate_delete_volume" do
+    it "status is in-use" do
+      expect(cloud_volume).to receive(:status).and_return("in-use")
```
agrare, line 10: `:status` is a column on the `cloud_volumes` table so instead of mocking this you can set the `:status` attribute as part of your `let(:cloud_volume)` above

You can use contexts to do something like

```
let(:cloud_volume) { FactoryBot.create(:cloud_volume_ibm_cloud_power_virtual_servers, :ext_management_system => ems_storage, :status => status) }
describe "#validate_delete_volume" do
  context "status is in-use" do
    let(:status) { "in-use" }

    it "validates unavailable" do
    end
  end

  context "status is available"
    let(:status) { "available" }

    it "validates available" do
    end
  end
end
```

The reason I don't like to mock stuff like this is it can hide things that might end up failing at runtime but be passing specs.  Theoretically if we renamed the `CloudVolume#status` column to something else this spec would still pass but it would blow up at runtime.

agrare (commented):
Looking much much better @rspradeep just the one comment then this LGTM

rspradeep:
Hi @agrare can you Please review on this commits

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:ems_cloud) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_cloud) }
+  let(:ems_storage) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_storage, :parent_ems_id => ems_cloud.id) }
+  let(:cloud_volume) { FactoryBot.create(:cloud_volume_ibm_cloud_power_virtual_servers, :ext_management_system => ems_storage ) }
+
+  describe "#validate_delete_volume" do
+  context "status is in-use" do
+    let(:status) { "in-use" }
+
+    it "validates unavailable" do
+      def validate_delete_volume
```
jaywcarman, line 12: Looks like this is just a copy of [`validate_delete_volume`](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb#L31) from our `CloudVolume` class.

I think there's still some fundamental misunderstanding here. We don't want to copy these class methods to the test, we want to call them from the test.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+    end
+  end
+
+  context "status is available"
+    let(:status) { "available" }
+
+    it "validates available" do
+      def validation_failed(operation, reason)
+        {:available => false, :message   => _("Validation failed for %{name} operation %{operation}. %{reason}") % {:name => name,:operation => operation, :reason => reason}}
```
jaywcarman, line 29: This line isn't necessary. I think all we need to set is done above on line 25: `let(:status) { "available" }`

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        {:available => true, :message => nil}
+      end
+    end
+  end
+
+  context "status is available"
+    let(:status) { "available" }
+
+    it "validates available" do
+      def validation_failed(operation, reason)
```
jaywcarman, line 28: Again, we aren't defining methods in the tests, we're calling existing methods to test them.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        {:available => false, :message   => _("Validation failed for %{name} operation %{operation}. %{reason}") % {:name => name,:operation => operation, :reason => reason}}
+        expects(cloud_volume.validates_delete_volume['available']).to falsy
```
jaywcarman, line 30: This line looks awesome! But still needs some work:
- `expect()` is not plural
- [`validate_delete_volume`](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb#L31) "validate" is not plural
- The return hash key is `:available`, so you need to access it as `validates_delete_volume[:available]`
- The "false" matches is `be_falsy` 
- I think we this test we actually want `be_truthy` - if the volume is available we want to allow deletion.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,34 @@
+describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume do
+
+  let(:ems_cloud) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_cloud) }
+  let(:ems_storage) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_storage, :parent_ems_id => ems_cloud.id) }
+  let(:cloud_volume) { FactoryBot.create(:cloud_volume_ibm_cloud_power_virtual_servers, :ext_management_system => ems_storage ) }
+
+  describe "#validate_delete_volume" do
+  context "status is in-use" do
+    let(:status) { "in-use" }
```
jaywcarman, line 9: I don't think this is doing what we want either. I tried using `update_attributes` and it works, e.g. `cloud_volume.update_attributes(:status => "in-use")`.

I went ahead and pushed a working copy of the spec to my fork:
https://github.com/rspradeep/manageiq-providers-ibm_cloud/compare/cloudvolume_spec...jaywcarman:cloudvolume_spec

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
agrare: Yeah @rspradeep you were a lot closer with this: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/b9ecdf97efa719e9b2a5916dbe9cebd8fb5cc5ca/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
In that you were testing the methods on the objects rather than copying the method into the tests.

What would happen if we modified the method and introduced a bug but didn't update the tests?  The tests would still be passing,

jaywcarman:
@agrare Cloud volume attach/detach specs are on my TODO list so I combined them with this WIP.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,81 @@
+describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume do
+  let(:ems_cloud) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_cloud) }
+  let(:ems_storage) { FactoryBot.create(:ems_ibm_cloud_power_virtual_servers_storage, :parent_ems_id => ems_cloud.id) }
```
agrare, line 3: You don't need to pass the ID, just the object is enough: `:parent_ems => ems_cloud)`

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:cloud_volume) { FactoryBot.create(:cloud_volume_ibm_cloud_power_virtual_servers, :ext_management_system => ems_storage) }
+
+  describe "#validate_delete_volume" do
+    it "validates volume delete when in-use" do
+      cloud_volume.update(:status => "in-use")
```
agrare, line 8: Usually with rspec we'll do something like this:
```
let(:cloud_volume) { FactoryBot.create(..., :status => status) }

describe "#validate_delete_volume" do
  context "when available" do
    let(:status) { "available" }
  end

  context "when unavailable" do
    let(:status) { "unavailable" }
  end
end
```

instead of `cloud_volume.update(:status)` since cloud_volume is actually created every `it` you're basically doing `FactoryBot.create().update()` here

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(cloud_volume.validate_delete_volume[:available]).to be_falsy
+    end
+
+    it "validates volume delete when available" do
+      cloud_volume.update(:status => "available")
+      expect(cloud_volume.validate_delete_volume[:available]).to be_truthy
+    end
+
+    it "validates volume delete when ems is missing" do
+      expect(cloud_volume).to receive(:ext_management_system).and_return(nil)
```
agrare, line 18: ext_management_system is a proper association so you shouldn't stub it, just create a context like "when ems is missing" and `let(:ems_storage) { nil }`

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  describe "#validate_attach_volume" do
+    context "in-use" do
+      before do
+        cloud_volume.update(:status => "in-use")
+      end
+
+      it "validates volume attach when in-use and not sharable" do
+        cloud_volume.update(:multi_attachment => false)
```
agrare, line 31: Same comment here re: .update

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
jaywcarman: Done.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
jaywcarman: Done, and I finally understand how `let()` is used. There is some clutter from contexts with only one test. Would it be better to try to clean that up, or to keep a uniform format across all tests?

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
jaywcarman: Done.

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume_spec.rb
agrare: One thing you could do is set some defaults at the top level like multi-attachment false so you don't have to set it in every context if it isn't relevant to the context you're testing

agrare (approved):
Looks great

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -47,7 +47,14 @@ def raw_delete_volume
   end
 
   def validate_attach_volume
-    validate_volume_available
+    msg = validate_volume_available
+    return {:available => msg[:available], :message => msg[:message]} unless msg[:available]
+
+    if status == "in-use"
+      return validation_failed("Attach Volume", "Can't attach non-shareable volume that is in use.") unless multi_attachment
```
agrare, line 54: Looks a little strange to split the conditionals, can this be just `if status == "in-use" && !multi_attachment` ?

agrare (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
agrare: Yeah I only use unless for a single negative conditional, anything more complex gets `if`

PR #29, opened 2020-09-16
Initial addition of IBM provider. 

Real-Omar-Afifi:


app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def raw_connect(api_key)
+      if api_key.blank?
+        raise MiqException::MiqInvalidCredentialsError, _('Missing credentials')
+      end
+      require 'ibm-cloud-sdk'
+      IBM::Cloud::SDK::IAM.new(api_key).get_identity_token
+    end
+
+    def raw_tenant_id(creds)
```
agrare, line 94: Is this used anywhere?  I saw it left over in the original powervs code at some point so maybe copied from before it was deleted?

agrare:
Just some rubocop issues but overall this looks great @Real-Omar-Afifi

agrare:
ibm-cloud-sdk v0.1.4 released, restarting the tests

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
```diff
@@ -0,0 +1,19 @@
+class ManageIQ::Providers::IbmCloud::Inventory::Collector::VPC < ManageIQ::Providers::IbmCloud::Inventory::Collector
+  require_nested :CloudManager
+
+  def connection
+    manager.connect
```
agrare, line 5: Probably want to cache this

```
def connection
  @connection ||= manager.connect
end
```

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
```diff
@@ -0,0 +1,19 @@
...
+  def vms
+    connection.instances.all
+  end
+
+  def images
+    connection.images.all
+  end
+
+  def image(image_id)
+    connection.images.instance(image_id).details
```
agrare, line 17: If this can return `nil` you might want to `&.details` so it doesn't raise an exception if it can't e.g. find that image by id (got deleted between calling `connection.images.all` and calling `connection.images.instance(image_id)` for example)

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def parse
+    images
+    instances
+  end
+
+  def images
+    collector.images.each do |image|
+      img_to_os[image[:id]] = image&.dig(:operating_system, :name)
```
agrare, line 17: If we do this in the collector it'll be a little easier to work with when we do targeted refresh (because we can control the targets from there unlike the parser).

e.g.
```
class Collector
  def images
    @images ||= collector.images.all
  end

  def operating_system_by_image_id
    @operating_system_by_image_id ||= images.each_with_object({}) { |img, h| h[img[:id]] = img.dig(:operating_system, :name) }
  end
end
```

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -0,0 +1,73 @@
...
+      )
+    end
+  end
+
+  def instances
+    collector.vms.each do |instance|
+      persister_instance = persister.vms.build(
+        :description      => "IBM Cloud Server",
+        :ems_ref          => instance[:id],
+        :location         => instance&.dig(:zone, :name),
```
agrare, line 38: `:location` has a `validates_presence_of` so if it `instance&.dig(:zone, :name` can be `nil` you probably want to `instance&.dig(:zone, :name) || "unknown"`

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      instance_hardware(persister_instance, instance)
+      instance_operating_system(persister_instance, instance)
+    end
+  end
+
+  def instance_hardware(persister_instance, instance)
+    persister.hardwares.build(
+      :vm_or_template  => persister_instance,
+      :cpu_sockets     => Float(instance&.dig(:vcpu, :count)).ceil,
```
agrare, line 55:
```
>> Float(nil)
TypeError (can't convert nil into Float)
```

How about we do
```
vcpu_count = instance&.dig(:vcpu, :count)
cpus = Float(vcpu_count).ceil if vcpu_count
persister.hardwares.build(
  :cpu_sockets => cpus, :cpu_total_cores => cpus
)
```

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -0,0 +1,73 @@
...
+      :cpu_total_cores => Float(instance&.dig(:vcpu, :count)).ceil,
+      :memory_mb       => Integer(instance[:memory]) * 1024
```
agrare, line 57: Same with memory if this can be nil

agrare:
Couple minor things but overall looking good

agrare:
Would be nice to add a Refresher spec but we can do that in a follow-up

agrare (approved):
:+1: LGTM awesome work @Real-Omar-Afifi

PR #30, opened 2020-09-16
Changes to add the system & storage type collection information to a cloud instance

Kuldip-Nanda:


Kuldip-Nanda:
Adam is helping me in this. I have made this as a draft and may close this PR without merging.

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -38,4 +38,23 @@ def ports(network_id)
   def sshkeys
     connection.get_ssh_keys
   end
+
+  def system_pool
+    connection.get_system_pool
```
agrare, line 43: @Kuldip-Nanda what does a `storage_pool` response look like?  Wondering if Flavor is the right way to model this or if we should do something else

Kuldip-Nanda: collector.system_pool.values returns json list of available system pool (Hardware platform detailing its limits and statistics) within a particular data center. I just wanted to adhere to the practice of not interpreting the return value. The parsing of the system pools object is giving the system available. I can change it to collector.system_pool.values

jaywcarman: You can return an array of system types (all that we're really collecting right now) with `connection.get_system_pool.keys`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -197,4 +199,22 @@ def sshkeys
       persister.auth_key_pairs.build(:name => tenant_key[:name])
     end
   end
+
+  def systemtypes
+    systype_lst = []
+    collector.system_pool.values.each { |v| systype_lst |= [v['type']] }
```
agrare, line 205: You can do this with `.map` which builds an array out of each value returned by the block so e.g. `collector.system_pool.values.map { |v| v["type"] }`

app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
```diff
@@ -45,6 +45,16 @@ def initialize_cloud_inventory_collections
       builder.add_default_values(:ems_id => ->(persister) { persister.cloud_manager.id })
     end
 
+    add_collection(cloud, :system_types) { |builder|
```
agrare, line 48: Blocks should be `{ |var| }` if one line and `do |var|\nend` if mutli-line

app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
```diff
@@ -45,6 +45,16 @@ def initialize_cloud_inventory_collections
...
+      builder.add_properties(:system_types => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemTypes)
```
agrare, line 49: I think what you're looking for here is `builder.add_properties(:type => ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemTypes)`

app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
```diff
@@ -45,6 +45,16 @@ def initialize_cloud_inventory_collections
...
+      builder.add_default_common_values
+    }
+
+    add_collection(cloud, :storage_types) { |builder|
+      builder.add_properties(:storage_types => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemTypes)
+      builder.add_default_common_values
```
agrare, line 55: `builder.add_common_default_values` will add `:ems_id` as your `persister.manager.id`, `add_default_common_values` isn't an existing method

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
Kuldip-Nanda: I am returning collector.system_pool.values as suggested by Adam.

app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
Kuldip-Nanda: Oops. thanks.

app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
```diff
@@ -45,6 +45,16 @@ def initialize_cloud_inventory_collections
       builder.add_default_values(:ems_id => ->(persister) { persister.cloud_manager.id })
     end
 
+    add_collection(cloud, :system_types) do |builder|
+      builder.add_properties(:types => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemTypes)
+      builder.add_common_default_values
+    done
+
+    add_collection(cloud, :storage_types) do |builder|
+      builder.add_properties(:types => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemTypes)
```
agrare, line 54: The column name is `:type`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -197,4 +199,26 @@ def sshkeys
       persister.auth_key_pairs.build(:name => tenant_key[:name])
     end
   end
+
+  def systemtypes
+    collector.system_pool.each do |v|
+      persister.flavor.build(
+        :types   => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType,
```
agrare, line 206: this should just be `:type`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -197,4 +199,26 @@ def sshkeys
       persister.auth_key_pairs.build(:name => tenant_key[:name])
     end
   end
+
+  def systemtypes
+    collector.system_pool.each do |v|
+      persister.flavor.build(
```
agrare, line 205: Think this should be `persister.flavors`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -197,4 +199,26 @@ def sshkeys
...
+        :types   => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType,
+        :ems_ref => v['type']
+      )
+    end
+  end
+
+  def storagetypes
+    region = collector.region
+    collector.storage_types[region].each do |v|
```
agrare, line 214: If we're using `collector.region` and passing it back to `collector.storage_types[region]` could we just do `connection.get_storage_types[region]` unless you want to be able to get storage types from other regions elsewhere

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -197,4 +199,26 @@ def sshkeys
...
+      next unless v['state'] == 'active'
+
+      persister.flavors.build(
+        :types   => ::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::StorageType,
```
agrare, line 218: Same, `:type`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: again, I initially thought of doing it there, but decided as I wanted to do no parsing in the collector functionality. But it can be put there. Either way this changes when the backend finalizes/corrects their web api. I can move it in the collector code though.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -15,6 +15,18 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
           :autosave    => true,
           :dependent   => :destroy
 
+  has_many :system_types,
+           :foreign_key => :parent_ems_id,
```
agrare, line 19: The foreign key should be `:ems_id`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -15,6 +15,18 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
...
+           :class_name  => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType",
+           :autosave    => true,
+           :dependent   => :destroy
```
agrare, line 22: For this association you don't need autosave or dependent destroy

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -15,6 +15,18 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
...
+
+  has_many :storage_types,
+           :foreign_key => :parent_ems_id,
```
agrare, line 25: Again should be ems_id

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -15,6 +15,18 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
...
+           :class_name  => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::StorageType",
+           :autosave    => true,
+           :dependent   => :destroy
```
agrare, line 28: And don't need these two

Kuldip-Nanda:
This has been reviewed and ready to be merged.

agrare:
@Kuldip-Nanda looks like you need to re-record the vcr specs.  Remove the `spec/vcr_cassettes/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher.yml` set the rails secrets and re-run the specs

chessbyte:
@Kuldip-Nanda looks like this PR needs to be rebased

Kuldip-Nanda:
1. Can the unmergable label be removed as this PR has been rebased.
2. I have run new VCR file has been successfully locally on my laptop. The travis build has also successfully run.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -59,4 +65,8 @@ def self.ems_type
   def self.description
     @description ||= "IBM Power Systems Virtual Servers".freeze
   end
+
+  def region
+    provider_region.empty? ? "us-south" : provider_region
+  end
```
agrare, line 71: This isn't needed because you're handling it in the collector here https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/30/files#diff-53f571cf57b42afefdd9b8a927cff8beR50-R52

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -38,4 +38,23 @@ def ports(network_id)
...
+    connection.get_storage_types[region]
+  end
+
+  def region
+    manager.provider_region || "us-south"
+  end
+
+  def guid
+    manager.uid_ems
+  end
```
agrare, line 56: This doesn't look like it is used anywhere

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -38,4 +38,23 @@ def ports(network_id)
...
+
+  def refresh
+  end
```
agrare, line 59: This doesn't look like it is used either

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -204,4 +206,26 @@ def sshkeys
       persister.auth_key_pairs.build(:name => tenant_key[:name])
     end
   end
+
+  def systemtypes
+    collector.system_pool.each do |v|
+      persister.flavors.build(
+        :type    => "::ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType",
```
agrare, line 213: Minor but you don't need the leading `::` before the `ManageIQ::...etc`

That only is used to indicate a top-level class name, e.g. if you are in a namespace like
`module ManageIQ::Providers` and there is a `class ManageIQ::Providers::Flavor`, it is ambiguous when you just say `Flavor` because ruby will look to your current namespace first and find something.  If you really need the top-level class you can use `::Flavor` and it will skip your current namespace.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/system_type.rb
```diff
@@ -0,0 +1,13 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType < ::Flavor
+  supports :provisioning do
+    if ext_management_system
+      unsupported_reason_add(:provisioning, ext_management_system.unsupported_reason(:provisioning)) unless ext_management_system.supports_provisioning?
+    else
+      unsupported_reason_add(:provisioning, _('not connected to ems'))
+    end
+  end
```
agrare, line 8: Are flavors actually supported for provisioning?  I thought templates were and you selected flavors as part of the workflow

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/system_type.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def provider_object(_connection = nil)
+    ext_management_system.connect
+  end
```
agrare, line 12: The `provider_object` method is supposed to return the native object, `with_provider_connection` is what would return the connection.

It doesn't look like you're using either though so could probably be dropped

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/storage_type.rb
```diff
@@ -0,0 +1,13 @@
...
+    if ext_management_system
+      unsupported_reason_add(:provisioning, ext_management_system.unsupported_reason(:provisioning)) unless ext_management_system.supports_provisioning?
+    else
+      unsupported_reason_add(:provisioning, _('not connected to ems'))
+    end
+  end
+
+  def provider_object(_connection = nil)
+    ext_management_system.connect
+  end
```
agrare, line 12: Both comments from :point_down: apply here as well

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
Kuldip-Nanda: was existing already. Don't know why it was put. Going to remove it also.

agrare: Maybe a merge conflict thing?  It looks like it is being added by your commit

Kuldip-Nanda: yes, maybe

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/system_type.rb
Kuldip-Nanda: Not supported. I have removed these

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
Kuldip-Nanda: removed

agrare (approved):

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -38,4 +38,16 @@ def ports(network_id)
...
+  def system_pool
+    connection.get_system_pool.values
+  end
+
+  def storage_types
+    connection.get_storage_types[region]
+  end
+
+  def region
+    manager.provider_region || "us-south"
```
agrare, line 51: We can look at this in a follow-up but I think `provider_region` is required (could be added as a validate presence to the cloud_manager) so the `|| "us-south"` shouldn't be needed.  Would rather enforce that it be set than handle the possibility that it might be `nil` everywhere.

jaywcarman: #40 removes this region method

Kuldip-Nanda: sure will do it.

jaywcarman: We can still set the `provider_region` at registration time, especially if it's used for UI elements.

PR #32, opened 2020-09-17
Add credential validation to the provider form

agrare:
Screenshot from 2020-09-17 10-54-07

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
@@ -33,23 +33,52 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::ManagerMixin
       @params_for_create ||= {
         :fields => [
           {
-            :component  => "text-field",
-            :name       => "uid_ems",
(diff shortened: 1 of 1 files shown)


agrare:
cc @skateman

skateman:
@agrare schema looks fine to me, does the validation actually work?

agrare:
It successfully fails :laughing: I don't have a valid key but it does fail for an invalid api key and prevent submitting the form

skateman:
> It successfully fails  I don't have a valid key but it does fail for an invalid api key and prevent submitting the form

Sounds like it's working then :laughing: can we get some creds to try this out? I have the same issue at the NSX-T provider :confused:

jaywcarman:
It works!

If I try to validate with a working API Key and an invalid PowerVS Service GUID then I get:
`Validation failed: undefined method `crn' for nil:NilClass`

For a future PR would it be possible to catch that error and return a translation? (e.g. invalid GUID for API Key)

Fryguy:
i-have-the-power-he-man-min

agrare:
@jaywcarman yes we should be catching those types of exception and raising better errors

jaywcarman (approved):

PR #33, opened 2020-09-18
adding availability zones with links to the VM. Added ssh keys (linking to the VM will come after sdk support)

Real-Omar-Afifi:


agrare:
Kicking the spec tests

PR #38, opened 2020-09-21
Re-record VCRs for update to Resource API

agrare:
https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/pull/18 changed what HTTP call was made to retrieve a resource by GUID which broke our VCRs


agrare:
cc @jaywcarman

PR #41, opened 2020-09-23
Added Network Manager. More discovery, linking and specs.

Real-Omar-Afifi:
Discovering flavors, security groups, cloud networks, and cloud subnets. Linking ssh keys to VM instances. Added cloud manager refresher and refresher spec

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -1,5 +1,6 @@
 class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Providers::IbmCloud::Inventory::Parser
   require_nested :CloudManager
+  require_nested :NetworkManager
 
   attr_reader :img_to_os
 
@@ -8,8 +9,12 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   end
 
   def parse
+    cloud_networks
+    cloud_subnets
+    security_groups
     availability_zones
     auth_key_pairs
+    flavors
     images
     instances
   end
@@ -40,6 +45,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
         :location          => instance&.dig(:zone, :name) || "unknown",
         :genealogy_parent  => persister.miq_templates.lazy_find(instance&.dig(:image, :id)),
         :availability_zone => persister.availability_zones.lazy_find(instance&.dig(:zone, :name)),
+        :flavor            => persister.flavors.lazy_find(instance&.dig(:profile, :name)),
+        :key_pairs         => instance_key_pairs(instance[:id]).compact,
         :name              => instance[:name],
         :vendor            => "ibm",
         :connection_state  => "connected",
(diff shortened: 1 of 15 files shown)


Real-Omar-Afifi:
@agrare need your opinion on https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/41/files#diff-97aa719b0165ce9439cb019b669a87faR49

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -83,6 +105,14 @@ def auth_key_pairs
     end
   end
 
+  def instance_key_pairs(instance_id)
+    keys = []
+    collector.vm_key_pairs(instance_id)[:keys]&.each do |key|
+      keys << persister.auth_key_pairs.lazy_find(key[:name])
+    end
+    keys
```
agrare, line 113: You could use `.map` to handle the array concatenation for you like:

```
def instance_key_pairs(instance_id)
  collector.vm_key_pairs(instance_id)[:keys].to_a.map do |key|
    persister.auth_key_pairs.lazy_find(key[:name])
  end
end
```

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -40,6 +45,8 @@ def instances
         :location          => instance&.dig(:zone, :name) || "unknown",
         :genealogy_parent  => persister.miq_templates.lazy_find(instance&.dig(:image, :id)),
         :availability_zone => persister.availability_zones.lazy_find(instance&.dig(:zone, :name)),
+        :flavor            => persister.flavors.lazy_find(instance&.dig(:profile, :name)),
+        :key_pairs         => instance_key_pairs(instance[:id]).compact,
```
agrare, line 49: I don't think the `.compact` is necessary because you're not e.g. using `next if something` in the block which would lead to `nil`s in your array.

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.rb
```diff
@@ -0,0 +1,39 @@
...
+    ems.reload
+    expect(ems.vms.count).to eq(2)
+    expect(ems.miq_templates.count).to eq(39)
+    expect(ems.key_pairs.count).to eq(1)
+    expect(ems.availability_zones.count).to eq(3)
+
+    vm = ems.vms[1]
+    expect(vm.availability_zone.name).to eq('us-east-3')
+    expect(vm.cpu_total_cores).to eq(2)
+    expect(vm.hardware.memory_mb).to eq(16 * 1024)
```
agrare, line 18: Really minor but this can be `16.gigabytes`

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(ems.cloud_networks.count).to eq(1)
+    expect(ems.cloud_subnets.count).to eq(1)
+    expect(ems.cloud_subnets[0].cloud_network_id).to eq(ems.cloud_networks[0].id)
+
+    ## linking key pairs to vms
+    expect(vm.key_pairs.count).to eq(1)
+    expect(vm.key_pairs[0].name).to eq('cloudforms')
+    expect(vm.key_pairs[0].fingerprint).to eq('SHA256:w6v2HXjIgk/2yxiVs8cvnt1AxyxVVDsVRcWlNRoyCRE')
+
+    ems.destroy
```
agrare, line 37: Shouldn't need to destroy the ems after, with transactional fixtures the db changes during the spec are rolled back

agrare (approved):

PR #42, opened 2020-09-23
Add power state mapping

chadh1313:


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -1,2 +1,20 @@
 class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers::CloudManager::Vm
+  # https://cloud.ibm.com/apidocs/vpc#get-instance
+  POWER_STATES = {
+    "running"       => "on",
+    "failed"        => "off",
+    "paused"        => "suspended",
+    "pausing"       => "suspended",
+    "pending"       => "suspended",
+    "restarting"    => "powering_up",
```
agrare, line 9: Rubocop is complaining because you have too many spaces here.  You want to align the `=>` with the longest key so

```
POWER_STATES = {
  "running"    => ...
  "restarting" => ...
}
```

agrare (commented):
Looks great, just the one rubocop issue

Can you add a spec [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb#L39) to check the `vm.raw_power_state` and the `vm.power_state`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -1,2 +1,20 @@
 class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers::CloudManager::Vm
+  # https://cloud.ibm.com/apidocs/vpc#get-instance
+  POWER_STATES = {
+    "running"       => "on",
+    "failed"        => "off",
+    "paused"        => "suspended",
```
jaywcarman, line 6: There's a couple of these states that you could map more directly and still get [icons](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/lib/gtl_formatter.rb#L19-L41):
```
'paused'     => 'paused',
'pausing'    => 'paused',
'restarting' => 'reboot_in_progress',
```

chadh1313: Thanks Jay - will add these.

agrare (approved):

PR #43, opened 2020-09-23
Refactor VPC::CloudManager::Refresher Spec

agrare:
Follow-up to https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/41 I missed that the refresher spec file name was incorrect and thus wasn't being run on travis


spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb
```diff
@@ -0,0 +1,53 @@
+describe ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher do
+  let(:ems) do
+    api_key = Rails.application.secrets.ibmcvs.try(:[], :api_key) || "IBMCVS_API_KEY"
+    FactoryBot.create(:ems_ibm_cloud_vpc, :provider_region => "us-east").tap do |ems|
+      ems.authentications << FactoryBot.create(:authentication, :auth_key => api_key)
+    end
+  end
+
+  it "tests the refresh" do
+    2.times do
```
agrare, line 10: We like to run the refresh twice to make sure the refresh is idempotent

PR #44, opened 2020-09-23
Set PowerVS SystemType name

jaywcarman:
Flavors are displayed in the UI with a subset of the DB table columns.
Setting ':name' here so the user doesn't get a UI table with a bunch of
blank rows.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -211,7 +211,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
     collector.system_pool.each do |v|
       persister.flavors.build(
         :type    => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType",
-        :ems_ref => v['type']
+        :ems_ref => v['type'],
+        :name    => v['type']
       )
     end
   end


jaywcarman:
Before:
image

After:
image

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -211,7 +211,8 @@ def systemtypes
     collector.system_pool.each do |v|
       persister.flavors.build(
         :type    => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType",
-        :ems_ref => v['type']
+        :ems_ref => v['type'],
+        :name => v['type']
```
agrare, line 215: Can you line up the `=>` ?

agrare: D'oh too slow https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/44#issuecomment-697956776

jaywcarman: > but you can still beat him to it

Winner: @miq-bot  :smile:

agrare (approved):

agrare:
@jaywcarman https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/45 should fix the specs starting to fail after an hour :)

PR #46, opened 2020-09-23
Collect assigned virtual cores for PowerVS Instances

Kuldip-Nanda:
Also correcting the get_storage_types call as we are getting region specific info

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -45,8 +45,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
       # saving hardware information (CPU, Memory, etc.)
       ps_hw = persister.hardwares.build(
         :vm_or_template  => ps_vmi,
-        :cpu_total_cores => Float(instance["processors"]).ceil,
-        :memory_mb       => Integer(instance["memory"]) * 1024
+        :cpu_total_cores => instance['virtualCores']['assigned'],
+        :memory_mb       => instance["memory"] * 1024
       )
 
       # saving instance disk information


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -45,7 +45,7 @@ def instances
       # saving hardware information (CPU, Memory, etc.)
       ps_hw = persister.hardwares.build(
         :vm_or_template  => ps_vmi,
-        :cpu_total_cores => Float(instance["processors"]).ceil,
+        :cpu_total_cores => Float(instance['virtualCores']['assigned']),
```
jaywcarman, line 48: `instance['virtualCores']['assigned']` should already be an integer

https://cloud.ibm.com/apidocs/power-cloud#pcloud-pvminstances-get

Kuldip-Nanda: the column is original a Float Jay. So anyways it will be typecast automatically to Float. Unless we want to change that.

jaywcarman:
restarting Travis

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -45,7 +45,7 @@ def instances
       # saving hardware information (CPU, Memory, etc.)
       ps_hw = persister.hardwares.build(
         :vm_or_template  => ps_vmi,
-        :cpu_total_cores => Float(instance["processors"]).ceil,
+        :cpu_total_cores => Integer(instance['virtualCores']['assigned']),
```
jaywcarman, line 48: `instance['virtualCores']['assigned']` is already an Integer type so there's no need to pass it to `Integer()`:
```
irb(main):039:0> instance['virtualCores']['assigned'].kind_of? Integer
=> true
```

We can do this in another commit, but the same goes for the next line with `instance["memory"]`.

jaywcarman (approved):

agrare (approved):

PR #47, opened 2020-09-24
Added support for Cloud Volume creation in IBM PVS provisioning

iv1111:
- conerns IBM Power Virtual Server provisioning form
- it is possible now to create new cloud volumes during
  the provisioning itself
- added "New Volumes" tab to the form
- fields: (volume) name, size (gb), shareable (bool), disk-type
  (currently) text field
- multiple volumes can be created
- current limitations are missing validation on the form itself,
  disk-type is not a pull-down field, missing clear volume button,
  missing the ignoring of empty fields of the volume added on UI
  dynamically
- corresp. UI classic changes are in this PR:
  https://github.com/ManageIQ/manageiq-ui-classic/pull/7338

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -50,10 +50,42 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
     Hash[cloud_volumes || {}]
   end
 
-  def validate_entitled_processors(_field, values, _dlg, _fld, value)
+  def set_request_values(values)
+    values[:new_volumes] = parse_new_volumes_fields(values)
+    super
+  end
+
+  def parse_new_volumes_fields(values)
+    stop = false
+    new_volumes = []
+
+    while not stop
+      new_volume = {}
+
+      %w(name size diskType shareable).map do |fld|
+        cnt = new_volumes.length+1
+        key = (:"#{fld}_#{cnt}").to_sym
+        new_volume[fld.to_sym] = values[key] if values.key?(key)
+      end
+
+      stop = new_volume.empty?
+
+      if not stop
+        new_volume[:name] = nil if new_volume[:name].blank?
+        new_volume[:diskType] = nil if new_volume[:diskType].blank?
+        new_volume[:size] = new_volume[:size].blank? ? nil : new_volume[:size].to_i
+        new_volume[:shareable] = ['null', nil].exclude?(new_volume[:shareable])
+        new_volumes << new_volume
+      end
+    end
+
+    new_volumes
+  end
+
+  def validate_entitled_processors(field, values, dlg, fld, value)
     dedicated = values[:instance_type][1] == 'dedicated'
 
-    fval = /^\s*[\d]*(\.[\d]+)?\s*$/.match?(value) ? value.strip.to_f : 0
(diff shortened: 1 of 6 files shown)


jaywcarman (approved):
I'm OK with leaving the remaining miq-bot & code climate items as-is.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -50,6 +54,38 @@ def allowed_cloud_volumes(_options = {})
...
+  def set_request_values(values)
+    values[:new_volumes] = parse_new_volumes_fields(values)
+    super
+  end
+
+  def parse_new_volumes_fields(values)
+    stop = false
+    new_volumes = []
+
+    while !stop
```
agrare, line 66: Feel like we can do this a little better, something like
```
new_volumes = []
values.select { |k, _v| k =~ /(name|size|diskType|shareable)_(\d+)/ }.each do |key, value|
  field, cnt = key.split("_")
  cnt = Integer(cnt)

  new_volumes[cnt] ||= {}
  new_volumes[cnt][field] = value
end
```

agrare: You could even do `/(#{volume_dialog_keys.join("|")})_(\d+)/`

agrare:
@iv1111 `parse_new_volumes_fields` seems sufficiently complex to require a spec test

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
iv1111: Hi. The "values" hash doesn't always contain sequential number after the '_' symbol. I.e. the stucture can look like this: 
{name_**1** => valueA, name_**3** => valueB, ...}.

iv1111: So I'm afraid we'd have to use a counter here.

iv1111: or maybe we could just add to the end of the 'new_volumes' array

iv1111: Or we could leave it like this and just do "new_volumes.compact!" after the loop. What do you think? If I'm not wrong in Ruby the unset array values between indices "a" and "b" will be "nil".

iv1111: I've tested it. "shareable" has to be explicitely checked to be not "nil" because only in that case the value must be set to "true". Also there might be an inconsistency if you add multiple volumes in one of which you'd leave all fields blank. Then UI would show all volumes up to the "blank one" but the back-end would consider all new volumes including the ones after the "blank one".

iv1111:
I did the corresponding changes as discussed with slight modifications as it won't work otherwise.

iv1111:
A _spec_ is pending.

iv1111:
@jaywcarman has implemented the _spec_ and I added one more test. I guess we're good to go. Let's wait for the Travis build.

iv1111:
Oh, there are some indentation problems. I will fix now.

iv1111:
Ok. Fixed indentation now. We can wait for the Travis only now.

agrare (approved):

PR #48, opened 2020-09-24
Switch to OpenAPI based gem

agrare:
This is now using the official `ibm_vpc` gem https://rubygems.org/gems/ibm_vpc

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector.rb
@@ -1,4 +1,25 @@
 class ManageIQ::Providers::IbmCloud::Inventory::Collector < ManageIQ::Providers::Inventory::Collector
   require_nested :PowerVirtualServers
   require_nested :VPC
+
+  private
+
+  def all(api_client, list_method, args, collection_key = nil, opts = {})
+    return enum_for(:all, api_client, list_method, args, collection_key, opts) unless block_given?
+
+    collection_key ||= list_method.to_s.sub("list_", "")
+    result = api_client.send(list_method, *args, opts)
+
+    result.send(collection_key)&.each { |i| yield i }
+
+    next_link = result._next&.href if result.respond_to?(:_next)
+    return unless next_link
+
+    start = parse_next_link(next_link)["start"]&.first
+    all(api_client, list_method, args, collection_key, opts.merge(:start => start))
+  end
+
+  def parse_next_link(next_link)
+    CGI.parse(URI(next_link).query)
+  end
 end
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
@@ -3,63 +3,108 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::VPC < ManageIQ::Provi
   require_nested :NetworkManager
   require_nested :StorageManager
 
-  def connection
-    @connection ||= manager.connect
+  def collect
   end
 
   def vms
-    connection.instances.all
+    all("instances")
   end
 
   def vm_key_pairs(vm_id)
-    connection.instances.instance(vm_id)&.initialization || {}
+    vpc.get_instance_initialization(:id => vm_id)&.result&.dig("keys") || []
   end
 
   def flavors
-    connection.instance_profiles.all
+    vpc.list_instance_profiles&.result&.dig("profiles")
   end
 
   def images
-    @images ||= connection.images.all.to_a
+    @images ||= images_by_id.values
   end
 
   def images_by_id
-    @images_by_id ||= images.index_by { |img| img[:id] }
+    @images_by_id ||= all("images").index_by { |image| image["id"] }
   end
 
   def image(image_id)
-    connection.images.instance(image_id)&.details
+    vpc.get_image(:id => image_id)
   end
 
   def keys
-    connection.keys.all
+    vpc.list_keys&.result&.dig("keys")
   end
 
   def availability_zones
-    connection.regions.instance(manager.provider_region).zones.all
+    vpc.list_region_zones(:region_name => manager.provider_region)&.result&.dig("zones")
   end
 
   def security_groups
-    connection.security_groups.all
+    all("security_groups")
   end
 
   def cloud_networks
-    connection.vpcs.all
+    all("vpcs")
   end
 
   def cloud_subnets
-    connection.subnets.all
+    all("subnets")
   end
 
   def floating_ips
-    connection.floating_ips.all
+    all("floating_ips")
+  end
+
+  def tags(attached_to)
+    global_tagging.list_tags(:attached_to => attached_to)&.result&.dig("items") || []
   end
 
   def volumes
-    connection.volumes.all
+    all("volumes")
   end
 
   def volume(volume_id)
-    connection.volumes.instance(volume_id)&.details
+    vpc.get_volume(:id => volume_id)&.result
+  end
+
+  private
+
+  def api_key
+    @api_key ||= manager.authentication_key
+  end
+
+  def token
+    @token ||= manager.class.raw_connect(api_key)
+  end
+
+  def vpc
+    @vpc ||= begin
+      require "ibm_vpc"
+      service_url = "https://#{manager.provider_region}.iaas.cloud.ibm.com/v1"
+      IbmVpc::VpcV1.new(:version => "2020-08-01", :authenticator => token, :service_url => service_url)
+    end
+  end
+
+  def global_tagging
+    @global_tagging ||= begin
+      require "ibm_cloud_global_tagging/global_tagging_v1"
+      IbmCloudGlobalTagging::GlobalTaggingV1.new(:authenticator => token)
+    end
+  end
+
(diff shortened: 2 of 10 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -3,47 +3,92 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
...
 
-  def system_pool
-    connection.get_system_pool.values
+  def system_pools
+    @system_pools ||= system_pools_api.pcloud_systempools_get(cloud_instance_id)
   end
 
   def storage_types
-    connection.get_storage_types
+    # TODO
```
jaywcarman, line 51: I don't think the [/broker/v1/storage-types](https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/blob/master/lib/ibm/cloud/sdk/power_iaas.rb#L169) is documented anywhere and I'm not sure when it will be added to the Power Cloud API.

is it a bad idea to a) manually patch the generated SDK b) make a `RestClient` call directly from here?

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -3,47 +3,92 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
...
+  def connection
+    @connection ||= manager.connect
+  end
+
+  def cloud_instance_id
+    @cloud_instance_id ||= manager.uid_ems
+  end
+
+  def tenant_id
+    # TODO
```
jaywcarman, line 67: Do we need to generate the entire [IBM Cloud Resource Controller API](https://cloud.ibm.com/apidocs/resource-controller/resource-controller) to grab tenant id and region?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
```diff
@@ -11,16 +11,22 @@ def supported_auth_attributes
...
-      IBM::Cloud::SDK::PowerIaas.new(region, guid, token, crn, tenant)
-    else
-      raise ArgumentError, "Unknown target API set: '#{options[:service]}''"
-    end
+    token, power_iaas_service = self.class.raw_connect(auth_key, uid_ems)
+
+    require "ibm_cloud_power"
+
+    # TODO
+    region = "us-south"
```
jaywcarman, line 19: Same comment about the [IBM Cloud Resource Controller API](https://cloud.ibm.com/apidocs/resource-controller/resource-controller).

jaywcarman (commented):
This OpenAPI work is awesome enough to leave a review on Saturday evening :laughing: Nice work @agrare

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
agrare: Already done, I'm using it [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/48/files#diff-b636667a0f7f3e051ca4608c0ab24a4bR126) I'm just working on a clean way of returning it from connect.

I had started using the api_client instance as the return from connect on the vpc side because that's all that was needed to pass in to the other apis but I'll need to modify that for the power side

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
agrare: Yeah I was looking for this on the API docs site and didn't see anything, would like to know if this is something that will be added to the openapi spec in the future or were living dangerously using undocumented apis 😆

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
agrare: I think this actually is missing in the openapi spec as a property of the service instance, I see region_id in the response payload but it isn't making it to the deserialized object.  Just doing this until I figure out or fix-up the spec and regen the client

agrare:
I appreciate the weekend review @jaywcarman !

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
jaywcarman: A new API call should be added in the future to return storage availability (I'm guessing that it will be similar to [PCloudSystemPools](https://cloud.ibm.com/apidocs/power-cloud#pcloud-systempools-get)). The return data structure might change, this [/broker/v1/storage-types](https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/blob/master/lib/ibm/cloud/sdk/power_iaas.rb#L169) call could disappear, I'm not sure - so yeah this could be kind of risky. I can get more details next week.

If you think it's a bad idea to use this interim call we could just use a hard coded list of storage types. Not all types are available in all regions so it could cause some volume provisioning fails.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
jaywcarman: The region is embedded (as 'location') in the [CRN](https://cloud.ibm.com/docs/account?topic=account-crn), which is included in the openapi spec. We'd have to parse it out.

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
agrare: > this /broker/v1/storage-types call could disappear

Yeah this is the part that worries me, if we release with this and it goes away refreshes will fail (or at least we won't be pulling storage-types and provisions wouldn't work) until the customer upgraded.  If the intention is to deprecate this API but isn't completed by the time we release I think a hard coded list would be safest.

We don't need to solve this in this PR, I'm going to replace with a single RestClient call for now.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
agrare: :+1: Parsing the CRN to get the tenant_id and the region, both of those TODOs dropped now

app/models/manageiq/providers/ibm_cloud/inventory/collector.rb
```diff
@@ -1,4 +1,25 @@
...
+    collection_key ||= list_method.to_s.sub("list_", "")
+    result = api_client.send(list_method, *args, opts)
+
+    result.send(collection_key)&.each { |i| yield i }
+
+    next_link = result._next&.href if result.respond_to?(:_next)
+    return unless next_link
+
+    start = parse_next_link(next_link)["start"]&.first
+    all(api_client, list_method, args, collection_key, opts.merge(:start => start))
```
Fryguy, line 19: we want to avoid recursion only because it hurts the stack depth.  I think we could do all of this with a simple while loop.

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
```diff
@@ -3,63 +3,108 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::VPC < ManageIQ::Provi
...
+  end
+
+  def token
+    @token ||= manager.class.raw_connect(api_key)
+  end
+
+  def vpc
+    @vpc ||= begin
+      require "ibm_vpc"
+      service_url = "https://#{manager.provider_region}.iaas.cloud.ibm.com/v1"
```
Fryguy, line 82: I feel like it would be safer to use a URL generator here, just in case?

```ruby
URI::HTTPS.build(:host => "#{manager.provider_region}.iaas.cloud.ibm.com", :path => "/v1")
```

Fryguy:
Does https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/48/files#diff-7e8539b60953dd529dfe4556ed32f7fc27eeb909768f0084cea1c066229717fdR7 go away with the new VPC gem?

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
```diff
@@ -3,63 +3,108 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::VPC < ManageIQ::Provi
...
+      IbmVpc::VpcV1.new(:version => "2020-08-01", :authenticator => token, :service_url => service_url)
```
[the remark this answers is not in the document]
agrare: :+1: yeah I like that, will update

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
agrare: :+1:

agrare:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/137

PR #50, opened 2020-09-28
PowerVS: Use 'cloud_volume_types' table for storage types

jaywcarman:
Storing both system types and storage types in the 'flavors' table is
confusing. Storage types for 'CloudVolume' objects fits perfectly into
the 'cloud_volume_types' table.


agrare (approved):
:+1: this does make more sense, I forgot we added this for openstack v2v so sorry @Kuldip-Nanda for pointing you in the wrong direction

PR #54, opened 2020-09-29
Initial storage function for IBM Cloud Provider

blancett:


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
```diff
@@ -5,20 +5,31 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager < ManageIQ::Providers::Cl
...
+          :class_name  => "ManageIQ::Providers::IbmCloud::VPC::StorageManager",
+          :autosave    => true,
+          :dependent   => :destroy
   before_create :ensure_managers
   before_update :ensure_managers_zone
 
   def ensure_managers
     ensure_network_manager
     ensure_managers_zone
+    ensure_storage_manager
```
agrare, line 20: Let's move this up before `ensure_managers_zone` so that we can keep the storage_manager.zone_id in sync

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
```diff
@@ -5,20 +5,31 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager < ManageIQ::Providers::Cl
...
   end
 
+  def ensure_storage_manager
+    build_storage_manager(:type => 'ManageIQ::Providers::IbmCloud::VPC::StorageManager') unless storage_manager
+    storage_manager.name = "Storage-Manager of '#{name}'"
+  end
+
   def ensure_managers_zone
     network_manager.zone_id = zone_id if network_manager
   end
```
agrare, line 35: And then we can do `storage_manager.zone_id = zone_id if storage_manager`

agrare:
High level this looks great @blancett, will need VCRs to be re-recorded

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -18,6 +19,7 @@ def parse
     flavors
     images
     instances
+    volumes
```
agrare, line 22: Oh, you'll also need to define the `volumes` parser method

blancett:
will do - thanks for the review

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
agrare: :+1: nice

agrare:
FYI @blancett your commits are mapped to a different github user (https://github.com/sametimesolutions) you might want to make sure your `git config user.email` matches your @blancett github user email address

PR #55, opened 2020-09-29
Enable manageiq-provider-ibm_cloud logging

Kuldip-Nanda:
This is part of the change to enable login in ibmcloud.log.

Depends on https://github.com/ManageIQ/manageiq/pull/20624


jaywcarman (approved):

agrare:
@Kuldip-Nanda you'll want to add a setting for the log level in the `config/settings.yml` file here.  See https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/config/settings.yml#L102-L103 for an example

agrare (approved):

PR #57, opened 2020-09-30
Bound IBM machine types on PVS prov. to API returned values

iv1111:


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -19,8 +19,10 @@ def get_timezones(_options = {})
   end
 
   def allowed_sys_type(_options = {})
-    # TODO: replace with api provided values, once issue '114' is solved and merged
-    {0 => "s922", 1 => "e880"}
+    ar_ems = ar_ems_get
+    ar_sys_types = ar_ems&.flavors
+    sys_types = ar_sys_types&.map&.with_index { |sys_type, i| [i + 1, sys_type['name']] }
+    Hash[sys_types || {}]
```
agrare, line 25: You can use `each_with_index` to make this a bit simpler `Hash[ar_ems.flavors.to_a.each_with_index.map { |sys_type, i| [i, sys_type] }]` that way you don't need to increment `i` yourself

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -19,8 +19,10 @@ def get_timezones(_options = {})
   end
 
   def allowed_sys_type(_options = {})
-    # TODO: replace with api provided values, once issue '114' is solved and merged
-    {0 => "s922", 1 => "e880"}
+    ar_ems = ar_ems_get
+    ar_sys_types = ar_ems&.flavors
```
agrare, line 23: > ar_ems&.flavors

If we can't find the EMS I wonder if we should just raise an error instead of trying to handle it here, that would be catastrophic down the line anyway

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
iv1111: Hi, Adam. I think that _with_index_ supports that and _each_with_index_ expects no parameters (starts from '0'). I will change now accordingly.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
iv1111: i did the corresponding changes.

agrare (approved):

PR #65, opened 2020-10-01
PowerVS: Add Availability Zone

jaywcarman:
PowerVS resource availability is bound by the service instance (which
maps directly to a Cloud Manager ems). There's existing logic we can
re-use (e.g. cloud volume availability) if we create a single
availability zone for each Cloud Manager instance.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -18,6 +18,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
   end
 
   def parse
+    availability_zones
     images
     volumes
     instances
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -229,4 +232,12 @@ def cloud_volume_types
       )
     end
   end
+
+  def availability_zones
+    # Single availability zone per PowerVS Cloud Manager
```
jaywcarman, line 237: This looks a little strange, but _there can be only one_ availability zone per PowerVS service instance. At first we didn't implement it, but now I'm noticing that existing Cloud Volume logic makes use of it (and maybe elsewhere?).

agrare: @jaywcarman how about using the service guid aka `cloud_manager.uid_ems` for the az.ems_ref?  That way if someone changes the name of the provider we don't delete & re-create the AZ record.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -123,14 +125,15 @@ def images
...
       persister.cloud_volumes.build(
-        :ems_ref       => vol['volumeID'],
-        :name          => vol['name'],
-        :status        => vol['state'],
-        :bootable      => vol['bootable'],
-        :creation_time => vol['creationDate'],
-        :description   => 'IBM Cloud Block-Storage Volume',
-        :volume_type   => vol['diskType'],
-        :size          => vol['size']&.gigabytes
+        :availability_zone => persister.availability_zones.lazy_find(persister.cloud_manager.name),
```
agrare, line 128: Just need to update the lazy_find to match now

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -31,15 +32,16 @@ def instances
...
-        :description      => "IBM Cloud Server",
-        :ems_ref          => instance["pvmInstanceID"],
-        :flavor           => "",
-        :location         => "unknown",
-        :name             => instance["serverName"],
-        :vendor           => "ibm",
-        :connection_state => "connected",
-        :raw_power_state  => instance["status"],
-        :uid_ems          => instance["pvmInstanceID"]
+        :availability_zone => persister.availability_zones.lazy_find(persister.cloud_manager.name),
```
agrare, line 35: Same here

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
jaywcarman: Done. Nice catch!

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
jaywcarman: Does lazy_key off of uid_ems only?

agrare (approved):

PR #66, opened 2020-10-01
Enabled hard and soft reset to the VM instance

Kuldip-Nanda:


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -19,7 +20,14 @@ def raw_stop
 
   def raw_reboot_guest
     with_provider_connection(:service => 'PowerIaas') do |power_iaas|
-      power_iaas.reboot_pvm_instance(ems_ref)
+      power_iaas.soft_reboot_pvm_instance(ems_ref)
+    end
+    update!(:raw_power_state => "off")
```
agrare, line 25: For `raw_power_state` I think the options for power are `ACTIVE`/`INACTIVE` according to https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb#L33-L40
(I'm guessing on the `INACTIVE` @jaywcarman keep me honest here)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -19,7 +20,14 @@ def raw_stop
...
+  end
+
+  def raw_reset
+    with_provider_connection(:service => 'PowerIaas') do |power_iaas|
+      power_iaas.hard_reboot_pvm_instance(ems_ref)
     end
     update!(:raw_power_state => "off")
```
agrare, line 32: Is "off" the correct power state for `reset`?  If anything I think it'd be something like restarting or just still "ACTIVE"

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
jaywcarman: @agrare @Kuldip-Nanda  it's actually `ACTIVE` / `SHUTOFF` :shrug:

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
jaywcarman: You're right @agrare. I just fired up my `while true;` looper and the state remains `ACTIVE` during resets.

agrare:
@Kuldip-Nanda specs failures look related to your changes: https://travis-ci.com/github/ManageIQ/manageiq-providers-ibm_cloud/jobs/394281817#L2038-L2044

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
Kuldip-Nanda: thanks, just missed it.

jaywcarman (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
jaywcarman: For the record, the power states in PowerVS come from OpenStack:
https://docs.openstack.org/nova/latest/reference/vm-states.html

MIQ OpenStack provider power states.
https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/cloud_manager/vm/operations/power.rb

PR #67, opened 2020-10-01
PowerVS: Remove storage type collector

jaywcarman:
The underlying Power Cloud API call to /broker/v1/storage-types is
unsupported. Using a hard-coded list within the refresher allows us to
keep the rest of the back-end logic in place and hopefully facilitate an
easy drop in once the official storage type API call is added.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -44,6 +44,13 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
   end
 
   def storage_types
-    connection.get_storage_types
+    # TODO: The Power Cloud API does not yet have a call to retrieve
+    # available storage types.
+    [
+      {"description" => "Tier 1", "state" => "active", "type" => "tier1"},
+      {"description" => "Tier 3", "state" => "active", "type" => "tier3"},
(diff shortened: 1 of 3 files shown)


jaywcarman:
See this discussion for more context:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/48#discussion_r495601948

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -217,8 +217,17 @@ def systemtypes
   end
 
   def cloud_volume_types
-    # get only the active storage
-    collector.storage_types.each do |v|
+    # TODO: The Power Cloud API does not yet have a call to retrieve
+    # available storage types. When availabile this hardcoded list should
+    # be removed and the list should come from collector.storage_types
+    all_possible_storage_types =
```
agrare, line 223: @jaywcarman could you return this list from the [collector](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/67/files#diff-53f571cf57b42afefdd9b8a927cff8beR47) so the parser can stay the same when we switch over to using an API call?

jaywcarman: @agrare I can definitely do that, but full disclosure: I have not been able to get a guarantee that this data structure will stay intact.

agrare (approved):

PR #68, opened 2020-10-01
PowerVS: Update registration auth_key label

jaywcarman:
For now users must input an API key each time they register a new
provider. The "(if not using an existing provider)" text can be added
back once we implement a top-level provider.


agrare (approved):
:+1: good catch

PR #75, opened 2020-10-07
adding refresher logic

Real-Omar-Afifi:


agrare (approved):

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresh_worker/runner.rb
```diff
@@ -0,0 +1,9 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::RefreshWorker::Runner < ManageIQ::Providers::BaseManager::RefreshWorker::Runner
+  def do_before_work_loop
+    @emss.select { |ems| ems.kind_of?(EmsCloud) }.each do |ems|
```
agrare, line 3: Now that we're not combining queue names I think we can go back to a single parent EMS in the RefreshWorker so we might be able to drop this going forward

PR #77, opened 2020-10-07
Update PowerVS cloud subnet parser

jaywcarman:
- Add AZ relationship
- Save network type to 'extra_attributes' column

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -161,15 +161,17 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
       )
 
       persister_cloud_subnet = persister.cloud_subnets.build(
-        :cloud_network    => persister_cloud_networks,
-        :cidr             => network['cidr'],
-        :ems_ref          => network['networkID'],
-        :gateway          => network['gateway'],
-        :name             => network['name'],
-        :status           => "active",
-        :dns_nameservers  => network['dnsServers'],
-        :ip_version       => '4',
-        :network_protocol => 'IPv4'
+        :cloud_network     => persister_cloud_networks,
+        :cidr              => network['cidr'],
+        :ems_ref           => network['networkID'],
+        :gateway           => network['gateway'],
+        :name              => network['name'],
+        :status            => "active",
+        :dns_nameservers   => network['dnsServers'],
+        :ip_version        => '4',
+        :network_protocol  => 'IPv4',
+        :availability_zone => persister.availability_zones.lazy_find(persister.cloud_manager.uid_ems),
+        :network_type      => network['type']
       )
 
       mac_to_port = {}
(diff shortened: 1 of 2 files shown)


agrare:
Looks good @jaywcarman can you add a check in the `assert_specific_cloud_subnet` spec?  Something like `expect(cloud_subnet.availability_zone&.ems_ref).to eq(ems.uid_ems)`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -161,15 +161,17 @@ def networks
...
+        :cidr              => network['cidr'],
+        :ems_ref           => network['networkID'],
+        :gateway           => network['gateway'],
+        :name              => network['name'],
+        :status            => "active",
+        :dns_nameservers   => network['dnsServers'],
+        :ip_version        => '4',
+        :network_protocol  => 'IPv4',
+        :availability_zone => persister.availability_zones.lazy_find(persister.cloud_manager.uid_ems),
+        :extra_attributes  => network['type']
```
jaywcarman, line 174: @agrare Should I store this as a hash `{'type' => network['type']}` ?

We weren't using it before, but this is in our current refresher spec:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb#L152

jaywcarman: Do any extra attributes passed to [`persister.cloud_subnets.build`](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb#L163) automatically get added to the `extra_attributes` `cloud_subnets` `extra_attributes` DB column?

agrare: That's interesting, wasn't aware of this but we define some accessor methods that use extra_attributes as the backing: https://github.com/ManageIQ/manageiq/blob/master/app/models/cloud_subnet.rb#L35-L43
so `cloud_subnet.ip_version=` gets saved to `extra_attributes`

In this case maybe it is best to define a `network_type=` which saves it to `extra_attribtues`

I learn something new in this huge product everyday

jaywcarman: :) I know the feeling, that for sure! I was very confused as to how the specs were passing without any existing `extra_attributes`, haha.

I agree defining `cloud_subnet.network_type=` so we can keep `ip_version`. I've noticed the providers are inconsistent on using `IPv4` vs `ipv4` for network protocol so I could see it being useful.

jaywcarman: https://github.com/ManageIQ/manageiq/pull/20672

jaywcarman:
Cross repo test with https://github.com/ManageIQ/manageiq/pull/20672 is passing:
https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/188

agrare (approved):
:+1: awesome @jaywcarman !

PR #80, opened 2020-10-08
adding bitness, and linking security groups to instances

Real-Omar-Afifi:


Real-Omar-Afifi:
@agrare I was thinking of adding a specific test for a security group to assert the count of network ports, but I felt that it might be an overkill for this specific case. Lemme know if you want me to add it.

agrare (approved):

PR #81, opened 2020-10-08
Automatically replace the Bearer auth expiration timestamp

agrare:
The ibm-cloud-sdk gem automatically tries to renew an expired Bearer token which causes VCR UnhandledHTTPInteractions exceptions 1 hour after the vcr cassette is recorded.

We can automatically replace the expiration timestamp with one way in the future to mitigate this.


agrare:
cc @Fryguy not sure if you know of a slicker way of doing this

spec/spec_helper.rb
```diff
@@ -11,4 +11,15 @@
...
   config.ignore_hosts 'codeclimate.com' if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
+  config.before_record do |i|
+    # The ibm-cloud-sdk gem attempts to auto-renew the Bearer token if it
+    # detects that it is expired.  This causes unhandled http interactions
+    # after the expiration time.  We can replace the expiration time with
+    # one way in the future to prevent this.
+    if i.request.uri == "https://iam.cloud.ibm.com/identity/token"
+      body = JSON.parse(i.response.body)
+      body["expiration"] = 4102444800
```
jaywcarman, line 21: Seems like an acceptable stopgap for the next 80 years :smile:

jaywcarman (approved):

Fryguy:
Typically the way you do this is to change the matching criteria for VCR to not include that header.  For example, https://github.com/ManageIQ/manageiq-providers-azure/blob/e48d031f4822fdf0a9b7b2b9f22b2cb471a393f1/spec/spec_helper.rb#L14-L19

Fryguy:
Though now that I'm reading deeper, it seems we essentially get "extra" calls, which I'm not sure you can ignore with VCR...yeah I think settings a future time is probably good, unless may you can support "nil" time?

agrare:
@Fryguy the problem isn't that VCR isn't matching the request, it is that our sdk gem tries to renew the bearer token

agrare:
Yeah if we update the ibm-cloud-sdk gem to handle `nil` expiration we can just delete they key instead of setting it to a future time

Fryguy:
You might be able to tell VCR to ignore recording specific requests

agrare:
Is ignoring whole requests better?  This seems like a more targeted fix.

PR #82, opened 2020-10-09
PowerVS: Enforce single volume type in provisioning forms

jaywcarman:
PowerVS VMs require all attached volumes to be the same type. This PR removes unnecessary storage type fields in the VM provisioning 'New Volume' tab, and sets all new volume types to the selection made in the 'Profiles' tab.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -42,7 +42,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
         :vendor            => "ibm",
         :connection_state  => "connected",
         :raw_power_state   => instance["status"],
-        :uid_ems           => instance["pvmInstanceID"]
+        :uid_ems           => instance["pvmInstanceID"],
+        :format            => instance["storageType"]
       )
 
       # saving hardware information (CPU, Memory, etc.)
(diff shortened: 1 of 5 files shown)


jaywcarman:
The 'power_virtual_servers' specs are passing locally. Once the VPC specs get cleaned up this _should_ re-build green.

iv1111:
Hi. I've tested the code now.

If I open the provisioning workflow and set some values **except** system type and storage type I see this when submitting the request although the UI shows some default values set in the fields.

image

It is as if the default preset values are not propagated to the back-end.

@jaywcarman Could you test and confirm?

iv1111:
I think I know why the default value is not selected in the back-end. It think it is because with_index(1) starts in the corresp. methods indexing items from 1, whereas the YAML file declares default index as '0', which just doesn't exist in your case.

iv1111:
```
diff --git a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
index dcbd26f..a51f674 100644
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -24,13 +24,13 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
 
     ar_sys_types = ar_ems.flavors
-    sys_types = ar_sys_types&.map&.with_index(1) { |sys_type, i| [i, sys_type['name']] }
+    sys_types = ar_sys_types&.map&.each_with_index { |sys_type, i| [i, sys_type['name']] }
     Hash[sys_types || {}]
   end
 
   def allowed_storage_type(_options = {})
     ar_storage_types = ar_ems.cloud_volume_types
-    storage_types = ar_storage_types&.map&.with_index(1) { |storage_type, i| [i, storage_type['name']] }
+    storage_types = ar_storage_types&.map&.each_with_index { |storage_type, i| [i, storage_type['name']] }
     Hash[storage_types || none]
   end

``` 
This will solve the problem. Not sure how to put it into your branch directly.

jaywcarman:
Great catch @iv1111 - thanks!

agrare (approved):

agrare:
@jaywcarman can you open a schema PR to remove the "deprecated" description from `vms.format` ?

PR #83, opened 2020-10-09
Fix VPC CloudManager verify_credentials

agrare:
I noticed this when I deleted the `cloud_manager.yml` vcr cassette and the specs were passing without any VCR at all :laughing: 

VPC verify_credentials (class and instance levels) weren't actually issuing any API calls that would check if the credentials are valid or not.

Updated to fetch the authorization_header from the IAM token which does issue an API call to the IAM token service.


agrare:
cc @jaywcarman

jaywcarman (approved):

PR #84, opened 2020-10-12
Declared DDF schema for cloud volume creation and editing

skateman:
This is an initial Cloud Volume form schema for the fields that are specific to the ibm-cloud-provided cloud volumes.

@miq-bot add_reviewer @agrare

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7334

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -7,6 +7,100 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
     availability_zone.vms.select { |vm| vm.format == volume_type }
   end
 
+  def self.params_for_create(ems)
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'size',
+          :id         => 'size',
+          :label      => _('Size (in bytes)'),
+          :type       => 'number',
+          :step       => 1024 * 1024 * 1024,
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+        },
+        {
+          :component => 'select',
+          :name      => 'volume_type',
+          :id        => 'volume_type',
+          :label     => _('Cloud Volume Type'),
+          :condition => {
+            :when => 'edit',
+            :is   => false,
+          },
+          :options   => ems.cloud_volume_types.map do |cvt|
+            {
+              :label => cvt.description,
+              :value => cvt.name,
+            }
+          end,
+        },
+        {
+          :component => 'switch',
+          :name      => 'multi_attachment',
+          :id        => 'multi_attachment',
+          :label     => _('Shareable'),
+          :onText    => _('Yes'),
+          :offText   => _('No'),
+        },
+        {
+          :component    => 'select',
+          :name         => 'affinity_policy',
+          :id           => 'affinity_policy',
+          :label        => _('Affinity Policy'),
+          :initialValue => 'off',
+          :condition    => {
+            :when => 'edit',
+            :is   => false,
+          },
+          :options      => [
+            {
+              :label => 'Off',
+              :value => 'off',
+            },
+            {
+              :label => 'Affinity',
+              :value => 'affinity',
+            },
+            {
+              :label => 'Anti-affinity',
+              :value => 'anti-affinity',
+            },
+          ],
+        },
+        {
+          :component  => 'select',
+          :name       => 'affinity_volume_id',
+          :id         => 'affinity_volume_id',
+          :label      => _('Affinity Volume'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+          :condition  => {
+            :and => [
+              {
+                :not => {
+                  :when => 'affinity_policy',
+                  :is   => 'off',
+                },
+              },
+              {
+                :when => 'edit',
+                :is   => false,
+              },
+            ],
+          },
+          :options    => ems.cloud_volumes.map do |cv|
+            {
+              :value => cv.id,
+              :label => cv.name,
+            }
+          end,
+        },
+      ],
+    }
+  end
+
   def self.validate_create_volume(ext_management_system)
     validate_volume(ext_management_system)
   end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager.rb
@@ -27,6 +27,9 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager < Manag
   virtual_delegate :cloud_tenants, :to => :parent_manager, :allow_nil => true
   virtual_delegate :volume_availability_zones, :to => :parent_manager, :allow_nil => true
 
+  supports :cloud_volume
+  supports :cloud_volume_create
+
   def image_name
     "ibm"
   end


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,73 @@ def available_vms
...
+              :value => 'anti-affinity'
+            }
+          ]
+        },
+        {
+          :component => 'select',
+          :name      => 'volume_type',
+          :id        => 'volume_type',
+          :label     => _('Cloud Volume Type'),
+          :options   => self.class.CLOUD_VOLUME_TYPES, # TODO: the types are in the angular controller, need to extract them
```
skateman, line 47: @agrare @jaywcarman I am not aware of the cloud volume types for this provider, does it have any? [This](https://github.com/ManageIQ/manageiq-providers-amazon/pull/657/files#diff-8f332bdbb8551a6a7e20fb13ad23aa04R5-R11) is what I have for amazon.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,73 @@ def available_vms
...
+          :component => 'switch',
+          :name      => 'shareable',
+          :id        => 'shareable',
+          :label     => _('Shareable'),
+          :onText    => _('Yes'),
+          :offText   => _('No')
+        },
+        {
+          :component    => 'select',
+          :name         => 'affinity_policy',
```
skateman, line 23: This is not a valid field in the database, and I guess the API would not be able to deal with this either. Where is this information getting stored?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,73 @@ def available_vms
...
+          :options   => self.class.CLOUD_VOLUME_TYPES, # TODO: the types are in the angular controller, need to extract them
+          :condition => {
+            :when => 'affinity_policy',
+            :is   => 'off'
+          },
+          :validate  => [{:type => 'required'}]
+        },
+        {
+          :component => 'select',
+          :name      => 'affinity_volume_id',
```
skateman, line 56: Same here, I don't see this anywhere in the DB...

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
jaywcarman: We're using the `cloud_volume_types` table. [OpenStack does the same](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/inventory/parser/storage_manager/cinder_manager.rb#L65).

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
jaywcarman: We deferred affinity controls/inventory info in the current release. Perhaps we'll need to add a column to `cloud_volumes` next release?

I think the options will always be `['none', 'affinity', 'anti-affinity']`. Can we just hard-code these values into the schema here?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
jaywcarman: All of the affinity controls/info will be a new feature for us.

skateman: Okay, I'm going to omit this from the schema and it should be re-added when the feature is implemented...

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
skateman: If the column is not in the database, this form will just not work, so I'm removing this field until then...

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
skateman: As far as I saw, the OpenStack form doesn't expose the `volume_type` field in its form.

jaywcarman: There is a "Cloud Volume Type" field for cinder volumes, and I think the volumes come from the `cloud_volume_types` table:

https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/cloud_volume/_common_new_edit.html.haml#L81-L92
image

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
jaywcarman: I'm OK with that plan - remove it and I'll add it back later when I implement affinity controls.

I don't fully understand why each field needs to be associated with a DB column, though. Would it not be possible to pass some general parameter to our IBM Cloud API without persisting anything to the DB?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
skateman: Okay, my bad, I see it has a `CloudVolumeType` model, will update this accordingly.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,73 @@ def available_vms
     availability_zone.vms
   end
 
+  def self.params_for_create(ems_id)
+    {
+      :fields => [
+        {
+          :component => 'switch',
+          :name      => 'shareable',
```
skateman, line 15: Is this field backed by a DB column?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
skateman: @agrare this question is for you I guess

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
jaywcarman: Shareable here just means that the volume can be attached to more than one VM, so I _think_ the `multi_attachment` column fits.

skateman: @agrare can you confirm this?

agrare: Openstack Cinder refresh sets `multi_attachment`, should the powervs parser also collect&set this property?
Also could we match the column name with the DDF forms?

jaywcarman: > should the powervs parser also collect&set this property?
Yes - I can submit a PR for this

> Also could we match the column name with the DDF forms?
Yes, but I'd like to keep the label as `Shareable` to match the PowerVS web interface.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
agrare: I think we should try to keep the params similar to the db columns if they exist (e.g. shareable vs multi_attachment above) but if it doesn't exist yet there isn't a requirement that every param match a db column for operations.

agrare: We just couldn't reuse these same params to then implement an "edit" form

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
agrare: :+1:

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
agrare: @skateman we're not create a cloud_volume database record, so not sure why it wouldn't work if the column doesn't exist?

skateman: Wait, this form isn't creating a DB record?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
skateman: The label can stay `shareable` but the name/id should be matching the DB colum, which I changed...

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
agrare: I assume this is the form that collects params that ends up being sent to `CloudVolume.create_volume_queue`

@jaywcarman ?

skateman: Oh, in that case fields can be there, but then I need to hide them in edit mode...

jaywcarman: Correct - the form form data will go to `CloudVolume.create_volume_queue` where we make the IBM Cloud API call.

I like the idea of matching this form to the inventory picked up later by refresh, but yeah I was scratching my head on the hard requirement. I think everything makes sense now :)

skateman: I will put back the removed fields then...thanks for the clarification.

skateman:
@jaywcarman I need to know what fields are available for changes when editing a CloudVolume on top of name and size (assuming these are editable).

jaywcarman:
@skateman
> I need to know what fields are available for changes when editing a CloudVolume on top of name and size (assuming these are editable).

The 'multi_attachment' (aka Shareable) field should be included in the edit form.

Our API also allows updating the 'bootable' flag ([apidocs](https://cloud.ibm.com/apidocs/power-cloud#pcloud-cloudinstances-volumes-post)), but that's not included in the create form. I'm going to look into this and will report back.

skateman:
> The 'multi_attachment' (aka Shareable) field should be included in the edit form.

Okay I will add it, but this doesn't answer my question. From the fields specified in the schema, which should be changeable  when you're editing a CloudVolume (not creating).

skateman:
@jaywcarman is it correct now?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,100 @@ def available_vms
...
+          :id        => 'volume_type',
+          :label     => _('Cloud Volume Type'),
+          :condition => {
+            :when => 'edit',
+            :is   => false,
+          },
+          :options   => ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolumeType.where(:ext_management_system_id => ems_id).map do |cvt|
+            {
+              :label => cvt.name,
+              :value => cvt.value,
```
jaywcarman, line 35: I'm trying to understand these - does `cvt.value` get set from the `cvt.name` value?

As I look more closely at cloud_volume_types table I'm thinking it would be best to set:
```
:label => cvt.description,
:value => cvt.name,
```

skateman: Yeah, the `value` is incorrect, it should be `type` instead...at least compared with OpenStack.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -7,6 +7,100 @@ def available_vms
...
+          :id         => 'affinity_volume_id',
+          :label      => _('Affinity Volume'),
+          :isRequired => true,
+          :validate   => [{:type => 'required'}],
+          :condition  => {
+            :and => [
+              {
+                :not => {
+                  :when => 'affinity_policy',
+                  :is   => 'affinity',
```
jaywcarman, line 84: Found an error in the original schema I gave you @skateman 

This should be: `:is   => 'off',`. An 'affinity volume' is only applicable when `affinity` or `anti-affinity` was selected.

skateman: :wrench: :hourglass:

jaywcarman:
I'm testing this out now, and am seeing something strange with the volume_type select field:

After making a selection if I open the pull down again it shows all items as selected:
image

Not sure if it's relevant, but my browser version is Chrome 86.0.4240.183 (Official Build) (64-bit)

jaywcarman:
Another test observation - the 'size' field is not showing up:
image

jaywcarman:
I tested this and everything LGTM.

PR #85, opened 2020-10-12
Set Cloud Volume 'multi-attachment' column

jaywcarman:
PowerVS volumes can optionally be shareable (able to attach to multiple
VMs). This attribute fits in the 'cloud_volumes' table,
'multi-attachment' column and thus should be collected and persisted
during refresh.


agrare:
@jaywcarman can you add an `assert_specific_cloud_volume` spec to the power refresher specs that shows this column being set?

jaywcarman:
> can you add an assert_specific_cloud_volume spec to the power refresher specs that shows this column being set?

@agrare I added a test (and am working on getting specs into my initial thought processes, thanks for the patience).

agrare (approved):

PR #86, opened 2020-10-14
Filter existing volume types in provisioning form

jaywcarman:
During VM provisioning existing volumes can be selected to be attached
to the new VM. The selection list can be filtered to show only volumes
available for attachment. Volumes are available if all of the following
conditions are true:

- Volume type matches the selected VM storage type
- Volume is shareable ('multi_attachment' == true)
- Volume status is 'available'

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -48,8 +48,15 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
   end
 
   def allowed_cloud_volumes(_options = {})
-    ar_volumes = ar_ems.cloud_volumes
+    storage_type = values&.dig(:storage_type, 1)
+
+    ar_volumes = ar_ems.cloud_volumes.select do |cloud_volume|
+      cloud_volume['volume_type'] == storage_type &&
+        (cloud_volume['multi_attachment'] || cloud_volume['status'] == 'available')
+    end
+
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -48,8 +48,15 @@ def allowed_subnets(_options = {})
   end
 
   def allowed_cloud_volumes(_options = {})
-    ar_volumes = ar_ems.cloud_volumes
+    storage_type = values&.dig(:storage_type, 1)
+
+    ar_volumes = ar_ems.cloud_volumes.select do |cloud_volume|
+      ((cloud_volume['volume_type'] == storage_type) &&
+      (cloud_volume['multi_attachment'] || cloud_volume['status'] == 'available'))
```
agrare, line 55: Couple extra parens here, can just be `cv['volume_type'] == storage_type && (cv['mutli_attachment'] || cv['status'] == 'available')`

jaywcarman: Is it correct that Ruby doesn't require parenthesis for multi-line statements if you put the operator at the end of the line to be continued? So this is evaluated as a single statement:
```
cloud_volume['volume_type'] == storage_type &&
(cloud_volume['multi_attachment'] || cloud_volume['status'] == 'available')
```

agrare: Correct ruby does not require parens around a multi-line conditional

agrare (approved):

PR #87, opened 2020-10-14
Filter VMs by 'volume_type' in volume attach

jaywcarman:
Volume 'volume_type' must match VM 'format' for the Power Virtual Server
backend to allow an attach operation.


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -4,7 +4,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
   supports_not :update
 
   def available_vms
-    availability_zone.vms
+    availability_zone.vms.select { |vm| vm.attributes['format'] == attributes['volume_type'] }
```
agrare, line 7: You don't need to use `attributes[]` for something which is a column on the model since ActiveRecord auto-defines accessors for them.  So just `vm.format == volume_type` should work.

agrare (approved):

PR #90, opened 2020-10-15
Apply gettext for string internationalization

jaywcarman:


content/miq_dialogs/miq_provision_ibm_dialogs_template.yaml
```diff
@@ -16,7 +16,7 @@
       :description: Customization
       :fields:
         :ip_addr:
-          :description: Specify an IPv4 address
+          :description: Specify an IPv4 address  #TODO: Apply Gettext?
```
jaywcarman, line 19: Is there a way to apply gettext to these yaml templates? I couldn't find any examples from templates defined in other MIQ providers.

Fryguy: This happens automatically in core.

Fryguy:
@mzazrivec can you review as well?

content/miq_dialogs/miq_provision_ibm_dialogs_template.yaml
jaywcarman: Coolness, thanks @Fryguy. I removed the `#TODO`.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -34,10 +34,10 @@ def instances
       # saving general VMI information
       ps_vmi = persister.vms.build(
         :availability_zone => persister.availability_zones.lazy_find(persister.cloud_manager.uid_ems),
-        :description       => "PVM Instance",
+        :description       => N_("PVM Instance"),
```
mzazrivec, line 37: I'm having bit of a problem to understand why is `N_()` used here. Is the `:description` string then passed to `_()` at some point later?

jaywcarman: @mzazrivec, I spent some more time understanding gettext and searching through the code - I don't think `:description` ever does get passed to `_()`. I'll update the PR soon to replace `N_()` with `_()`.

I was following the example of some existing lines in this module, which I now think are also incorrectly calling `N_()` instead of `_()`:
```
|powervs_internationalization|14340f9| manageiq-providers-ibm_cloud > ack "N_" app/models/manageiq/providers/ibm_cloud                                          
app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
37:        :description       => N_("PVM Instance"),
40:        :location          => N_("unknown"),
92:        :display_name => N_('Entitled Processors'),
93:        :description  => N_('The number of entitled processors assigned to the VM'),
102:        :display_name => N_('Processor type'),
103:        :description  => N_('dedicated: Dedicated, shared: Uncapped shared, capped: Capped shared'),
148:        :description       => N_('IBM Cloud Block-Storage Volume'),
```

mzazrivec: Yup. I think `_()` should be used here. Thanks!

agrare (approved):

PR #91, opened 2020-10-20
mapping VM labels

Real-Omar-Afifi:


app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -133,6 +135,41 @@ def hardware_networks(persister_hardware, instance)
     end
   end
 
+  def vm_and_template_labels(resource, tags)
+    tags.each do |tag|
+      formatted_tag = tag[:name].split(":")
+      tag_key = tag[:key] || formatted_tag[0] if formatted_tag.length >= 1
+      tag_value = tag[:value] || formatted_tag[1] if formatted_tag.length >= 2
```
agrare, line 142: This looks identical to the logic below https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/91/files#diff-be65168162a51f26d1c4e78ad610a312bea5fbefd1cff9fb19152ef1e89b9992R159-R161

Maybe pull this out into a helper method like `parse_tag_key_value`?

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -133,6 +135,41 @@ def hardware_networks(persister_hardware, instance)
     end
   end
 
+  def vm_and_template_labels(resource, tags)
+    tags.each do |tag|
+      formatted_tag = tag[:name].split(":")
+      tag_key = tag[:key] || formatted_tag[0] if formatted_tag.length >= 1
```
agrare, line 141: This conditional isn't doing what you think, if `formatted_tag.length < 1` then `tag_key` will be `nil` because the if is applying to the whole line not just after the `||`

It is equivalent to saying
```
if formatted_tag.length >= 1
  tag_key = tag[:key] || formatted_tag[0]
end
```

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -133,6 +135,41 @@ def hardware_networks(persister_hardware, instance)
     end
   end
 
+  def vm_and_template_labels(resource, tags)
+    tags.each do |tag|
+      formatted_tag = tag[:name].split(":")
```
agrare, line 140: Probably the best way to handle this is to do something like:
```
formatted_key, formatted_value = tag[:name]&.split(":")
key = tag[:key] || formatted_key
val = tag[:value] || formatted_value
```

Real-Omar-Afifi:
@agrare fixed

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -60,6 +60,8 @@ def instances
       instance_hardware(persister_instance, instance)
       instance_operating_system(persister_instance, instance)
       instance_network_interfaces(persister_instance, instance)
+      vm_and_template_labels(persister_instance, instance.tags.to_a)
+      vm_and_template_taggings(persister_instance, map_labels("VmIBM", instance.tags.to_a))
```
agrare, line 64: So I believe you need to add `VmIBM` to https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/ops_controller/settings/label_tag_mapping.rb#L17-L47

agrare (approved):

PR #107, opened 2020-10-27
VPC Power states

jaredm-ibm:
* Dependent on https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/pull/43 being merged.
* Add start, stop, reboot actions.
* VPC doesn't have pause so using start.
* Logic for reset is the same as reboot with an added parameter so added optional keyword parameter to raw_reboot_guest and reused the logic. The keyword shouldn't affect the internals of the automated methods as they usually don't used parameters.
* Added the $ibm_cloud_log to the CloudSDK init, so logging should now flow between the two and the stdout logger will be suppressed.
* The rescues are to keep the cloud log in synch with what is happening. I'm reraising the same error so that the evm log will log the correct traceback. I don't know if there is a better way to handle the errors.
* Need to write spec tests before merge.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
@@ -21,4 +21,36 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
   def self.display_name(number = 1)
     n_('Instance (IBM)', 'Instances (IBM)', number)
   end
+
+  # Send a start action to IBM Cloud. Wait for state to change to started, then update the raw_power_state to the current instance state.
+  def raw_start
+    with_provider_connection() do |vpc|
+      instance = vpc.instances.instance(ems_ref)
+      instance.actions.start
+      _log.info("Starting instance #{ems_ref} in state #{instance.status}.")
+      instance.wait_for(started: true)
+      update!(:raw_power_state => instance.status)
+      _log.info("Started instance #{ems_ref} in state #{instance.status}.")
+    end
+  end
+
+  # IBM Cloud does not support suspend. Hiding it from UI.
+  supports_not :suspend
+
+  # Send a stop action to IBM Cloud. Wait for state to change to stopped, then update the raw_power_state to the current instance state.
+  def raw_stop
+    with_provider_connection() do |vpc|
+      instance = vpc.instances.instance(ems_ref)
+      instance.actions.stop
+      _log.info("Stopping instance #{ems_ref} in state #{instance.status}.")
+      instance.wait_for(started: false)
+      update!(:raw_power_state => instance.status)
+      _log.info("Stopped instance #{ems_ref} in state #{instance.status}.")
+    end
+  end
+
+  # IBM Cloud does not support pause. Using stop since can't unsupport it.
+  def raw_pause
+    raw_stop
+  end
 end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
@@ -74,7 +74,7 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
   # @param force [Boolean] Ungracefully reboot VM.
   def raw_reboot_guest(force: false)
     with_provider_object do |instance|
-      instance.actions.reboot(force)
+      instance.actions.reboot(:force => force)
       sleep 5 # Sleep for 5 seconds to allow for reboot sequence to start.
       instance.wait_for! do
         sdk_update_status(instance)
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
@@ -27,8 +27,8 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
     with_provider_connection do |vpc|
       instance = vpc.instances.instance(ems_ref)
       instance.actions.start
-      instance.wait_for do
-        update!(:raw_power_state => instance.status)
+      instance.wait_for! do
+        sdk_update_status(instance)
         instance.started?
       end
     end
@@ -46,8 +46,8 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
     with_provider_connection do |vpc|
       instance = vpc.instances.instance(ems_ref)
       instance.actions.stop
-      instance.wait_for do
-        update!(:raw_power_state => instance.status)
+      instance.wait_for! do
+        sdk_update_status(instance)
         instance.stopped?
       end
     end
@@ -67,12 +67,15 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
     unsupported_reason_add(:reboot_guest, _('The VM is not powered on')) unless current_state == 'on'
   end
 
+  # Gracefully reboot the quest.
+  # @param force [Boolean] Ungracefully reboot VM.
   def raw_reboot_guest(force: false)
     with_provider_connection do |vpc|
       instance = vpc.instances.instance(ems_ref)
       instance.actions.reboot(force)
-      instance.wait_for do
-        update!(:raw_power_state => instance.status)
+      sleep 5 # Sleep for 5 seconds to allow for reboot sequence to start.
+      instance.wait_for! do
+        sdk_update_status(instance)
         instance.started?
       end
     end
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -21,4 +21,87 @@ def self.calculate_power_state(raw_power_state)
   def self.display_name(number = 1)
     n_('Instance (IBM)', 'Instances (IBM)', number)
   end
+
+  # Send a start action to IBM Cloud. Wait for state to change to started, then update the raw_power_state.
+  def raw_start
+    with_provider_connection do |vpc|
+      instance = vpc.instances.instance(ems_ref)
```
agrare, line 28: We have a `with_provider_object` method [[ref]](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/provider_object_mixin.rb#L10) which is intended to be used for things like this.

You can define a `def provider_obect(vpc)` that returns `vpc.instances.instance(ems_ref)`

jaredm-ibm: Like this?

```
  def provider_object(vpc)
    vpc.instances.instance(ems_ref)
  end

  # Send a start action to IBM Cloud. Wait for state to change to started, then update the raw_power_state.
  def raw_start
    with_provider_object do |instance|
      instance.actions.start
      ....
    end
  rescue
    .... logging
  end
```

agrare: :+1: exactly

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -21,4 +21,89 @@ def self.calculate_power_state(raw_power_state)
...
+        instance.stopped?
+      end
+    end
+  rescue => e
+    $ibm_cloud_log.error(e.to_s)
+    $ibm_cloud_log.log_backtrace(e)
+    raise
+  end
+
+  # IBM Cloud does not support pause. Using stop since can't hide it in UI.
```
agrare, line 63: We have suspend and pause?  But only suspend can be hidden?

That sounds like a bug, I'm good with this for now (or you could raise NotImplementedError up to you) and we can remove it later.

IMO suspend and pause are the same thing

agrare: @jaredm-ibm try `def raw_pause; validate_unsupported(_("Pause Operation")); end`

jaredm-ibm: I'll have to look again. I looked at the support types and didn't see it, however my miq repo may have been out of date.

jaredm-ibm:
@agrare Is there a way to tell codeclimate that the blocks it is flagging are not the same?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
jaredm-ibm: Yeah, I found :suspend in the queryable features but nothing about pause. The UI has pause without enabling it so I'm guessing that it can't be hidden.

jaredm-ibm: Sorry didn't see your potential fix. It doesn't change the UI. A started VM still has pause as an option.

agrare: Hm okay, that code is what the Google provider currently does so I guess that's not working either.  I opened and issue to track this.

agrare:
We can just ignore those code climate issues

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -21,4 +21,89 @@ def self.calculate_power_state(raw_power_state)
...
+  supports :reboot_guest do
+    unsupported_reason_add(:reboot_guest, _('The VM is not powered on')) unless current_state == 'on'
+  end
+
+  # Gracefully reboot the quest.
+  # @param force [Boolean] Ungracefully reboot VM.
+  def raw_reboot_guest(force: false)
+    with_provider_object do |instance|
+      instance.actions.reboot(force)
+      sleep 5 # Sleep for 5 seconds to allow for reboot sequence to start.
```
agrare, line 78: This is unfortunate but I get what you're trying to avoid here, is there any chance that starting a graceful shutdown might take longer than 5 seconds?

jaredm-ibm: Right, I agree. I had a wait_for transitional? but the VMs I'm testing again reboot way faster than the wait_for loop allow even with a 0.5 second sleep. It was for this purpose that I decided now was the time to implement the block. However when I tried testing my code the vm got stuck in the transitional loop. I guess we could do a wait for transitional in a begin with a 20 second timeout and just ignore the exception on timeout.

jaredm-ibm: Either we do a begin and ignore the exception, which is bad style, or we open up the SDK again and implement the tuple wait_for. It wouldn't be hard.

```
def wait_for(sleep_time: 5, timeout: 600, &block)
  @logger.info("Starting wait for instance #{id}. Starts in state #{status}.")
  loop do
    refresh
    return (false, "VM #{id} is in a failed state.") if failed?
    break if block.call(self)

    timeout = sleep_counter(sleep_time, timeout)
    return (false, "Time out while waiting #{id} to be stable.") if timeout <= 0
  end
  @logger.info("Finished wait for instance #{id}. Ends in state #{status}.")
  return (true, 'ok')
end

def wait_for!(sleep_time: 5, timeout: 600, &block)
  status, msg = wait_for(sleep_time: sleep_time, timeout: timeout, &block)
  return if status
  raise msg
end
```

jaredm-ibm: As for your initial question. The 5 second wait is to allow the VM to go into starting, stopping or some other state. The wait_for without a sleep will just see the started vm and return, which is a null operation.

agrare: Right I assumed it was only in the reboot operation because you're waiting for the VM to be started and if you check too quickly the vm will still be running and hasn't started to reboot yet.

agrare: Worst case if we return before a VM has started rebooting, we still issued the API call successfully.  I'd have to check if any other providers wait for the Vm to fully restart before returning.  I know VMware doesn't because it isn't an operation that returns a task.

Returning before the operation has completed is fine as long as it was initiated successfully

jaredm-ibm: Ok for the time being we can keep the sleep. I need to get some specs written for this because I keep testing in the UI, changing something and forgetting to test again.

jaredm-ibm:
With a passing spec, this item is code complete.

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'logger'
+
+class ParentClass
```
agrare, line 3: We shouldn't declare a class in the global context, I think you can accomplish this with an rspec mock
```
vpc = double("IBM::Cloud::SDK::Vpc")
allow(vpc).to receive(:logger).and_return(Logger.new(nil))
allow(vpc).to receive(:url)
etc...
```

jaredm-ibm: Ahh yes, hadn't thought of that. I'll try it.

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    let(:parent) do
+      vpc = double("IBM::Cloud::SDK::Vpc")
+      allow(vpc).to receive(:logger).and_return(Logger.new($stdout))
+      allow(vpc).to receive_messages(:url => nil, :token => nil, :connection => nil)
+      vpc
+    end
+
+    let(:actions) do
+      actions = IBM::Cloud::SDK::VPC::INSTANCE::Actions.new(parent)
```
agrare, line 52: Just need to require the sdk gem here and below before using it then the specs should be green

agrare (approved):

gtanzillo:
@chadh1313 Is this something that you need to get in Kasparov?

chadh1313:
@gtanzillo - yes if possible.

simaishi:
Kasparov backport details:

```
$ git log -1
commit bb7837d3e157229701e526c92f5e629cff3efadd
Date:   Mon Nov 2 10:55:19 2020 -0500

    Merge pull request #107 from jaredm-ibm/power_states

    VPC Power states

    (cherry picked from commit dd0bf29db050a31dce3024c827fdd30bd29aaf85)
```

Issue #98, opened 2020-10-27
IBM Cloud (VPC) custom images do not have the operating system linked


chadh1313:
Since the OS is available in the description, if not elsewhere in the API response, a mapping should be possible.

![image](https://user-images.githubusercontent.com/3876941/97313085-9f9e7b00-183c-11eb-8f28-289f6a388794.png)

chadh1313:
Fixed via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/113

Issue #94, opened 2020-10-27
Some VM operating systems are not mapped to a proper icon


chadh1313:
Add any associated mapping for non-standard OS names coming from IBM Cloud.  For example `red-7-amd64`:

![image](https://user-images.githubusercontent.com/3876941/97311665-facf6e00-183a-11eb-8072-dbb5b2ce62c8.png)

jaywcarman:
PowerVS uses a lookup table in our parser to translate to the `OperatingSystem` model `os_map`:

https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb#L8-L14

https://github.com/ManageIQ/manageiq/blob/master/app/models/operating_system.rb#L13-L34

chadh1313:
Fixed via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/113

PR #113, opened 2020-11-02
Operating system mapping updates

chadh1313:
Addresses: 
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/94
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/98

Summary:
- normalizes OS names not captured by operating_system.rb name matching logic
- add version info to image/instance OS
- add OS info for private images

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -5,6 +5,11 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
 
   attr_reader :img_to_os
 
+  # Extensions to the list in operating_system.rb
+  VPC_OS_NAMES = [
+    ["linux_redhat",    %w(red-7 red-8)]
+  ].freeze
+
   def initialize
     @img_to_os = {}
   end
@@ -25,7 +30,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   def images
     collector.images.each do |image|
       img_to_os[image[:id]] = image&.dig(:operating_system, :name)
-      persister.miq_templates.build(
+      persister_image = persister.miq_templates.build(
         :uid_ems            => image[:id],
         :ems_ref            => image[:id],
         :name               => image[:name],
@@ -37,6 +42,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
         :template           => true,
         :publicly_available => true
       )
+
+      image_operating_system(persister_image, image)
     end
   end
 
@@ -119,9 +126,21 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   def instance_operating_system(persister_instance, instance)
     image_id = instance&.dig(:image, :id)
     os = img_to_os[image_id] || pub_img_os(image_id)
+    os_name = normalize_os(os)
     persister.operating_systems.build(
       :vm_or_template => persister_instance,
-      :product_name   => os
+      :product_name   => os_name,
+      :version        => os
+    )
+  end
+
+  def image_operating_system(persister_image, image)
+    os_name = image&.dig(:operating_system, :name)
+    normalized_os_name = normalize_os(image&.dig(:operating_system, :name))
+    persister.operating_systems.build(
+      :vm_or_template => persister_image,
+      :product_name   => normalized_os_name,
+      :version        => os_name
     )
   end
 
@@ -278,5 +297,23 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
 
   def pub_img_os(image_id)
     collector.image(image_id)&.dig(:operating_system, :name)
+  end
+
+  def normalize_os(os_name)
+    normalized_name = OperatingSystem.normalize_os_name(os_name)
+    if normalized_name == "unknown"
+      normalize_vpc_os_name(os_name)
+    else
+      normalized_name
+    end
+  end
+
+  def normalize_vpc_os_name(osName)
+    VPC_OS_NAMES.each do |a|
+      a[1].each do |n|
+        return a[0] unless osName.index(n).nil?
+      end
+    end
+    "unknown"
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -5,6 +5,11 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
 
   attr_reader :img_to_os
 
+  # Extensions to the list in operating_system.rb
+  VPC_OS_NAMES = [
+    ["linux_redhat",    %w(red-7 red-8)]
```
Fryguy, line 10: Is this a VPC specific thing?  If not, I'd accept this change directly in core.

chadh1313: OS names like `red-7-amd64`, `red-7-amd64-sap-applications` coming back from VPC.  Not sure if more odd names will come in the future, so thought having a local map to deal with them as they come could be useful.  But......I'm fine to move this into core if that's the right thing to do.

agrare: So this would match if it were `redhat-7-abcd`:
```
>> OperatingSystem.normalize_os_name("redhat-7-amd64-sap-applications")
=> "linux_redhat"
```

We could fix this in the parser to normalize the `red-` names to `rhel` or `redhat`

chadh1313: @agrare So that's basically what I added, just at the provider level.  Do you recommend moving this mapping to core?  Wasn't sure if `red-` would potentially conflict with something from another provider (say, Redfish).

agrare: I don't think it would conflict with redfish, and as long as it was `"red-7"` and not just `"red-"` it would probably be fine.  I was thinking even simpler than what you have here though.

If we know that `red-*` should be `redhat-*` we could just change that before calling the core normalize_os_name without incurring the overhead of the `VPC_OS_NAMES` and the extra `normalize_vpc_os_name` method.

Something like:
```
os_name.sub!("red-", "redhat-") if name.start_with("red-")
OperatingSystem.normalize_os_name(os_name)
```

chadh1313: Works for me, much simpler.  Don't anticipate too many OS names falling through the cracks of the core mapping.

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -119,9 +121,21 @@ def instance_storage(persister_hardware, instance)
...
       :vm_or_template => persister_instance,
-      :product_name   => os
+      :product_name   => os_name,
+      :version        => os
+    )
+  end
+
+  def image_operating_system(persister_image, image)
+    os_name = image&.dig(:operating_system, :name)
+    normalized_os_name = normalize_os(image&.dig(:operating_system, :name))
```
agrare, line 134: Looks like you can just pass `os_name` in here instead of the `image&.dig` twice

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb
```diff
@@ -49,7 +49,7 @@ def assert_specific_vm
     expect(vm.hardware.cpu_total_cores).to eq(2)
     expect(vm.hardware.cpu_sockets).to eq(2)
     expect(vm.hardware.bitness).to eq(64)
-    expect(vm.operating_system[:product_name]).to eq('red-7-amd64')
+    expect(vm.operating_system[:product_name]).to(eq('linux_redhat'))
```
agrare, line 52: This rubocop warning is actually an error, we have spec files excluded because `expect().to eq()` is the right pattern but they aren't being excluded for some reason after our switch to manageiq-style from manageiq/guides.  You can just ignore the rubocop warning about parens for things like this.

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -119,9 +121,21 @@ def instance_storage(persister_hardware, instance)
...
+    persister.operating_systems.build(
+      :vm_or_template => persister_image,
+      :product_name   => normalized_os_name,
+      :version        => os_name
```
agrare, line 138: Future refactoring, looks like we could combine these methods a bit by doing something like:
```
def instance_operating_system(persister_instance, instance)
  image_id = instance.dig(:image, :id)
  image = collector.images_by_id[image_id] || collector.image(image_id)
  vm_or_template_operating_system(persister_instance, image)
end
```

Since the only difference is looking up the image from the instance

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb
chadh1313: Ok - so line 67 in the refresher_spec.rb (which was hit by rubocop previously) should also be reverted back to the old style?

```
    expect(vm.labels.count).to(eq(3))
```

agrare: :+1: yes

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
agrare: @chadh1313 once you update for https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/113/files#r516935249 I'll take up this refactoring in a follow-up PR

chadh1313: Sounds good - the spec cleanup and update for:  https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/113/files#r516935249 are pushed.

agrare (approved):

simaishi:
Kasparov backport details:

```
$ git log -1
commit 3ca5463dfdd68257894ffc851fec01db4c6f2ecf
Date:   Tue Nov 3 16:08:38 2020 -0500

    Merge pull request #113 from chadh1313/vpc-os-updates

    Operating system mapping updates

    (cherry picked from commit 0f1199827ecafc215bae50d37b6ca6c0a7eef059)
```

PR #115, opened 2020-11-03
Refactor vm_or_template operating_system builder

agrare:
Follow-up to https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/113#discussion_r516937819

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
@@ -20,7 +20,11 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::VPC < ManageIQ::Provi
   end
 
   def images
-    connection.images.all
+    @images ||= connection.images.all.to_a
+  end
+
+  def images_by_id
+    @images_by_id ||= images.index_by { |img| img[:id] }
   end
 
   def image(image_id)
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -3,12 +3,6 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   require_nested :NetworkManager
   require_nested :StorageManager
 
-  attr_reader :img_to_os
-
-  def initialize
-    @img_to_os = {}
-  end
-
   def parse
     floating_ips
     cloud_networks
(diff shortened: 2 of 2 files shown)


agrare:
`-13 +24` I'll take it :laughing:

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -120,21 +113,17 @@ def instance_storage(persister_hardware, instance)
...
-      :product_name   => os_name,
-      :version        => os
-    )
+    image    = collector.images_by_id[image_id] || collector.image(image_id) if image_id
+
+    vm_or_template_operating_system(persister_instance, image) if image
   end
 
-  def image_operating_system(persister_image, image)
+  def vm_or_template_operating_system(vm_or_template, image)
```
chadh1313, line 121: Is the rename of the method to `vm_or_template_operating_system` a best practice or standard used in other providers?

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
```diff
@@ -20,7 +20,11 @@ def flavors
   end
 
   def images
-    connection.images.all
+    @images ||= connection.images.all.to_a
+  end
+
+  def images_by_id
+    @images_by_id ||= images.index_by { |img| img[:id] }
```
chadh1313, line 27: Is it a best practice to build these accessibility maps in the collector rather than the parser?  Makes sense.

chadh1313 (approved):
LTGM - just had a few questions on some of the changes for my own understanding.

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
agrare: Yeah since this is now used by both images and instances I thought it better to name the method for the manager_ref (vm_or_template) [down here :point_down:](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/115/files#diff-be65168162a51f26d1c4e78ad610a312bea5fbefd1cff9fb19152ef1e89b9992R125)

app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
agrare: Yeah this is commonly where we do this sort of thing, it isn't strictly required but when we start doing targeted refresh doing this in the Collector class gives us some freedom on building these hashes without changing the Parser at all.

jaywcarman (approved):

PR #117, opened 2020-11-09
avoiding errors if os_name is nil, and if region has no description.

Real-Omar-Afifi:
@chadh1313 ^

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -295,7 +295,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   end
 
   def normalize_os(os_name)
-    os_name.sub!("red-", "redhat-") if os_name&.start_with?("red-")
+    return "unknown" if os_name.nil?
+    os_name.sub!("red-", "redhat-") if os_name.start_with?("red-")
     OperatingSystem.normalize_os_name(os_name)
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
```diff
@@ -56,7 +56,7 @@ def self.description
   def self.provider_region_options
     ManageIQ::Providers::IbmCloud::VPC::Regions
       .all
-      .sort_by { |r| r[:description].downcase }
+      .sort_by { |r| r[:description]&.downcase }
```
agrare, line 59: @Real-Omar-Afifi what does the UI show if there isn't a description?  I wonder if this is a possibility if we should be dropping any regions without a description in `ManageIQ::Providers::IbmCloud::VPC::Regions.all` instead?

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -295,7 +295,7 @@ def pub_img_os(image_id)
   end
 
   def normalize_os(os_name)
-    os_name.sub!("red-", "redhat-") if os_name.start_with?("red-")
+    os_name.sub!("red-", "redhat-") if os_name&.start_with?("red-")
```
agrare, line 298: Won't this just fail later if `os_name` is nil:
```
>> OperatingSystem.normalize_os_name(nil)
        1: from app/models/operating_system.rb:76:in `normalize_os_name'
NoMethodError (undefined method `downcase' for nil:NilClass)
```

Lets either do `normalize_os(os_name) if os_name` in the callers or `return if os_name.nil?` here

agrare (approved):

simaishi:
Kasparov backport details:

```
$ git log -1
commit 15d6a17c2ad7a73ae942306d988d48a73556f079
Date:   Mon Nov 9 13:05:47 2020 -0500

    Merge pull request #117 from Real-Omar-Afifi/master

    avoiding errors if os_name is nil, and if region has no description.

    (cherry picked from commit d4a782ee13833a453d416b04ad472fa204388dda)
```

PR #129, opened 2021-01-24
Fix cloud volume operations

jaywcarman:
When testing ManageIQ/manageiq-ui-classic#7399 I noticed several bugs.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -20,7 +20,6 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
           :id         => 'size',
           :label      => _('Size (GB)'),
           :type       => 'number',
-          :step       => 1024 * 1024 * 1024,
           :isRequired => true,
           :validate   => [{:type => 'required'}],
         },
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -18,7 +18,7 @@ def self.params_for_create(ems)
           :component  => 'text-field',
           :name       => 'size',
           :id         => 'size',
-          :label      => _('Size (in bytes)'),
+          :label      => _('Size (GB)'),
           :type       => 'number',
           :step       => 1024 * 1024 * 1024,
```
jaywcarman, line 23: @skateman I'm confused as to what `:step` here defines. The resulting options hash passed to `self.raw_create_volume` ends up with the value set by the user (no bytes->GB conversion or anything).

skateman: Other providers require here a size in bytes, the old form consumed gigabytes and then scaled it up in the JS code. I basically eliminated this by going back to bytes and the `step` param is for the [arrow buttons in the numeric input](https://www.w3schools.com/tags/att_input_step.asp).

skateman: If your backend goes with gigabytes, it's fine, just please drop the `step` attribute here...

jaywcarman: That makes sense. Our backend does use GB so I pushed a commit to drop `step`. Thanks @skateman !

agrare (approved):

PR #135, opened 2021-02-11
The CloudVolume create size should be in bytes

agrare:
For consistency with the other providers the size that is passed in
raw_create_volume should be in bytes, then converted to GB before
passing to the provider API.

Follow-up to: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/129
Ref: https://github.com/ManageIQ/manageiq-ui-classic/pull/7399#issuecomment-776929179


jaywcarman (approved):

agrare:
Thanks @jaywcarman !

PR #136, opened 2021-02-11
Fix PowerVS system_pools (flavors) parser

jaywcarman:
The 'system_pools' collector returns a nested hash. The top level key
seems to match the lower level 'type' value, but is not documented in
the API docs
(https://cloud.ibm.com/apidocs/power-cloud#pcloud-systempools-get).

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -235,11 +235,11 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
   end
 
   def flavors
-    collector.system_pools.each do |type|
+    collector.system_pools.each_value do |value|
       persister.flavors.build(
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -235,11 +235,11 @@ def sshkeys
   end
 
   def flavors
-    collector.system_pools.each do |type|
+    collector.system_pools.each do |_key, value|
```
Fryguy, line 238: If you don't need the key, it's slightly cheaper to do 

```suggestion
    collector.system_pools.each_value do |value|
```

jaywcarman: The key is throw-away so using `each_value` is perfect. Done.

agrare (approved):

agrare:
I'm surprised this wasn't failing specs before?

jaywcarman:
> I'm surprised this wasn't failing specs before?

Yeah, we should improve the specs here. The entire hash was being written to the flavor 'name', and we don't have any specs to validate VM relationships to flavors.

PR #137, opened 2021-02-11
New SDK

jaredm-ibm:


.rubocop_local.yml
```diff
@@ -1,3 +1,6 @@
+AllCops:
+  TargetRubyVersion: 2.5
```
Fryguy, line 2: This should not be necessary as the default coming from the shared manageiq-style, is this value - https://github.com/ManageIQ/manageiq-style/blob/70ddb2400036f31404c58a143aff8ed82645b1a0/styles/base.yml#L10

lib/manageiq/providers/ibm_cloud/cloud_tools/sdks/vpc.rb
```diff
@@ -0,0 +1,71 @@
...
+
+require 'ibm_vpc'
+
+dir_name = File.basename(__FILE__).split('.')[0]
+Dir.glob(File.join(dir_name, '*.rb'), :base => File.dirname(__FILE__)) { |req| require_relative req }
+
+module ManageIQ
+  module Providers
+    module IbmCloud
+      module CloudSDK
```
Fryguy, line 11: The module path should match the file path for consistency (and for when we get to Zeitwerk)

```
"ManageIQ::Providers::IbmCloud::CloudSDK::Vpc".underscore + ".rb"
# => "manageiq/providers/ibm_cloud/cloud_sdk/vpc.rb"
```

.rubocop_local.yml
jaredm-ibm: Strange, I saw differences in the linting behaviour when I include the all cops. I'll remove it and retest.

agrare:
https://travis-ci.com/github/ManageIQ/manageiq-providers-ibm_cloud/jobs/482685616#L2110-L2120
```
  1) ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher tests the refresh
     Failure/Error: expect(ems.cloud_subnets.first.cloud_network_id).to eq(ems.cloud_networks[0].id)

       expected: 64000000000011
            got: 64000000000010

       (compared using ==)

     # ./spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb:36:in `assert_ems_counts'
     # ./spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb:17:in `block (3 levels) in <top (required)>'
     # ./spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb:10:in `times'
     # ./spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb:10:in `block (2 levels) in <top (required)>'
```

This looks like it is likely a sporadic failure because there are 4 cloud subnets and `.first` isn't guaranteed to return a consistent order.

Honestly we can probably delete that spec or update it to check that `expect(ems.cloud_subnets.first.cloud_network_id).to be_in(ems.cloud_networks.pluck(:id))`

jaredm-ibm:
Yeah, I noticed that. I've commented the test out for the time being.

agrare:
@jaredm-ibm take this out of WIP when you're ready for this to be merged

jaredm-ibm:
@agrare Working on getting the namespacing reworked today and then it should be done.

lib/manageiq/providers/ibm_cloud/cloud_tools/sdks/vpc.rb
jaredm-ibm: Ok, I've reworked the namespacing to follow the directory structure.

jaredm-ibm:
Work item is ready to review and merge.

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -22,7 +22,7 @@ def connect(options = {})
   # Same as calling connect.
   # @param _auth_type [nil] Not used
   # @param options [Hash] Connection options.
-  # @return [ManageIQ::Providers::IbmCloud::CloudSDK::Vpc]
+  # @return [ManageIQ::Providers::IbmCloud::CloudTools::Vpc]
```
chessbyte, line 25: singular or plural?

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -11,7 +11,7 @@ def supported_auth_attributes
 
   # Return a cloudtools vpc object.
   # @param options [Hash] Hash of options. Default to an empty Hash.
-  # @return [ManageIQ::Providers::IbmCloud::CloudSDK::Vpc]
+  # @return [ManageIQ::Providers::IbmCloud::CloudTools::Vpc]
```
chessbyte, line 14: singular or plural?

lib/manageiq/providers/ibm_cloud/cloud_tools.rb
```diff
@@ -10,9 +10,13 @@
...
   module Providers
     module IbmCloud
+      # Collect SDKs and tools into a single place.
+      module CloudTools
+      end
+
       # Collects and simplifies the intialization of IBM Cloud SDKs.
       # @param apk_key [String] A valid IAM API Key.
-      class CloudTools
+      class CloudTool
```
chessbyte, line 19: very confused on the use of BOTH singular and plural here.

Fryguy: it should be plural to match the file name (or the filename should be singularized)

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk.rb
```diff
@@ -0,0 +1,75 @@
...
+Dir.glob(File.join(dir_name, '*.rb'), :base => File.dirname(__FILE__)) { |req| require_relative req }
+
+module ManageIQ
+  module Providers
+    module IbmCloud
+      module CloudTools
+        module VpcSdk
+        end
+
+        class Vpc < ManageIQ::Providers::IbmCloud::CloudTools::Base::Sdk
```
Fryguy, line 15: I'm not sure I understand why a module of VpcSdk was created (which matches the file name), but then also Vpc was created (which should be in its own file).

Fryguy: or should this be _inside_ VpcSdk?

lib/manageiq/providers/ibm_cloud/cloud_tools.rb
jaredm-ibm: CloudTool is the entry class and CloudTools is the organizing module.

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk.rb
jaredm-ibm: Vpc is the class that is used. VpcSdk is the namespace that organizes the subclasses used by the Vpc class.

jaredm-ibm: There should have been a comment on the VpcSdk to be used for documentation.

Fryguy: I would recommend those living in separate files named vpc_sdk.rb and vpc.rb, respectively

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
jaredm-ibm: Plural for the module. Singular for the class. Normally I would use all uppercase for the organizing module but Zeitwerk doesn't support that.

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
jaredm-ibm: Plural

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
Fryguy: classes are modules in Ruby, so if the only purpose if for subclasses, you can also just nest them under the parent class.  I don't really understand the need for a distinction between the module and the class.

agrare (commented):
Really high level, I still think this is more complicated than it needs to be especially when wrapping existing SDK gems.  The actual https://github.com/IBM/vpc-ruby-sdk will be updated automatically, we'll have to keep our e.g. `VpcSdk::Instance` wrapper up to date if they add fields or actions instead of just being able to start using new items.

If this is the direction you guys want to go though you can

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
jaredm-ibm: It's not for OOP subclassing. It is for organizing. In Vpc the method instances returns VpcSdk::Instances.new(:parent => self).

When you say nest do you mean define a concrete class within the Vpc class?. I don't want to do that. I want a separate file for each all grouped under the same organizing namespace.

Fryguy:
I agree with @agrare ... the more that can be automated via automatic upstream updates, the better.

It seems this mostly just wraps everything in OO patterns, but the maintenance overhead for that is going to be high and I'm not sure it's worth it.  The common stuff like Proxy support and Timeout support seems like a good addition.

agrare:
> The common stuff like Proxy support and Timeout support seems like a good addition.

:+1: agreed, I wonder if we could get that accepted into the sdk generator so everyone can benefit and it'll be part of the upstream gems

jaredm-ibm:
You need to instantiate and play with it. This adds logging, common proxy, common timeout, pagination and response manipulation. For the most part, with the exception of the instances, it passes through the request to core SDK unhindered and then updates the response to something the old SDK used, thus alleviating the need to change every hash access from a symbol to a double quoted string. The instances is a bit more complicated because I didn't want to rewrite the wait for logic in the vm operations.

jaredm-ibm:
@agrare The common gems has this, which is what I am using. I'm simply collecting the various sdks and giving them a central data store to pull from.

agrare:
> response to something the old SDK used, thus alleviating the need to change every hash access from a symbol to a double quoted string.

This is a one-time change though, not sure why we would add a wrapper that we need to maintain long term to avoid a one-time change.

And if we want to move the resource-controller and power gems over to the new format anyway then everything will be returning a consistent format anyway

jaredm-ibm:
When all the SDKs are the same then we can go through and standardize. At the moment the requirement is to replace the previous SDK with the new ones. This does that with the least amount of work.

Fryguy:
I dug in a bit and laid out the class structure, and I think I see what's confusing me

This PR has:

```
CloudTool

CloudTools
  Base
    Instance
    Sdk
  Common
    IamAuth
    Proxy
    Timeout
    GlobalTags
    GlobalTag < CloudTools::Base::Sdk

VpcSdk
  Instance < CloudTools::Base::Instance
  InstanceActions < CloudTools::Base::Instance
  Instances < CloudTools::Base::Instance

Vpc < CloudTools::Base::Sdk
```

So, it feels in many places that subclassing is being used where method sharing is what you want, and subclassing just for method sharing can cause a lot of problems.  Subclassing should answer an "is a" question.  So, some strange things I see immediately are things like `Instances is a CloudTools::Base::Instance` and `InstanceActions < CloudTools::Base::Instance`.  In these cases, it seems you really just want to share the helper methods `wait_for` and `logger`, so for those the general approach is to use a module mixin.

Additionally, there are namespaces being used for organization, but those namespaces themselves could just be the classes in question and then you wouldn't need an extra level of layering.  For example, `CloudTools::Base::Sdk` is designed to be the "base class" for all of the specific Sdks (e.g. `Vpc < CloudTools::Base::Sdk`), but that can also just _be_ the base class (e.g. `Vpc < CloudTools::Base`)  This also would then match other common Ruby organization patterns. such as is done with ActiveRecord::Base, and would bring familiarity to casual readers.

Offhand, this feels like a simpler layout while still providing for the organization you are looking for:

```
CloudTools (formerly CloudTool)
  Base (formerly CloudTools::Base::Sdk)
    Common (formerly CloudTools::Base::Instance)
  IamAuth
  Proxy
  Timeout

Vpc < CloudTools::Base
  Instances (include CloudTools::Base::Common as a mixin)
  Instance (include CloudTools::Base::Common, Vpc::Instance::Actions as a mixin)
    Actions (change to module which will be included in Instance, formerly VpcSdk::InstanceActions)

GlobalTag < CloudTools::Base
```

I could also see a special `CloudTools::Collection` class which `Vpc::Instances` could derive from (or perhaps just use it directly, so you don't need specialized subclasses).

jaredm-ibm:
@Fryguy Thanks for the input. That is partially correct. CloudTool is the entry point into the SDK classes. No other classes need to be instantiated manually. The other SDKs are instantiated by calling methods on CloudTool. Which means that the SDKs can have different implementations behind the scenes, yet still provide a standardized interface to the programmer. It also means that a lot of this discussion about namespaces is academic and used for my internal organization only. The rest of the plugin does not directly access any namespaces outside of the CloudTool class. I'll take a look at your suggestions to see if there is room for improvement.

Here is a repl session of the two SDKs I've integrated (tagging and VPC) working together with minimal repetition. Each request & collection method passes the kwargs as is to the SDK client.

```
[16] pry(main)> ct = ManageIQ::Providers::IbmCloud::CloudTool.new(:api_key => ak, :logger => $stdout)
=> #<ManageIQ::Providers::IbmCloud::CloudTool:0x00007f96c2954dd0
[17] pry(main)> connection = ct.vpc(region: 'us-east')
=> #<ManageIQ::Providers::IbmCloud::CloudTools::Vpc:0x00007f96bc975040
 @cloudtools=#<ManageIQ::Providers::IbmCloud::CloudTool:0x00007f96c2954dd0
[18] pry(main)> ct.tagging
=> #<ManageIQ::Providers::IbmCloud::CloudTools::GlobalTag:0x00007f96b7af8778
 @cloudtools=#<ManageIQ::Providers::IbmCloud::CloudTool:0x00007f96c2954dd0
[19] pry(main)> ls ct.tagging.client
IbmCloudGlobalTagging::GlobalTaggingV1#methods: attach_tag  delete_tag  delete_tag_all  detach_tag  list_tags
[20] pry(main)> ct.tagging.request(:list_tags)
=> {:total_count=>422,
 :offset=>0,
 :limit=>100,
 :items=>
  [{:name=>"a-b:c_e4.32:6784"},...]
[21] pry(main)> ct.tagging.collection(:list_tags)
=> #<Enumerator: ...>
[22] pry(main)> ct.tagging.collection(:list_tags).first
=> {:name=>"a-b:c_e4.32:6784"}
[23] pry(main)> collection = ct.vpc(region: 'us-east')
=> #<ManageIQ::Providers::IbmCloud::CloudTools::Vpc:0x00007f96bc975040
 @cloudtools=#<ManageIQ::Providers::IbmCloud::CloudTool:0x00007f96c2954dd0
[24] pry(main)> ls collection 
ManageIQ::Providers::IbmCloud::CloudTools::Base::Sdk#methods: client  cloudtools  collection  logger  request
ManageIQ::Providers::IbmCloud::CloudTools::Vpc#methods: instances
instance variables: @cloudtools  @sdk_params
[25] pry(main)> ls collection.client
IbmVpc::VpcV1#methods: 
  add_endpoint_gateway_ip                     delete_flow_log_collector                  get_instance_group_manager                  list_endpoint_gateways
  ...
[27] pry(main)> collection.request(:list_instances)
=> {:first=>{:href=>"https://us-east.iaas.cloud.ibm.com/v1/instances?limit=50"},
 :instances=>
  [{:bandwidth=>4000,
    :boot_volume_attachment=>
     {:device=>{:id=>"0777-de4544bd-58bb-4924-8f0c-4edf4fc994b9-8fbnl"},
[28] pry(main)> collection.collection(:list_instances)
=> #<Enumerator: ...>
[29] pry(main)> collection.collection(:list_instances).first
=> {:bandwidth=>4000,
 :boot_volume_attachment=>
  {:device=>{:id=>"0777-de4544bd-58bb-4924-8f0c-4edf4fc994b9-8fbnl"},

[32] pry(main)> ls ManageIQ::Providers::IbmCloud
constants: CloudTool  CloudTools  VERSION
[38] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTool
Object.methods: yaml_tag
ManageIQ::Providers::IbmCloud::CloudTool#methods: authenticator  logger  proxy  tagging  timeout  vpc
[33] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTools
constants: Base  Common  GlobalTag  GlobalTags  Vpc  VpcSdk
[34] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTools::Base
constants: Instance  Sdk
[35] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTools::Common
constants: IamAuth  Proxy  Timeout
[36] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTools::VpcSdk
constants: Instance  InstanceActions  Instances
[37] pry(main)> ls ManageIQ::Providers::IbmCloud::CloudTools::Vpc
Object.methods: yaml_tag
ManageIQ::Providers::IbmCloud::CloudTools::Vpc#methods: instances
```

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
```diff
@@ -0,0 +1,48 @@
...
+module ManageIQ
+  module Providers
+    module IbmCloud
+      module CloudTools
+        module Common
+          # Set proxy configurations dynamically.
+          class Proxy
+            def initialize
+              @keys = %i[address port username password headers]
+              @keys.each { |key| self.class.attr_accessor(key) }
```
Fryguy, line 12: This can be done right in the Proxy class itself as opposed to initialize, which will save on doing this once per instance, which can be expensive.

```ruby
class Proxy
  attr_accessor :address, :port, :username, :password, :headers]
```

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
```diff
@@ -0,0 +1,48 @@
...
+              @password = password
+              self
+            end
+
+            # Convert non-nil keys into a hash.
+            # @return [Hash]
+            def to_hash
+              data = {}
+              @keys.each do |key|
+                variable = instance_variable_get("@#{key}")
```
Fryguy, line 36: This can end up exposing passwords.  You should at least hardcode password as a key to skip over.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
```diff
@@ -0,0 +1,48 @@
...
+                next if variable.nil?
+
+                data[key] = variable # unless variable.nil?
+              end
+              data
```
Fryguy, line 41: Slightly simpler

```suggestion
              @keys.each_with_object({}) { |k, h| h[k] = instance_variable_get("@#{key}") }.compact
```

lib/manageiq/providers/ibm_cloud/cloud_tools/taggings.rb
```diff
@@ -0,0 +1,43 @@
...
+            response = request(call_back, :offset => offset, **kwargs)
+            current_count = response.fetch(:offset) + response.fetch(:limit)
+
+            resources = response[:items]
+            resources&.each { |value| yield value } # rubocop:disable Style/ExplicitBlockArgument
+
+            return if resources.empty?
+
+            each_resource(call_back, :offset => current_count, **kwargs)
+          end
```
Fryguy, line 38: Minor, but this recursion creates a deep stack when a simple loop can do the same thing without the recursion. and thus without hurting the stack depth

```suggestion
          def each_resource(call_back, offset: 0, **kwargs)
            loop do
              response = request(call_back, :offset => offset, **kwargs)

              resources = response[:items]
              break if resources.empty?

              resources&.each { |value| yield value } # rubocop:disable Style/ExplicitBlockArgument

              offset = response.fetch(:offset) + response.fetch(:limit)
            end
          end
```

Fryguy: You might even be able to remove the offset parameter, because I think it was only exposed for the recursion itself.

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk.rb
```diff
@@ -0,0 +1,75 @@
...
+
+            # VPC has a next key that holds the next URL.
+            return unless response.key?(:next)
+
+            # The :next data structure is a hash with a href member.
+            next_url = response.dig(:next, :href)
+            return unless next_url
+
+            start = Addressable::URI.parse(next_url).query_values['start']
+            each_resource(call_back, :start => start, :array_key => array_key, **kwargs)
```
Fryguy, line 61: Similar comment here about simple loop vs recursion.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
jaredm-ibm: @keys is used in to_hash which is used to pass the proxy data to the IBM SDK init. I could use the standard attr_accessor, but then I'd have to maintain two lists. Also there should only be 1 instance of Proxy created per instance of CloudTool, so the performance hit shouldn't be too significant.

Fryguy:
> The rest of the plugin does not directly access any namespaces outside of the CloudTool class.

Ok, I think this is the big sentence for me as that means that most of these classes are really designed to create a chainable interface, where the user only actually instantiates the top-level class.  If that's the case, then, I think I'm good with this.  

Like you said, the organization itself then is mostly academic, and while I personally think we should remove extraneous layers of indirection, I don't think that should hold up this PR.  As we move forward and start incorporating more and more upstream SDKs, I think the overhead/maintenance of these classes is going to be too much and we should probably drop them in favor of just using the raw SDKs directly, but as a transitional approach to make the current code work, I think this is fine.

lib/manageiq/providers/ibm_cloud/cloud_tools/taggings.rb
jaredm-ibm: The :offset is used in the Tags list_tags API. It tells the server where to start its next page.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
Fryguy: It's still preferable to do it just once at the class level as this is essentially static data.

```ruby
class Proxy
  ACCESSOR_KEYS = %i[address port username password headers].freeze
  attr_accessor *ACCESSOR_KEYS

#...

  def to_hash
    self.class::ACCESSOR_KEYS.each_with_object({}) { |k, h| h[k] = instance_variable_get("@#{key}") }.compact
  end
```

jaredm-ibm:
@Fryguy The raw SDKs do not handle pagination. Pagination is not standard amongst SDKs. So you have to create a new pagination function for each API. You also need to figure out a place that makes sense to instantiate each SDK you use, and all of the boilerplate that entails. CloudTool does all of that and provides simplicity, standardization and options. It also keeps most of the SDK details out of the Plugin classes, and provides a way to instantiate the SDKs in isolation.

lib/manageiq/providers/ibm_cloud/cloud_tools/taggings.rb
jaredm-ibm: Reimplemented as loop.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
jaredm-ibm: Reimplemented in conjunction with suggestions in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/137#discussion_r577167542

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
jaredm-ibm: Reimplemented suggested code with minor fixes.

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk.rb
jaredm-ibm: Files split.

Fryguy:
@jaredm-ibm Can you please remove my github handle from commit https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/137/commits/309bef0340f592af7eaf2ce8ffe41df2ba98f49c (i.e. amend the commit)?  The reason is that I'm going to get an email every time somebody does something with this commit (like rebase, extract into a new repo, whatever)

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance_actions.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module ManageIQ
+  module Providers
+    module IbmCloud
+      module CloudTools
+        module VpcSdk
+          class InstanceActions < ManageIQ::Providers::IbmCloud::CloudTools::Base::Instance
```
agrare, line 8: Hey @jaredm-ibm is the reason you're subclassing `Base::Instance` here so that it can share some of those methods?  It is confusing for `InstanceActions` and `Instance` and `Instances` to all inherit from `Instance`.  At least `InstanceActions` isn't `kind_of?(Instance)`

If you're trying to pull e.g. `wait_for` into each of these maybe a module/mixin would be a better choice?

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
```diff
@@ -0,0 +1,39 @@
...
+            def authentication(username:, password:)
+              @username = username
+              @password = password
+              self
+            end
+
+            # Convert non-nil keys into a hash.
+            # @return [Hash]
+            def to_hash
+              self.class::ACCESSOR_KEYS.each_with_object({}) { |key, obj| obj[key] = instance_variable_get("@#{key}") }.compact # rubocop:disable Rails/IndexWith
```
Fryguy, line 32: Whoa, TIL about `index_with`...that's *awesome*, and we should totally use that.

```suggestion
              ACCESSOR_KEYS.index_with { |k| instance_variable_get("@#{key}") }.compact
```

jaredm-ibm:
@Fryguy  We can squash and merge the PR which will discard the commit message.

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance_actions.rb
jaredm-ibm: The reason for the subclass of Base::Instance is to have standard @parent and logger method. The wait_for is just there because it is generic enough to be useful elsewhere.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
jaredm-ibm: It is a cool method https://sebastian-palma.github.io/2020-06-08/rails-index-by-index-with. However, 'index_with' is added by rails. I don't want to use a non-standard method.

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance_actions.rb
agrare: :+1: that's where we were saying it'd be better to use a mixin if you're trying to share common code.

> So, some strange things I see immediately are things like Instances is a CloudTools::Base::Instance and InstanceActions < CloudTools::Base::Instance. In these cases, it seems you really just want to share the helper methods wait_for and logger, so for those the general approach is to use a module mixin.

Inheritance implies the child class is a type of the parent class /cc @Fryguy

jaredm-ibm: Changed the naming to SDK::Branch for SDK Wrappers that branch off of CloudTool and SDK::Leaf for Classes that deal with specific aspects of the SDK Branch. This gives you your 'is a' relationship without breaking my design.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
agrare: Given this plugin is a rails engine having rails is a pretty safe bet but I don't think we should hold this up for this

Fryguy: Rails (active_support) is guaranteed

agrare (approved):

Fryguy:
We intentionally don't squash and merge PRs, so please amend.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+dir_name = File.basename(__FILE__).split('.')[0]
```
Fryguy, line 3: Not for this PR, since we are close to merge but this is already provided by Ruby with the `__dir__` method

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
```diff
@@ -0,0 +1,16 @@
...
+Dir.glob(File.join(dir_name, '*.rb'), :base => File.dirname(__FILE__)) { |req| require_relative req }
```
Fryguy, line 4: So this all can just be:

```suggestion
Dir.glob(File.join(File.expand_path(__dir__), "*.rb").sort.each { |r| require r }
```

Note the `.sort`, which is handy to ensure consistent ordering, otherwise glob's order is not guaranteed.

spec/spec_helper.rb
```diff
@@ -3,23 +3,50 @@
   SimpleCov.start
 end
 
-Dir[Rails.root.join("spec/shared/**/*.rb")].each { |f| require f }
-Dir[File.join(__dir__, "support/**/*.rb")].each { |f| require f }
+Dir[Rails.root.join("spec/shared/**/*.rb")].sort.each { |f| require f }
+Dir[File.join(__dir__, "support/**/*.rb")].sort.each { |f| require f }
```
Fryguy, line 7: This is interesting...we should probably add this to the core generator for consistency - @agrare ?

spec/spec_helper.rb
```diff
@@ -3,23 +3,50 @@
...
+# Sanitize VPC VCR files.
+def vpc_sanitizier(interaction)
+  # Mask bearer token in recorded file.
+  interaction.request.headers['Authorization'] = 'Bearer xxxxxx' if interaction.request.headers.key?('Authorization')
+  # Replace IP V4 Addresses
+  interaction.response.body.gsub!(/([0-9]{1,3}\.){3}/, '127.0.0.')
+  # Replace ssh key data.
+  replace_ssh_keys(interaction.response)
+  interaction
+end
```
Fryguy, line 41: Not for this PR, but we commonly move all this into a specific helper file such as spec/support/vcr_helper.rb

That being said we also have some common patterns for dealing with hiding credentials that we will probably want to follow for some consistency, though I'm not sure they can handle everything in this PR.

cc @agrare

spec/vcr_cassettes/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.yml
```diff
@@ -5,81 +5,88 @@ http_interactions:
...
-      Accept:
-      - "*/*"
       User-Agent:
-      - Ruby
+      - ibm_vpc-ruby-sdk-0.1.0 x86_64-apple-darwin19.6.0 ruby-2.6.6
+      X-Ibmcloud-Sdk-Analytics:
+      - service_name=vpc;service_version=V1;operation_id=list_floating_ips
+      Accept:
+      - application/json
+      Authorization: Bearer xxxxxx
```
Fryguy, line 68: I noticed that before, this value was an Array and now it's a scalar.  Is that expected?  Not sure if that's a side effect of the sanitization, and whether or not it's a problem or not.

spec/spec_helper.rb
agrare: Do we need require order to be guaranteed?  Doesn't seem to be needed for core or any providers currently.

spec/spec_helper.rb
```diff
@@ -3,23 +3,50 @@
...
+def replace_token_contents(response)
+  data = JSON.parse(response.body, :symbolize_names => true)
+  data.merge!({:refresh_token => 'REFRESH_TOKEN', :ims_user_id => '22222', :expiration => Date.new(2100, 1, 1).to_time.to_i})
+  response.body = data.to_json.force_encoding('ASCII-8BIT')
+end
+
+# Sanitize VPC VCR files.
+def vpc_sanitizier(interaction)
+  # Mask bearer token in recorded file.
+  interaction.request.headers['Authorization'] = 'Bearer xxxxxx' if interaction.request.headers.key?('Authorization')
```
Fryguy, line 35: re: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/137/files#r577950732  Looks like this accidentally flattens the Array that was there previously, so I'm not sure if that could cause problems or not.

.rubocop_local.yml
```diff
@@ -4,4 +4,4 @@ Style/GlobalVars:
 
 Style/MethodCallWithArgsParentheses:
   Exclude:
-  - spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm_spec.rb
+  - spec/models/manageiq/providers/ibm_cloud/vpc/**/*.rb
```
agrare, line 7: Is this required?  I deleted this cop override entirely and don't see any offenses:
```
$ git diff
diff --git a/.rubocop_local.yml b/.rubocop_local.yml
index 268fbb2..77bf231 100644
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,7 +1,3 @@
 Style/GlobalVars:
   AllowedVariables:
   - $ibm_cloud_log # Global cloud plugin log from ManageIQ.
-
-Style/MethodCallWithArgsParentheses:
-  Exclude:
-  - spec/models/manageiq/providers/ibm_cloud/vpc/**/*.rb
```

```
$ bundle exec rubocop spec/models/manageiq/providers/ibm_cloud/vpc/

Inspecting 4 files
....

4 files inspected, no offenses detected
```

spec/spec_helper.rb
Fryguy: Probably not, but it's more deterministic just in case.

spec/vcr_cassettes/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.yml
jaredm-ibm: It is a product of the sanitization. It is okay. Spec tests pass because they aren't submitting a live request and thus don't care about authorization.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
jaredm-ibm: No it isn't. This is the name of the file minus the .rb. dir_name is a reference to the target directory for the relative import. So in this case common.rb will look for ruby files in the 'common' directory and import them relatively.

.rubocop_local.yml
jaredm-ibm: I see it in my linter, thus the reason I modified it. Leave it, please. Your rubocop may not be the same version as mine. I don't want this cop to show up in my linter.

spec/spec_helper.rb
jaredm-ibm: Spec tests pass, which is the reason for the VCR file.

.rubocop_local.yml
agrare: Your linter should be pulling in the base rubocop yml config though so my linter config is the same as miq-bot's should be the same as your.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
jaredm-ibm: require_relative is less overhead on the system as it skips a lot of logic path lookup logic. I don't want to get the .rb files in the parent directory I want to get the files in the sub directory.

```
puts 'My implementation'
dir_name = File.basename(__FILE__).split('.')[0]
puts dir_name
Dir.glob(File.join(dir_name, '*.rb'), :base => File.dirname(__FILE__)) { |req| puts "require_relative(#{req})"; require_relative req }

puts "Jason's change"
Dir.glob(File.join(File.expand_path(__dir__), "*.rb")).sort.each { |r| puts "require(#{r})"; require r }
```

```
My implementation
common
require_relative(common/timeout.rb)
require_relative(common/iam_auth.rb)
require_relative(common/proxy.rb)

Jason's change
require(/.../manageiq-providers-ibm_cloud/lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb)
require(/.../manageiq-providers-ibm_cloud/lib/manageiq/providers/ibm_cloud/cloud_tools/sdk.rb)
require(/.../manageiq-providers-ibm_cloud/lib/manageiq/providers/ibm_cloud/cloud_tools/taggings.rb)
require(/.../manageiq-providers-ibm_cloud/lib/manageiq/providers/ibm_cloud/cloud_tools/vpc.rb)
require(/.../manageiq-providers-ibm_cloud/lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk.rb)
```

.rubocop_local.yml
Fryguy: Not sure what linter you are using @jared-ibm, but ideally you should use the one that is used by @miq-bot so that we are all seeing the same thing.  This is probably why you are not seeing https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/137#issuecomment-780853035 in your personal linter.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
Fryguy: ohhh...yeah you're right...in that case you can also use `File.basename(__FILE__, ".*")`

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
agrare: If we're worried about the runtime cost of require vs require_relative we could also speed this up by dropping the `Dir.glob` and just list out the requires:
```
require_relative(common/timeout.rb)
require_relative(common/iam_auth.rb)
require_relative(common/proxy.rb)
```

.rubocop_local.yml
jaredm-ibm: I am using 1.3.1 which is from Nov 16, 2020. The current is 1.10, so both of us are behind.

spec/vcr_cassettes/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.yml
agrare: If we're the ones doing the sanitization we might as well have it match what the request would look like

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
jaredm-ibm: I use CloudTool without the rest of rails to verify what I expect to work does work, this would hinder that ability.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
jaredm-ibm: True, however listing out the requires would hinder reorganization. I'm pointing out that my implementation is both performative and flexible.

lib/manageiq/providers/ibm_cloud/cloud_tools/common/proxy.rb
agrare: We do want to be able to run e.g. the collectors without rails :+1: 
That said requiring the active_support gem is pretty common and doesn't incur the same overhead that all of rails does.

spec/vcr_cassettes/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.yml
jaredm-ibm: Why? Once the rest call has been authenticated and returned its payload this header is useless.

lib/manageiq/providers/ibm_cloud/cloud_tools/common.rb
Fryguy: I'm not saying your code is wrong, I'm saying it can be cleaner by using __dir__, which was a mistake on my part because it was hard to read.

That being said, if we care about performance, I believe the fastest is using plain require with absolute paths (which is what I was trying to write, and why I added expand_path), then require_relative then require with other paths.

Fryguy: Oh I see what I did...I *meant* to write,

```ruby
Dir.glob(File.join(File.expand_path(__dir__), "common", "*.rb").sort.each { |r| require r }
```

which is the require pattern we use elsewhere in the code base.

.rubocop_local.yml
Fryguy: The bot is using 0.82.0, which is what we should all use for consistency.  To get the right version you just need to run `bundle exec rubocop` since the manageiq-style gem drives the rubocop version both for the projects and for the bot, so they are in sync.  There are issues with inconsistencies in the configuration between rubocop versions which is why it's important we all match.

jaredm-ibm: Ummm, you are telling us that we should be able to react to the changes in the automated SDKs but we should also stay at an older version of a secondary check doesn't really work for me. I have no problem keeping up with SDK changes, just as I have no problem with keeping up with newer versions of rubocop, or ruby for that matter.

lib/manageiq-providers-ibm_cloud.rb
```diff
@@ -1,2 +1,3 @@
 require "manageiq/providers/ibm_cloud/engine"
 require "manageiq/providers/ibm_cloud/version"
+require_relative "manageiq/providers/ibm_cloud/cloud_tool"
```
Fryguy, line 3: @agrare, as you found in https://github.com/ManageIQ/manageiq-rpm_build/pull/137, it seems this line breaks the build as it eager loads this too early in the cycle.

agrare: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/145

PR #139, opened 2021-02-12
Powervs SAP provisioning

jaywcarman:


jaywcarman:
@miq-bot remove_label wip

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -124,7 +124,7 @@ def images
         :uid_ems            => id,
         :ems_ref            => id,
         :name               => ibm_image.name,
-        :description        => ibm_image.description.presence,
+        :description        => ibm_image.specifications.image_type,
```
agrare, line 127: Is this so that you can identify images as SAP images in provisioning later?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/options_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  def cloud_instance_id
+    source.ext_management_system.uid_ems
+  end
+
+  def vm_image
+    @vm_image ||= MiqTemplate.find_by(:id => get_option(:src_vm_id))
+  end
+
+  def sap_image?
+    vm_image.description == 'stock-sap'
```
agrare, line 11: Would it be better to check the vm_image flavor class name instead of using the description?

We could even add a method to your template subclass:
```
class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Template
  def sap_image?
    flavor.kind_of?(ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile)
  end
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -22,8 +22,53 @@ def volume_dialog_keys
     %i[name size shareable]
   end
 
+  def vm_image
+    template_id = values&.dig(:src_vm_id, 0)
+    ar_ems.miq_templates.select { |t| t.id == template_id }[0]
```
agrare, line 27: Might want to memoize this as `@vm_image ||= ` since it looks like you reference `sap_image?` and `sap_flavor` a few times.

Also `ems.miq_templates.select { |t| t.id == template_id }[0]` looks like just `ems.miq_templates.find_by(:id => template_id)`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -22,8 +22,53 @@ def volume_dialog_keys
...
+  end
+
+  def sap_image?
+    vm_image.description == 'stock-sap'
+  end
+
+  def sap_flavor
+    if sap_image?
+      selected_flavor_name = values&.dig(:sys_type, 1)
+      ar_ems.flavors.select { |flavor| flavor.name == selected_flavor_name }[0]
```
agrare, line 37: Same thing here, just `.find_by(:name => selected_flavor_name)` will work

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -22,8 +22,53 @@ def volume_dialog_keys
...
+        0 => "capped",
+        1 => "shared",
+        2 => "dedicated"
+      }
+    end
+  end
+
   def allowed_sys_type(_options = {})
-    ar_sys_types = ar_ems.flavors
+    flavor_type = sap_image? ? "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile" : "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SystemType"
```
agrare, line 68: Super minor but if you want to save some horizontal space you can do
`"ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::#{sap_image? ? "SAPProfile" : "SystemType"}"`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -22,8 +22,53 @@ def volume_dialog_keys
...
+
+    ar_sys_types = ar_ems.flavors.find_all { |flavor| flavor.type == flavor_type }
```
agrare, line 70: Maybe a good follow-up but we could define an association on your `PowerVirtualServers::CloudManager` that would let you do `ems.sap_flavors` and `ems.whatever_the_opposite_flavors` by specifying the `:class_name` in the assoc

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -43,6 +43,10 @@ def ports(network_id)
     networks_api.pcloud_networks_ports_getall(cloud_instance_id, network_id).ports || []
   end
 
+  def sap_profiles
+    @sap_profiles ||= sap_api.pcloud_sap_getall(cloud_instance_id)
```
agrare, line 47: Might want to stay consistent with the other collections which cache the resulting array not the whole response so:
`@sap_profiles ||= sap_api.pcloud_sap_getall(cloud_instance_id).profiles || []` 
Then you would do
`collector.sap_profiles.each` in the parser below.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/sap_profile.rb
```diff
@@ -0,0 +1,2 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile < ::Flavor
+end
```
agrare, line 2: Don't forget to `require_nested :SAPProfile` in the `PowerVirtualServers::CloudManager`

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
jaywcarman: Yeah, there are three image "types": `stock` (this could translate to the existing 'public' column maybe), `stock-sap` (special SAP images that you can provision with profiles to get a certified environment and better pricing) and `import` (user imported). The high level goal of this PR is to allow provisioning of these `stock-sap` images with the special profiles.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/options_helper.rb
jaywcarman: I'm not sure I understand fully. `vm_image` here doesn't have a direct flavor association here right (it's just a Template/image)? The user selected flavor is included in the clone_options, and could be queried directly in the cloning module.

agrare: Ah yes you're right, sorry idk why I thought your images had flavors

jaywcarman: Past the details here, though, I like the idea of checking a type rather than comparing string.

agrare (approved):

PR #140, opened 2021-02-16
Implemented DDF for cloud subnets and added support mixin call

skateman:
Screenshot from 2021-02-16 12-08-59

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/6861

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager/cloud_subnet.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager/cloud_subnet.rb
@@ -5,6 +5,52 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::NetworkManager::CloudS
     end
   end
 
+  def self.params_for_create(_ems)
+    {
+      :fields => [
+        {
+          :component    => 'select',
+          :name         => 'type',
+          :id           => 'type',
+          :label        => _('Type'),
+          :isRequired   => true,
+          :validate     => [{:type => 'required'}],
+          :initialValue => 'vlan',
+          :options      => [
+            {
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager/cloud_subnet.rb
```diff
@@ -5,6 +5,49 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::NetworkManager::CloudS
...
 
+  def self.params_for_create(_ems)
+    {
+      :fields => [
+        {
+          :component => 'select',
+          :name      => 'type',
+          :id        => 'type',
+          :label     => _('Type'),
+          :options   => [
```
skateman, line 16: @jaywcarman is this field a two-state, or can it be also empty?

jaywcarman: This field must be one of the two states.

skateman: Okay, in that case it should be required...changing

jaywcarman (approved):
LGTM. @agrare since this form update won't be included in Lasker should we wait to merge? I don't see any obvious harm, but can understand general prudence.

skateman:
> LGTM. @agrare since this form update won't be included in Lasker should we wait to merge? I don't see any obvious harm, but can understand general prudence.

@jaywcarman the change is harmless as long as the UI is not consuming the form schema...

Fryguy:
Skipping backport to `morphy`, because it is already in the branch.

PR #145, opened 2021-02-18
Don't require cloud_tools in lib/manageiq-providers-ibm_cloud.rb

agrare:
This is causing issues during the build because it is requiring runtime packages to be present during the build due to the eager loading of this class

https://github.com/ManageIQ/manageiq-rpm_build/pull/137#issuecomment-781568150


Fryguy:
Nice find.  Rails autoload should take care of this anyway, so it's unnecessary.

Fryguy (approved):

Fryguy:
cc @jaredm-ibm

PR #152, opened 2021-03-05
Use the combined child-manager refresh support from core

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21099


Fryguy:
Wow...that's a lot of savings. ` +17 −123`

agrare:
Backported in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/268

Issue #154, opened 2021-03-16
Allow for use of Bearer token initialization in CloudTool.


jaredm-ibm:
## Problem ##
* The CloudTool cannot be memoized in an memcache environment. 
* Each new CloudTool object using an IAM API key takes 2-3 seconds to retrieve the access_token. 
* In the provision where we want the option to query the cloud for information not stored in the local DB. 
* This delay is too long and leads to a laggy user interface. 

## Solution ##
A solution is to memoize the access_token with other pertinent information and use that as the basis for authentication. With the bearer token cached locally CloudTool initialization is almost instantaneous, leaving only the time to retrieve the specific information as a factor.

The memoization of the Bearer Auth should use ManageIQ::Password.encrypt/decrypt to secure the information within the ProvisonWorkflow.

PR #155, opened 2021-03-16
Provide method for CloudTool to authenticate with pre-fetched user_token.

jaredm-ibm:
# Issue
The issue is described in issue https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/154. Essentially, the CloudTool initialization can be reduced by 2-3 seconds by seeding the class with a valid Bearer Token instead of always using the API Key.

# Priority
It is my intention to use this within the VPC ProvisionWorkflow. The current plan is to add db fields to the objects that don't contain enough information. This work will be to provide options where that is not viable. It also allows for a more flexible design.

# Implementation
This batch of work adds the ability for the IBM Derived SDKs to use the subclassed IbmVpc::Authenticators::BearerTokenAuthenticator and IbmVpc::Authenticators::IamAuthenticator classes to fetch authentication information.

* The new classes define a standard bearer_info method which returns the user_token, expire_time and api_key in a hash.
* A new module was created to house the implementation.
```
ManageIQ::Providers::IbmCloud::CloudTools::Authentication
constants: BearerAuth  IamAuth
ManageIQ::Providers::IbmCloud::CloudTools::Authentication.methods: new_auth  new_bearer  new_iam
```
* The new_auth method is the only method that is used by CloudTool. The other provide support, and spec testing isolation.
* The new_auth method takes two kwargs `:api_key` & `:bearer_info`. Both can be present or just one. The return of the new_auth method is always a `BearerAuth` class. Internally the `IamAuth` will be called and then converted into a `BearerAuth`.
* An IAM `user_token` has a time-to-live of 1 hour. If that time is reached the BearerAuth class will try to get a new token using the `api_key` included with the bearer_info hash. If no API key is provided then it will throw an exception.
* CloudTool adds a new kwarg `:bearer_info` which is saved to an instance variable. The `#authenticator` method  adds this bearer_info and passes in the instance variable. The returned object should be a new BearerAuth class.
* Like the previous method the `#authenticator` method is only called in conjunction with requesting a CloudTool method that returns a Ibm-derived SDK wrapper class. At that point the supplied api_key is fetched. The bearer token can be fetched before hand using the `#authenticator` method. The bearer_info hash can be fetched using `#bearer_info` method on the `BearerAuth` instance.

# Not covered
* Validation of user_token. There may be an issue with the user token is rejected.  This should happen in the client request as a 401. There is no plan in place to handle this error, or provide users with a method of regenerating the token from a client. The expiry checking should kick in before the client request is issued. Handling of this error should happen at a higher level.

# Specs
* The CloudTool spec has been rewritten to test the various authentication scenarios.
* It also adds tests for the `#request` and `#collection` methods in VPC.
* There may be an unknown cop/unnecessary disable warning for rubocop. This is for the Style/MethodCallWithArgsParentheses in the cloud_tools_spec.rb. It is disabled because the cop does not provide valid advice for `expect` methods and junks up my IDE. It causes a warning because it is a newer cop not in the version MiqBot uses.
* VCRs regenerated based on changes to both test & class structure.

# Common files
* No changes to any common files.


lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
```diff
@@ -0,0 +1,138 @@
...
+              bearer_info
+            end
+
+            # Validate expire time is set and greater than now - 10 seconds.
+            #
+            # @return [Boolean]
+            def verify_valid(expire_time)
+              return false if expire_time.nil?
+
+              expire_time <= (Time.now.to_i - 10)
```
agrare, line 93: Wouldn't the token be valid if the `expire_time > Time.now.to_i` or `expire_time > Time.now.to_i + 10` if you want to give a little "buffer" when it is still valid but you want people to renew it?

agrare: Oh you're expecting `true` returned from `verify_valid` to mean it has expired?  That's a bit counterintuitive, maybe rename this method `#expired?`

I would expect a truthy return from a method called `valid` would mean it was valid not invalid

jaredm-ibm: The name of the method should be 'is_expired'. Expired is true, Valid is false.

jaredm-ibm: Yeah, that got me too. I must have named the method before finalizing the logic and forgot to update the method name.

agrare (approved):

agrare:
Backported in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/268

PR #159, opened 2021-03-19
Adding new IBM Cloud-Object-Storage provider

iv1111:
- this manager will implement a connection to the
  IBM Cloud-Object-Storage, for information see:
  https://www.ibm.com/cloud/object-storage

- adding skeleton code, currently working on
  the 'parse' and related methods

Depends on: https://github.com/ManageIQ/manageiq-ui-classic/issues/7675


iv1111:
@agrare Hey, Adam. We're starting out work on the IBM Cloud-Object-Storage provider, which will display cloud-buckets and objects in those cloud-buckets within MIQ. Only thing is, the registration of standalone Object-Providers in MIQ is disabled as of now.

iv1111:
Here the build seems to fail due to a ruby version.

agrare:
That should have been fixed by https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/160

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager.rb
```diff
@@ -0,0 +1,37 @@
...
+  require_nested :MetricsCapture
+  require_nested :MetricsCollectorWorker
+  require_nested :Refresher
+  require_nested :RefreshWorker
+
+  supports :ems_storage_new
+  include ManageIQ::Providers::StorageManager::ObjectMixin
+
+  def self.ems_type
+    @ems_type ||= "cloud_object_storage".freeze
```
agrare, line 11: This should have some indication that it is related to the ibm provider plugin, suggest `ibm_cloud_object_storage`

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def self.description
+    @description ||= "Cloud Object Storage".freeze
```
agrare, line 15: Same, this should probably be `IBM Cloud Object Storage`

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager.rb
iv1111: right, i will rename

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager.rb
iv1111: same, I'll rename now

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager.rb
```diff
@@ -0,0 +1,37 @@
+class ManageIQ::Providers::IbmCloud::CloudObjectStorage::StorageManager < ManageIQ::Providers::StorageManager
```
agrare, line 1: Since the vendor here is already `IbmCloud` I wonder if it'd be better (or at least shorter) to call the class name here just: `ManageIQ::Providers::IbmCloud::ObjectStorage::StorageManager`

IBM Cloud Cloud Object Storage is a little redundant

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager/event_catcher.rb
```diff
@@ -0,0 +1,3 @@
+class ManageIQ::Providers::IbmCloud::CloudObjectStorage::StorageManager::EventCatcher < ManageIQ::Providers::BaseManager::EventCatcher
```
agrare, line 1: Unless we're going to implement an event catcher in this PR lets delete these scaffolding classes

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager/metrics_capture.rb
```diff
@@ -0,0 +1,82 @@
+class ManageIQ::Providers::IbmCloud::CloudObjectStorage::StorageManager::MetricsCapture < ManageIQ::Providers::BaseManager::MetricsCapture
```
agrare, line 1: Same with metrics capture

app/models/manageiq/providers/ibm_cloud/cloud_object_storage/storage_manager/refresh_worker.rb
```diff
@@ -0,0 +1,3 @@
+class ManageIQ::Providers::IbmCloud::CloudObjectStorage::StorageManager::RefreshWorker < MiqEmsRefreshWorker
+  require_nested :Runner
+end
```
agrare, line 3: Believe you'll need to set the `def self.settings_name` here like in the `vpc/cloud_manager/refresh_worker` class

iv1111:
@agrare I incorporated your suggestions.

agrare:
This is looking good, just needs params_for_create, verify_credentials, and raw_connect at the class level, and verify_credentials, connect at the instance level.

We don't need to do refresh for the first PR as long as we can add the provider and it verifies credentials properly

iv1111:
> This is looking good, just needs params_for_create, verify_credentials, and raw_connect at the class level, and verify_credentials, connect at the instance level.
> 
> We don't need to do refresh for the first PR as long as we can add the provider and it verifies credentials properly

@jaywcarman Do you know if COS has an OpenAPI description so that we can generate the corresp. SDK/API gem for this an future tasks?

jaywcarman:
> @jaywcarman Do you know if COS has an OpenAPI description so that we can generate the corresp. SDK/API gem for this an future tasks?

Yes, the download button on the [COS API docs page](https://cloud.ibm.com/apidocs/cos/cos-compatibility) is not easy to find:
image

Direct link to the openapi def json:
https://cloud.ibm.com/apidocs/cos/cos-compatibility.json

We'll need to work with the IBM SDK generation team.

iv1111:
I added the same structure for COS provider as for VPC and PVS wrt. to the inventory. I've also added the _params_for_create_ implementation.

iv1111:
I've added the connect and verify_credentials methods, using IAM and S3 API for that.
Valid GUID check pending. 

This PR is still blocked by https://github.com/ManageIQ/manageiq-ui-classic/issues/7675 as the rest of the image import workflow.

iv1111:
Is this a legit build error? I can't understand from the logs.

iv1111:
@agrare oh, probably missing value in the _settings.yaml_

agrare:
@iv1111 that error means that you've added new workers that don't have equivalent systemd unit files.  Take a look at the existing .target and .service files in the systemd/ directory

iv1111:
> @iv1111 that error means that you've added new workers that don't have equivalent systemd unit files. Take a look at the existing .target and .service files in the systemd/ directory
Why isn't this necessary for e.g. Vpc and Power Virtual Servers?

iv1111:
> Take a look at the existing .target and .service files in the systemd/ directory

There are a bunch of systemd files in there but nothing for Vpc and PVS that I'd take as an example.

agrare:
@iv1111 I'm not sure what you mean, there is a .target and a .service file for the vpc refresh worker and the powervs refresh worker: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/tree/master/systemd

iv1111:
> @iv1111 that error means that you've added new workers that don't have equivalent systemd unit files. Take a look at the existing .target and .service files in the systemd/ directory

sorry, I was looking at the system directory of the base repo

agrare:
Ahh yeah that is pluggable so each provider owns their own units

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
```diff
@@ -0,0 +1,7 @@
+class ManageIQ::Providers::IbmCloud::Inventory::Collector::ObjectStorage < ManageIQ::Providers::IbmCloud::Inventory::Collector
+  require_nested :StorageManager
+
+  def connection
+    @connection ||= manager.connect
+  end
+end
```
Fryguy, line 7: Seems the bot isn't running on this repo otherwise it would have picked up on this?

jaywcarman:
I'm working on specs for this: https://github.com/jaywcarman/manageiq-providers-ibm_cloud/tree/cloud_object_storage_specs

@iv1111 I'm only going to push passing specs to my branch ^. Feel free to incorporate into this PR when you need Travis to pass so we can merge.

iv1111:
Apart from UI issues mentioned [here](  https://github.com/ManageIQ/manageiq-ui-classic/pull/7710#issuecomment-830233539) looking pretty good. We have registration/authentication/verification + inventory parsing (cloud buckets, cloud objects, corresp. stats).

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
iv1111: We do it in PowerVS too. Do you mean it should be done in a different way?

app/models/manageiq/providers/ibm_cloud/object_storage/object_manager.rb
```diff
@@ -0,0 +1,211 @@
+class ManageIQ::Providers::IbmCloud::ObjectStorage::ObjectManager < ManageIQ::Providers::StorageManager
```
agrare, line 1: I think we should follow the other (power and vpc) and go with Vendor::Service::ManagerType so e.g. `ManageIQ::Providers::IbmCloud::Vpc::CloudManager` which would mean this would be `ManageIQ::Providers::IbmCloud::ObjectStorage::StorageManager`.

We don't have an `ObjectManager` type which makes this confusing to me at least.

The only other pattern is when there are multiple types of storage managers (e.g. OpenStack/Amazon) which does:
`ManageIQ::Providers::Amazon::StorageManager::Ebs/S3`  so it is still obvious it is a Storage Manager.

jaywcarman: @iv1111 I put together a rename commit here (specs even pass, yea! :smile:):
https://github.com/iv1111/manageiq-providers-ibm_cloud/compare/cloud_object_storage...jaywcarman:cloud_object_storage_rename_manager

Also included is a commit for #202

agrare: 👍 that looks good @jaywcarman

iv1111: @agrare Our COS provider class is called ManageIQ::Providers::IbmCloud::ObjectStorage::Object**Manager**. We had to call it like that, since the UI code parses the classname in order to set the "kind" fild . It strips away the "Manager" part and leaves whatever is in front of it setting the "kind" value. Filtering by the value of "kind" is necessary, so that one would display only ObjectStorage providers supported by MIQ (and not e.g. Block Storage providers).

agrare: Oh you had to name it `ObjectManager`?  That's not good, block storage managers aren't named `BlockManager`.  I thought you were checking the `supports_object_storage?` feature.

iv1111: @agrare Yes, that's the reason. We were not able to use _supports_object_storage?_, because current API would not let one to get that value for a provider.

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -0,0 +1,211 @@
...
+                    :isRequired             => true,
+                    :validationDependencies => %w[type zone_id uid_ems],
+                    :fields                 => [
+                      {
+                        :component  => "text-field",
+                        :name       => "provider_region",
+                        :id         => "provider_region",
+                        :label      => _("Region"),
+                        :isRequired => true,
+                        :validate   => [{:type => "required"}],
```
agrare, line 40: Best for this to be a dropdown, is this the same list of regions as VPC?  See https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb#L42 and https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb#L65-L75

They are loaded from a yaml file: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/db/fixtures/ibm_cloud_vpc_regions.yml

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -0,0 +1,211 @@
...
+                        :name       => "uid_ems",
+                        :id         => "uid_ems",
+                        :label      => _("Resource Instance Id"),
+                        :isRequired => true,
+                        :validate   => [{:type => "required"}],
+                      },
+                      {
+                        :component  => "text-field",
+                        :name       => "endpoints.default.url",
+                        :id         => "endpoints.default.url",
```
agrare, line 53: Is a URL required or can it be built from the region?  Assuming we are talking to the service API here

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
```diff
@@ -0,0 +1,29 @@
...
+  def buckets
+    buckets = []
+
+    # XXX: filtering out buckets not from our region, as of now the REST call fails with an
+    # exception if used with from different region than the bucket's
+    (connection.list_buckets[:buckets] || []).each do |bucket|
+      connection.get_bucket_location(:bucket => bucket[:name]).location_constraint
+      buckets << bucket
+    rescue
+      _log.warn("bucket '#{bucket[:name]}' is not from our region '#{manager.provider_region}', skipping")
```
agrare, line 13: Is a bucket not in our region something that the user has misconfigured or is it totally normal?  If it is normal than warning is likely cause more alarm than necessary.  If normal we might not even need a log at all.

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
```diff
@@ -0,0 +1,29 @@
+class ManageIQ::Providers::IbmCloud::Inventory::Collector::ObjectStorage < ManageIQ::Providers::IbmCloud::Inventory::Collector
+  require_nested :StorageManager
+
+  def buckets
+    buckets = []
+
+    # XXX: filtering out buckets not from our region, as of now the REST call fails with an
+    # exception if used with from different region than the bucket's
+    (connection.list_buckets[:buckets] || []).each do |bucket|
+      connection.get_bucket_location(:bucket => bucket[:name]).location_constraint
```
agrare, line 10: Is there a way to check this that doesn't involve raising an exception?  A nicer way to do this might be

`connection.list_buckets[:buckets].to_a.reject { |bucket| bucket.in_my_region? # this part totally made up but you get the idea }`

app/models/manageiq/providers/ibm_cloud/inventory/parser/object_storage.rb
```diff
@@ -0,0 +1,53 @@
...
+      )
+    end
+  end
+
+  def process_bucket(bucket_id)
+    total_size = total_count = 0
+
+    BUCKET_TAB_LIMIT.times do
+      objects = collector.objects(bucket_id)
+      token = objects[:continuation_token]
```
agrare, line 32: This type of logic should live in the collector since it is related to actually fetching data from the provider API.  You can yield each object to the caller like:
```
collector.rb
def objects(bucket_id)
  params = {}
  1_000.times do
    objects = connection.list_objects_v2({:bucket => bucket_id}, params)
    params = {:continuation_token => objects[:continuation_token]}
    objects[:contents].to_a.each { |content| yield content }

    break if params[:contiuation_token].nil?
  end
end
```

Then your parser is just
```
parser.rb
collector.object_contents(bucket_id) do |content|
  ...
end
```

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -0,0 +1,211 @@
...
+    self.class.verify_bearer(region, endpoint, access_key, secret_key)
+    self.class.verify_iam(guid, apikey)
+
+    true
+  end
+
+  def self.verify_credentials(args)
+    guid     = args["uid_ems"]
+    auth_key = args.dig("authentications", "default", "auth_key")
+    apikey   = ManageIQ::Password.try_decrypt(auth_key)
```
agrare, line 147: We should make sure to handle the case where we are editing an existing provider (aka an "id" will be passed and none of the password fields will be present)

See https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb#L87

spec/factories/ext_management_system.rb
```diff
@@ -17,3 +17,9 @@
           :class  => "ManageIQ::Providers::IbmCloud::VPC::CloudManager",
           :parent => :ems_cloud
 end
+
+FactoryBot.define do
+  factory :ems_ibm_cloud_object_storage_object,
+          :class  => "ManageIQ::Providers::IbmCloud::ObjectStorage::ObjectManager",
```
agrare, line 23: Think you missed a rename here

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
iv1111: The list is a bit bigger than that for COS. There are regional, global and some other types of locations. See here: https://cloud.ibm.com/docs/cloud-object-storage?topic=cloud-object-storage-endpoints

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
iv1111: Seems like "list_bucket" returns buckets from all regions in a given instance and not from the current one. We need to filter those out as subsequent calls (e.g. list_objects, etc) will fail if used with buckets from a foreign region. Seems to be some arch. issue on the S3 side, I'm not totally sure. Similar issue is described here https://github.com/aws/aws-sdk-go/issues/720.

app/models/manageiq/providers/ibm_cloud/inventory/collector/object_storage.rb
iv1111: I'm not sure we can find out which buckets to filter out without using an "exception" handling block. @jaywcarman Who can we ask about this issue on our side?

app/models/manageiq/providers/ibm_cloud/inventory/parser/object_storage.rb
iv1111: Sure, we can refactor that.

spec/factories/ext_management_system.rb
iv1111: You're right. Thx. ;)

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
iv1111: I see. If I get the example right then it looks for the creds in the args and if not found then looks for the already existing/saved ones.

agrare: :+1: yes exactly

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -0,0 +1,211 @@
...
+    secret_key  = ManageIQ::Password.try_decrypt(secret_encr)
+    verify_bearer(region, endpoint, access_key, secret_key)
+
+    true
+  end
+
+  def self.verify_bearer(region, endpoint, access_key, secret_key)
+    begin
+      raw_connect(region, endpoint, access_key, secret_key).list_buckets({}, :params => {:max_keys => 1})
+    rescue Aws::Errors::ServiceError, Seahorse::Client::NetworkingError => err
```
chessbyte, line 163: Are we really expecting AWS errors here?

agrare: Yes, IBM uses an S3 compatible API so @iv1111 is using the `Aws::S3` gem: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/159/files#diff-ffe546bf7942848309735f1742f214e4af3c09c6bbacdf3ef0d1277981902b02R197-R209

agrare:
https://github.com/ManageIQ/manageiq-ui-classic/issues/7675 is now implemented, allowing the addition of standalone object storage managers

agrare (approved):

agrare:
We can fix any of these remaining issues in follow-up PRs, I'm going to assign to @jaywcarman for final review / merge since I have some commits in this branch

jaywcarman (approved):
LGTM, and everything is functional. I agree with @agrare's suggestion we address the remaining issues as follow-up PRs.

app/models/manageiq/providers/ibm_cloud/inventory/parser/object_storage.rb
jaywcarman: I'm working to clean up some of these old TODOs: #318

Issue #167, opened 2021-03-22
PowerVS API endpoint parsing needs to be updated for new regions


jaywcarman:
The PowerVS region naming and endpoint subdomain scheme has changed for new regions. Unfortunately, the [documentation](https://cloud.ibm.com/apidocs/power-cloud) is sparse and the [IBM Cloud resource controller API](https://cloud.ibm.com/apidocs/resource-controller/resource-controller) does not provide an instance's endpoint URL directly.

For the original regions, the CRN location segment was formatted as `{region}` or `{region}-{zone}` (e.g. `us-south` and `eu-de-1`).

New regions are formatted as `{region}{zone}`, without a `-` delimiter and always a two-digit zone (e.g. `tok04` and `syd05`).

I've already implemented an API endpoint override through user settings (bc64d12), but we should also update the [parser](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L45) to handle the new format.

PR #143, opened 2021-02-18
Fix PowerVS API endpoint special cases

jaywcarman:
fixes #133 

This PR fixes two bugs related to Power Cloud API endpoints:
- The CRN 'location' segment sometimes includes regional zone enumerations and needs to be parsed to get the correct endpoint subdomain (e.g. `eu-de-1` uses the endpoint URL `https://eu-de.power-iaas.cloud.ibm.com`)
- #133: The CRN 'location' segment is sometimes totally different than the endpoint URL (e.g. `dal12` -> `https://us-south.power-iaas.cloud.ibm.com`)

PR #168, opened 2021-03-22
Fix endpoint parsing for new PowerVS regions

jaywcarman:
New Power Virtual Servers regions are formatted as "{region}{zone}",
without a '-' and always with two zone digits.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb
@@ -43,6 +43,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::ManagerMixin
       url = api_endpoint_overrides[location.to_sym]
     else
       region = location.sub(/-\d$/, '')
+      region = region.sub(/\d\d$/, '')
       url = "#{region}.power-iaas.cloud.ibm.com"
     end
 


jaywcarman:
This issue, #167, is also present in `kasparov`, but this patch will not apply. Something similar to #143 will be required (I'll submit a PR soon).

Fryguy:
Marking this kasparov/yes and kasparov/conflict, so the automation will pick it up.  The "conflict" will be resolved when we merge #169

agrare:
@jaywcarman if the location -> region isn't consistent, I wonder if going forward it'd be better to have a simple lookup hash in settings that has e.g.:
```
{
  "eu-de-1" => {
    "name" => "Germany",
    "url" => "eu-de.iaas.whatever",
    "region" => "eu-de"
  }
}
```

agrare (approved):
:+1: for a short term fix

jaywcarman:
> @jaywcarman if the location -> region isn't consistent, I wonder if going forward it'd be better to have a simple lookup hash in settings that has e.g.:
> 
> ```
> {
>   "eu-de-1" => {
>     "name" => "Germany",
>     "url" => "eu-de.iaas.whatever",
>     "region" => "eu-de"
>   }
> }
> ```

@agrare I think you had mentioned that this type of config is used for manageiq-providers-amazon, so consistency would be nice. I also like the inclusion of human readable names. I'll add this to our to-do list for Morphy.

Fryguy:
Backport handled via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/169

Fryguy:
Backported to `lasker` in commit f723e1d846838fabf99f51fa88b2f3302585b8a1.

```text
commit f723e1d846838fabf99f51fa88b2f3302585b8a1
Date:   Tue Mar 23 15:21:03 2021 -0400

    Merge pull request #168 from jaywcarman/fix/powervs_new_region_endpoint_parsing
    
    Fix endpoint parsing for new PowerVS regions
    
    (cherry picked from commit eefc54c457e9f6d8d67b8439ab5c8fd62b7009d9)
```

PR #175, opened 2021-03-25
Set CloudVolumeType ems_id explicitly

jaywcarman:
'CloudVolumeType's should be children of a storage manager. The parent
'ems_id' value needs to be explicitly in the parser. If it is not set
the 'ems_id' default to null and new orphaned 'CloudVolumeType's are
created with _every_ refresh.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -270,6 +270,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
       persister.cloud_volume_types.build(
         :type        => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolumeType",
         :ems_ref     => v.type,
+        :ems_id      => persister.storage_manager.id,
         :name        => v.type,
         :description => v.description
       )


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -270,6 +270,7 @@ def cloud_volume_types
       persister.cloud_volume_types.build(
         :type        => "ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolumeType",
         :ems_ref     => v.type,
+        :ems_id      => persister.storage_manager.id,
```
jaywcarman, line 273: I'm not sure what changed to make it necessary to set this now (I know in the past the relationship to the parent storage manager got _auto-magically_ set).

I'll have to go back and test `lasker` and `kasparov` to see if this bug is present.

agrare:
Ah we should add this to the core storage manager collection here: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/storage_manager.rb with `add_common_default_values` like the other collections there

jaywcarman:
> Ah we should add this to the core storage manager collection here: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/storage_manager.rb with `add_common_default_values` like the other collections there

Adding it there will allow the relationship to be set without this patch?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
agrare: If these don't have a core persister collection which it looks like they don't then I'm guessing it was this: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/152/files#diff-e41645b87bcd5ddfe80fc3e2da3788b5a19b2da616dc10b96ea69715fbd4d344L68-L70

It looked indistinguishable from an override of a core collection to set the storage-manager instead of the cloud-manager.

agrare:
Correct, the `add_common_default_values` sets the `ems_id` to the storage manager automagically as you put it

jaywcarman:
Awesome, thanks @agrare. I need to logoff for the day. You're welcome to submit the change, else I'm happy to do it first thing tomorrow morning.

agrare:
https://github.com/ManageIQ/manageiq/pull/21126

PR #181, opened 2021-04-07
Changes in the volume_params so the volume can be created correctly

Kuldip-Nanda:
Changes in the volume_params so the volume can be created correctly.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -113,10 +113,11 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
 
   def self.raw_create_volume(ext_management_system, options)
     volume = nil
+    volume_params = nil
     ext_management_system.with_provider_connection(:service => 'PCloudVolumesApi') do |api|
       volume_params = IbmCloudPower::CreateDataVolume.new(
         'name'            => options['name'],
-        'size'            => options['size'] / 1.0.gigabyte,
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -113,10 +113,11 @@ def self.validate_create_volume(ext_management_system)
 
   def self.raw_create_volume(ext_management_system, options)
     volume = nil
+    volume_params = nil
     ext_management_system.with_provider_connection(:service => 'PCloudVolumesApi') do |api|
       volume_params = IbmCloudPower::CreateDataVolume.new(
         'name'            => options['name'],
-        'size'            => options['size'] / 1.0.gigabyte,
+        'size'            => options['size'].to_i / 1.0.gigabyte,
```
jaywcarman, line 120: Note: The value here ends up as a float. The Power Cloud backend will accept non-integer volume size values but will floor them into an integer. In the future we'd like to enhance the UI to provide input directly as an integer GB value. See the following discussion for details: https://github.com/ManageIQ/manageiq-ui-classic/pull/7399#issuecomment-781749487

jaywcarman (approved):

agrare (approved):

Fryguy:
Backported to `lasker` in commit 6d6540bffe9185ff33020872b77374787d4f1f34.

```text
commit 6d6540bffe9185ff33020872b77374787d4f1f34
Date:   Thu Apr 8 09:29:28 2021 -0400

    Merge pull request #181 from Kuldip-Nanda/master
    
    Changes in the volume_params so the volume can be created correctly
    
    (cherry picked from commit 91f7702936c112473a4401edef6ff2de9396ba5a)
```

PR #184, opened 2021-04-13
Remove default 'Network Interfaces' selection

jaywcarman:
The ':default' network selection is not functioning correctly. The
intention was to set the selection to 'None' (string) by default.
Regardless of what value is set to ':default' the provision form selects
_nothing_ by default, but displays the first item on the list (making it
appear as if a default selection has been made).

PowerVS VMs must be provisioning with at least one network interface.
The network can be selected from the 'Network Interfaces' pull-down or
by selecting the 'Attach to New Public Network' checkbox (or both). We
haven't been able to enforce these rules in the existing dialog, so the
_required_ 'Network Interfaces' selection includes a 'None' value to
allow selecting only the 'Attach to New Public Network'.

By removing the ':default' parameter from the dialog template the user
is forced to either select an existing network or 'None'. The 'None'
display value now includes a message calling out the requirement to
select 'Attach to New Public Network' since this is not enforced and
will cause the VM provisioning API call to fail.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -93,7 +93,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
   def allowed_subnets(_options = {})
     ar_subnets = ar_ems.cloud_subnets
     subnets = ar_subnets&.collect { |subnet| [subnet[:ems_ref], subnet[:name]] }
-    none = ['None', 'None']
+    none = ['None', 'None (Must attach to new public network)']
     Hash[subnets.unshift(none)]
   end
 
--- a/content/miq_dialogs/miq_provision_ibm_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_dialogs_template.yaml
@@ -172,7 +172,6 @@
             :method: :allowed_subnets
           :description: Network Interfaces
           :required: true
-          :default: None
           :display: :edit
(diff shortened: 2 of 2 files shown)


iv1111:
You're right. The 'default' is what has messed up the the values yesterday. Great that you've found it.

iv1111 (approved):

iv1111:
@jaywcarman Do I merge it or we wait for more approvals?

jaywcarman:
@iv1111 I've tested this, and tests looks green, so I think a merge is OK if you agree with the change.

Fryguy:
Backported to `lasker` in commit 34adaa9035c6b40b2e8e8a1ce921d1bb2d9a81cb.

```text
commit 34adaa9035c6b40b2e8e8a1ce921d1bb2d9a81cb
Author: Isgandar Valizada <69972357+iv1111@users.noreply.github.com>
Date:   Fri Apr 16 09:10:24 2021 +0200

    Merge pull request #184 from jaywcarman/fix/powervs_provision_default_network
    
    Remove default 'Network Interfaces' selection
    
    (cherry picked from commit c851b3f1d0b1ee6a4772a0729fd6a651ff5da1a2)
```

Fryguy:
Backported to `kasparov` in commit a1b48987584f1861963c45b01e04ec856e0e9b32.

```text
commit a1b48987584f1861963c45b01e04ec856e0e9b32
Author: Isgandar Valizada <69972357+iv1111@users.noreply.github.com>
Date:   Fri Apr 16 09:10:24 2021 +0200

    Merge pull request #184 from jaywcarman/fix/powervs_provision_default_network
    
    Remove default 'Network Interfaces' selection
    
    (cherry picked from commit c851b3f1d0b1ee6a4772a0729fd6a651ff5da1a2)
```

PR #187, opened 2021-04-16
Use consistent value types for 'affinity_policy'

jaywcarman:
The Power Cloud API new volume create call
(https://cloud.ibm.com/apidocs/power-cloud#pcloud-cloudinstances-volumes-post)
can set `affinityPolicy` to only 'affinity' or 'anti-affinity'. When no
affinity policy is required the 'affinityPolicy' is omitted (or set to
nil).

I attempted to implement this directly in params_for_create ddf
definition, but for some reason the 'affinity_policy' :label was set in
the resulting options instead of its :value. This caused the API call to
send '"affinityPolicy": "Off"', which is invalid.

This fix sets the off :value to 'Off' (same as the :label), both to be
consistent and to eliminate the need to distinguish between :label and
:value. That change requires additional logic in the 'raw_create_volume'
method in order to handle the 'Off' case correctly.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -54,7 +54,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
           :name         => 'affinity_policy',
           :id           => 'affinity_policy',
           :label        => _('Affinity Policy'),
-          :initialValue => nil,
+          :initialValue => 'Off',
           :condition    => {
             :when => 'edit',
             :is   => false,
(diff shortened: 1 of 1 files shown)


iv1111 (approved):

agrare:
> I attempted to implement this directly in params_for_create ddf
definition, but for some reason the 'affinity_policy' :label was set in
the resulting options instead of its :value. This caused the API call to
send '"affinityPolicy": "Off"', which is invalid.

This sounds like a bug that we should track down, is it a DDF core issue or an issue in how our the params are defined.  :+1: on this as a workaround though

jaywcarman:
@agrare agree - I'll write up an issue so we can track it.

I purposefully set this to `nil` in #129, but can't remember how thoroughly I tested the change (did it work and has been broken since? Did I push a breaking change without testing it :flushed:).

Fryguy:
Backported to `lasker` in commit f2faf105bb6a011767f8cc1d4db5c2eb0a5ebe21.

```text
commit f2faf105bb6a011767f8cc1d4db5c2eb0a5ebe21
Date:   Mon Apr 19 10:32:43 2021 -0400

    Merge pull request #187 from jaywcarman/fix/volume_create_form
    
    Use consistent value types for 'affinity_policy'
    
    (cherry picked from commit 6a64554bfed88de04a40c7edd1d9a969215f118a)
```

PR #189, opened 2021-04-20
Cloud tool authentication

jaredm-ibm:
# Issue
Long running instance does not automatically update token when expired. The original implementation checked expiry on init of a class, but since cloud tool memoizes the child classes this is insufficient for the desired goal. The new implementation checks expiry before every request. This is beneficial when using a REPL for testing code chunks.

# Priority
* Required to add resource groups to CloudTool for provisioning.

# Implementation
* On each request check the expiry. If the token is expired then fetch a new one.
* Trivial change to the timeout from 10 to 60 seconds. The tagging API sometimes takes more than 10 seconds to respond.

# Not covered
* n/a

# Specs
* Updated cloud_tools_spec with new functionality.
* New test to check that an instantiated vpc instance will refresh when the expiry_time has elapsed.
* Updates to existing tests to accommodate the changed behaviour.
* Added and updated VCRs for new test functionality.

# Common files
* Add `define_cassette_placeholder` to `spec_helper` for `IBM_CLOUD_VPC_API_KEY`. Without it the API Key is saved to the VCR file.

### DIFF
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -61,27 +61,37 @@ module ManageIQ
             # @return [void]
             def initialize(bearer_info, logger: nil)
               @logger = define_logger(logger)
-              @bearer_info = verify_info(bearer_info)
+              @bearer_info = bearer_info
               super({:bearer_token => bearer_info[:token]})
             end
 
             # @return [Hash{Symbol=>String, Integer}] bearer hash with token, expire_time and api_key as keys.
             attr_reader :bearer_info
 
+            def authenticate(headers)
+              verify_bearer
+              super(headers)
+            end
+
             private
 
-            # Verify that the bearer token hasn't expired. If it has and an API Key is present try to get a new bearer token.
-            #
-            # @return [Hash{Symbol => String, Integer}] @see #bearer_info
-            def verify_info(bearer_info)
-              # Raise standard error if expiration time expires in the next 10 second.
-              if expired?(bearer_info[:expire_time])
-                raise 'Bearer token has expired.' if bearer_info[:api_key].nil?
-
-                @logger.info('Bearer token expired. Fetching new one using API Key.')
-                return IamAuth.new(bearer_info[:api_key]).bearer_info
+            # Check to see if bearer token has expired. If it has fetch a new one.
+            # @return [void]
+            def verify_bearer
+              if expired?(@bearer_info[:expire_time])
+                @bearer_info = new_bearer(@bearer_info)
+                @bearer_token = @bearer_info[:token]
               end
-              bearer_info
+            end
+
+            # Get a new bearer token.
+            # @param bearer_info [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
+            # @return [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
+            def new_bearer(bearer_info)
+              raise 'Bearer token has expired.' if bearer_info[:api_key].nil?
+
+              @logger.info('Bearer token expired. Fetching new one using API Key.')
+              IamAuth.new(bearer_info[:api_key]).bearer_info
             end
 
             # Check to see if the token expiry time has elapsed.
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -102,6 +102,7 @@ module ManageIQ
               return true if expire_time.nil?
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
+              # True if now is greater than expiry time. False if now is less than expire time.
               Time.now.to_i >= (expire_time - 10)
             end
 
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -99,11 +99,13 @@ module ManageIQ
             #
             # @return [Boolean] True is expired. False is valid.
             def expired?(expire_time)
-              return true if expire_time.nil?
+              return false if expire_time.nil? # Assume that expire check is disabled.
+
+              return true unless expire_time.match?('^\d+$') # Verify that expire_time can be converted into an integer.
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
               # True if now is greater than expire time. False if now is less than expire time.
-              Time.now.to_i >= (expire_time - 10)
+              Time.now.to_i >= (expire_time.to_i - 10)
             end
 
             # Define a new logger.
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -102,7 +102,7 @@ module ManageIQ
               return true if expire_time.nil?
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
-              # True if now is greater than expiry time. False if now is less than expire time.
+              # True if now is greater than expire time. False if now is less than expire time.
               Time.now.to_i >= (expire_time - 10)
             end
 
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -101,7 +101,12 @@ module ManageIQ
             def expired?(expire_time)
               return false if expire_time.nil? # Assume that expire check is disabled.
 
-              return true unless expire_time.match?('^\d+$') # Verify that expire_time can be converted into an integer.
+              return true unless expire_time.respond_to?(:to_i) # Verify that expire_time can be converted into an integer.
+
+              # If expire_time is a string then ensure that it only has digits.
+              if expire_time.respond_to?(:match?)
+                return true unless expire_time.match?('^\d+$') # rubocop:disable Style/SoleNestedConditional
+              end
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
               # True if now is greater than expire time. False if now is less than expire time.
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
@@ -88,14 +88,14 @@ module ManageIQ
             # @param bearer_info [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
             # @return [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
             def new_bearer(bearer_info)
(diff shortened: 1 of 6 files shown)


spec/models/manageiq/providers/ibm_cloud/vpc/cloud_tools_spec.rb
```diff
@@ -146,15 +163,44 @@ def create_bearer_info(valid_expire_time: true, set_api_key: true)
...
+  # Test that the VPC client can be instantiated.
   it 'can get a VPC client' do |example|
     VCR.use_cassette("#{described_class.name.underscore}/#{example.description}") do
       expect(described_class.new(:api_key => api_key).vpc(:region => 'us-east').client).to be_a(IbmVpc::VpcV1)
     end
   end
 
+  # Test that a initialized VPC class can refresh its token if expired.
+  it 'can get a VPC client with expired auth' do |example|
+    VCR.use_cassette("#{described_class.name.underscore}/#{example.description}") do
```
Fryguy, line 175: Not for this PR, but there is a way to do this part automatically, I believe.  If I recall, you can do the following:

```suggestion
  it 'can get a VPC client with expired auth', :with_vcr => true do
```

This wraps the entire test in a well-defined name based on the example description

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
```diff
@@ -92,6 +102,7 @@ def expired?(expire_time)
               return true if expire_time.nil?
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
+              # True if now is greater than expiry time. False if now is less than expire time.
```
agrare, line 105: We should be consistent with either `expiry time` and `expire time`

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
```diff
@@ -61,27 +61,37 @@ class BearerAuth < IbmVpc::Authenticators::BearerTokenAuthenticator
...
+                @bearer_token = @bearer_info[:token]
               end
-              bearer_info
+            end
+
+            # Get a new bearer token.
+            # @param bearer_info [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
+            # @return [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
+            def new_bearer(bearer_info)
+              raise 'Bearer token has expired.' if bearer_info[:api_key].nil?
```
agrare, line 91: A little confusing that the exception message here is the token has expired, but the log message [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/189/files#diff-64f4a90e4d7c883bcf2208d05e76761872252560a972a7f4bfc46c084cf2e7baR93) is that it has expired.

Would it be more accurate to say `Cannot refresh bearer token without an API key` or something to that effect?

agrare: Is not providing an api_key a valid scenario?  Could we just ensure that the api_key is always passed in or are there some places where we have to only provide a bearer token?

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_tools_spec.rb
jaredm-ibm: I was playing with that when I created this the first time. The directory structure wasn't being created correctly. I can't remember what was actually generated. I can play with it later.

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
jaredm-ibm: Updating message to say 'Bearer token has expired and unable to refresh. An api key is not present in bearer_info hash.'

The api key should be automatically included when fetching the bearer_info via standard means. However it is just a hash and there may be times when that hash doesn't have the api key. I don't want to validate the hash because all that is necessary to work is a valid bearer_token.  If it is a short-lived operation this may be all that is necessary. The presence of the bearer_token is validated on init by super. The presence of expire_time is validated in `expired?`. Finally the presence of api_key is validated during `new_bearer`.

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_tools_spec.rb
agrare: > The directory structure wasn't being created correctly.

I wonder if the issue is this spec is in `spec/models` but the code lives in `lib/` not `app/models`

jaredm-ibm: If I remember correctly is was creating new directories for each namespace in the class hierarchy.

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
jaredm-ibm: Actually, my explanation about the short-lived is no longer true. All the keys are validated just not at once. This change makes it impossible to send a request unless all the keys are present and valid.

jaredm-ibm: If we wanted to we could return false on expire_time.nil? for expired? . This would allow for a simple bearer_token to be passed in. If it is expired then an 401 will be raised.

jaredm-ibm:
Ignore my last update. My integer check is broken.

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
```diff
@@ -61,38 +61,51 @@ class BearerAuth < IbmVpc::Authenticators::BearerTokenAuthenticator
...
             def expired?(expire_time)
-              return true if expire_time.nil?
+              return false if expire_time.nil? # Assume that expire check is disabled.
+
+              return true unless expire_time.match?('^\d+$') # Verify that expire_time can be converted into an integer.
 
               # Checks to see if the expire time will elapse in the next 10 seconds.
-              Time.now.to_i >= (expire_time - 10)
+              # True if now is greater than expire time. False if now is less than expire time.
+              Time.now.to_i >= (expire_time.to_i - 10)
```
agrare, line 108: Would `Integer(expire_time)` work?  It'll raise an exception if expire_time isn't a valid integer, not sure exactly what case you're trying to handle with https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/189/files#diff-64f4a90e4d7c883bcf2208d05e76761872252560a972a7f4bfc46c084cf2e7baR104 though

jaredm-ibm: In the discussion of validation it occurred to me that I'm not checking to ensure that expire_time is an integer. This check will remedy it. I don't want to throw an exception, I want it to just get a new token.

spec/models/manageiq/providers/ibm_cloud/vpc/cloud_tools_spec.rb
Fryguy: Yeah, it's possible it won't line up exactly with what you have now, but it might be cleaner to just follow the convention.

jaredm-ibm: Hah. I'm all for using the `vcr` metadata. That's what I used to configure my tests in the SDK.  I was just following the convention used by the other tests because I thought that was the standard for ManageIQ. If we want to convert to the metadata format we should update all the tests in this repo to use the same method.

Fryguy: I don't disagree :)

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
agrare: This method is starting to get more and more complicated trying to handle nil, integer, integer-as-a-string, random string...

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
```diff
@@ -61,38 +61,56 @@ class BearerAuth < IbmVpc::Authenticators::BearerTokenAuthenticator
...
+            # @return [Hash{Symbol=>String, Integer}] Bearer info hash from IamAuth
+            def new_bearer(bearer_info)
+              raise 'Bearer token has expired and unable to refresh. An api key is not present in bearer_info hash.' if bearer_info[:api_key].nil?
+
+              @logger.info('Bearer token expired. Fetching new one using API Key.')
+              IamAuth.new(bearer_info[:api_key]).bearer_info
+            end
+
+            # Check to see if the token expire_time has elapsed.
+            # @param expire_time [Integer, NilClass] The token expire_time provided by IAM.
```
agrare, line 98: This indicates expire_time should be an integer or nil, would it make sense to convert from a string to an integer elsewhere so it is only ever an int when it hits this method?  That would clean this up a lot

agrare (approved):
Nothing here that can't be handled in a follow-up

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
jaredm-ibm: Externally supplied data. It should be an int, but can't know for sure. Unfortunately doc hints aren't the same as type hints.

lib/manageiq/providers/ibm_cloud/cloud_tools/authentication.rb
jaredm-ibm: I can probably remove the string test since to_i returns 0 if it can't convert the integer cleanly. Which would do what I want.

Fryguy:
@jaredm-ibm A conflict occurred during the backport of this pull request to `morphy`.

Conflict details:

```diff
diff --cc spec/spec_helper.rb
index bc6a94f,ba767e8..0000000
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@@ -45,6 -45,9 +45,12 @@@ VCR.configure do |config
    config.ignore_hosts('codeclimate.com') if ENV['CI']
    config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
  
++<<<<<<< HEAD
++=======
+   # Used to replace the API Key with the placeholder in the saved VCR.
+   config.define_cassette_placeholder('IBM_CLOUD_VPC_API_KEY') { Rails.application.secrets.ibm_cloud_vpc[:api_key] }
+ 
++>>>>>>> c153355... Merge pull request #189 from jaredm-ibm/cloud_tool_authentication
    config.before_record do |i|
      replace_token_contents(i.response) if i.request.uri == "https://iam.cloud.ibm.com/identity/token"
      vpc_sanitizier(i) if i.request.uri.match?('iaas.cloud.ibm')
```

agrare:
@Fryguy backport PR https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/273

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/273

PR #191, opened 2021-04-22
Vpc spec vcr metadata

jaredm-ibm:
# Issue
When configured to use RSpec metadata, VCR can automatically define cassettes. This leads to simpler specs with less boilerplate.

# Priority
* For resource controller functionality the specs will need to be updated.

# Implementation
* Configure the spec_helper to use RSpec metadata.
* Add `:vcr` to each root `describe`.
* For refresher spec add `:allow_playback_repeats` so that the second iteration can use the same data.

# Not covered
* n/a

# Specs
* All VPC specs have been updated with this structure.
* All previous VCRs have been regenerated.

# Common files
* spec_helper updated with vcr metadata directives.


spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-describe ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher do
+# rubocop:disable Style/MethodCallWithArgsParentheses # Guidance does not conform to preferred expect formatting.
```
agrare, line 3: This is already excluded by manageiq-style's rubocop base file and isn't necessary:
```
spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb:3:1: W: [Correctable] Lint/RedundantCopDisableDirective: Unnecessary disabling of Style/MethodCallWithArgsParentheses.
# rubocop:disable Style/MethodCallWithArgsParentheses # Guidance does not conform to preferred expect formatting.
```

jaredm-ibm: My local rubocop warns me of this, thus the reason I disable it. I also get warnings about `{key: "b"}` should be `{:key => 'b'}` which means that my local rubocop is using the manageiq style. Might want to check that the directive is working as expected.

agrare: If I delete these `rubocop:disable` directives locally then rubocop doesn't warn about this, miq-bot also doesn't so it seems more likely an issue with your local rubocop.

jaredm-ibm: Dunno, it could be the way that VS Code calls rubocop messes with the paths. There have also been a number of fixes to this cop between the version run in miq-bot and the latest. Maybe miq-bot just doesn't see it because it is broken in that build.  In any case, when I am developing without these disables I get a ton of false positives. For me it is more important to have this work in VS Code than it is to have miq-bot pass.

agrare: This is working in Atom, not sure about Rubymine (@Fryguy ?), so probably vscode.  I see a number of issues with the vscode plugin not honoring inherit_from directives (e.g. https://github.com/rubyide/vscode-ruby/issues/227#issuecomment-694394518)

agrare: We shouldn't be adding these to the code base because an editor isn't honoring `.rubocop.yml` correctly.

Fryguy: I use Sublime - I had to just tweak the rubocop plugin to `bundle exec rubocop` instead of straight `rubocop` so it uses the right one, but that works

Fryguy: (actually to be honest, I think I just installed 0.82.0 and use that only, and didn't bother tweaking the plugin... our app will automatically install 0.82.0 via manageiq-style as a dependency, and I just intentionally don't install a newer version)

Fryguy: Also see https://github.com/rubyide/vscode-ruby/issues/227#issuecomment-740521812

agrare: And for atom you just have to drop
```
  "linter-rubocop":
    useBundler: true
```
in `.atom/config.cson` (or use the UI settings for the rubocop plugin).

I wonder if it'd be helpful to have this in the developer_setup guide

Fryguy: I recall in the past, there was a weird issue with sublime where it launches a non login shell to run its plugin scripts, and therefore would use the system Ruby instead of my Ruby with my Ruby version manager.  The two options to fix were to either install the correct rubocop in the system Ruby _or_ change my zprofile to switch to the expected Ruby.  I did the latter and then never looked back, because that also changes out for all my login shells as well.

jaredm-ibm: The proposed workaround for VSCode doesn't seem to work. Considering the issue has been open since 2017, I doubt there will be a resolution anytime soon. So that leaves 2 options:
1. Add this cop to the rubocop_local with the spec exclude.
2. Leave the disables in the specs.

I did 1 already and had the change removed, thus the reason I'm doing 2.

jaredm-ibm:
The VSCode plugin is the most up-to-date. The defect Jason mentioned has been around since 2017 and is still being commented on. I don't see how `bundle exec rubocop` helps with linting inside VS Code. I'm not making my development process more complicated so that miq-bot can be clean. This change makes development easier, is only noticeable by  people updating these specific specs and not running VS Code or running rubocop manually. The linter configuration  changes no functional elements. This is a moot discussion and holding up getting real functional work done.

agrare:
> The VSCode plugin is the most up-to-date. The defect Jason mentioned has been around since 2017

The VS Code plugin doesn't support well-established rubocop functionality, the fact that it hasn't been fixed in 4 years makes me less likely to want to merge this.

> I'm not making my development process more complicated so that miq-bot can be clean.

This is the community's code not just yours.  I'm sorry that your preferred editor doesn't have good rubocop integration but the solution to that isn't to markup the code to work around that.

If the vscode plugin doesn't work with rubocop inherit_gem, the command line version of `rubocop` certainly does.

agrare:
Maybe I'm missing something, does running `bundle exec rubocop` from a terminal on that file show the same error?

My understanding is this isn't about miq-bot at all because it is working for me with the latest rubopcop as well as the version in the miq bundle which means this is a vscode issue

jaredm-ibm:
Chad had a good idea to add the directive to rubocop_local and mark it assume unchanged. Problem fixed.

agrare (approved):

Fryguy:
@jaredm-ibm A conflict occurred during the backport of this pull request to `morphy`.

Conflict details:

```diff
diff --cc app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
index 9ccebf1,d46b165..0000000
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/vpc.rb
@@@ -30,8 -29,6 +30,11 @@@ class ManageIQ::Providers::IbmCloud::In
  
    def image(image_id)
      connection.request(:get_image, :id => image_id)
++<<<<<<< HEAD
 +  rescue IBMCloudSdkCore::ApiException
 +    nil
++=======
++>>>>>>> 4324823... Merge pull request #191 from jaredm-ibm/vpc_spec_vcr_metadata
    end
  
    def keys
* Unmerged path spec/vcr_cassettes/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher.yml
```

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/274

PR #192, opened 2021-04-25
Fix SAP profile (flavor) memory units

jaywcarman:
The Power Cloud API returns SAP profile memory in gigabytes, but the flavor table stores memory in bytes.

I also included 1dbb48b to write the core value to both `cpus` and `cpu_cores`, to be consistent with Azure and IBM Cloud VPC providers:
https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/inventory/parser/cloud_manager.rb#L43-L44
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb#L182-L183

But `manageiq-providers-google` omits it:
https://github.com/ManageIQ/manageiq-providers-google/blob/master/app/models/manageiq/providers/google/inventory/parser.rb#L73

And `manageiq-providers-azure_stack` seems to treat `cpu_cores` == sockets. (bug?)
https://github.com/ManageIQ/manageiq-providers-azure_stack/blob/master/app/models/manageiq/providers/azure_stack/inventory/parser/cloud_manager.rb#L36-L37

Thoughts @agrare ?

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -255,7 +255,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
           :ems_ref     => value.profile_id,
           :name        => value.profile_id,
           :cpus        => value.cores,
-          :memory      => value.memory,
+          :memory      => value.memory&.gigabytes,
           :description => description
         )
       end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -44,7 +44,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
 
   def sap_flavor_memory(_options = {})
     if sap_flavor
-      {"2" => sap_flavor.memory}
+      {"2" => (sap_flavor.memory / 1.0.gigabyte).to_i}
     end
   end
 


jaywcarman:
Marked as WIP because I realized I need to convert this value back to GB in the provisioning form:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb#L47

agrare:
@jaywcarman the cpu_cores attribute of the Flavor isn't used for anything other than display outside of a provider repo, so it is up to you if you want to set it or leave it blank.

Most cloud providers don't expose the concept of sockets/cores/hyperthreads they just give you a total number of vcpus.  For providers which do expose this property it is available to be set.

I agree I think azure stack has this backwards, the names are a little vague and I think the vm_or_template/hardware model has this better named: cpu_sockets, cpu_total_cores, cpu_cores_per_socket.

Looking at amazon (likely the first to implement flavors) https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/inventory/parser/cloud_manager.rb#L351-L352 it appears that `cpu_cores` is actually intended to be `cpu_cores_per_socket`

I'll open an issue in core to make sure that we're consistent with what cpus/cpu_cores means and propose renaming the cpu_cores column to reduce confusion.

As for this PR, up to you if you want to set it or not it isn't required.

agrare:
@jaywcarman https://github.com/ManageIQ/manageiq/issues/21192

jaywcarman:
> As for this PR, up to you if you want to set it or not it isn't required.

Thanks @agrare. I went ahead and removed setting `cpu_cores` here to keep things clear and simple.

agrare (approved):

agrare:
@jaywcarman do you need this backported to `lasker` ?

jaywcarman:
> do you need this backported to lasker ?

@agrare yes, please!

Fryguy:
Backported to `lasker` in commit 7049c093ecb3610124152dc3261c7d768586fd73.

```text
commit 7049c093ecb3610124152dc3261c7d768586fd73
Date:   Mon Apr 26 08:58:25 2021 -0400

    Merge pull request #192 from jaywcarman/fix_sap_flavors_data
    
    Fix SAP profile (flavor) memory units
    
    (cherry picked from commit d036a4dfccf6c4d570ee7ca98cb24e35ae7fd608)
```

PR #196, opened 2021-04-26
[LASKER] Fix AvailabilityZone and Flavor duplication

agrare:
Refreshing the CloudManager and the NetworkManager was causing inventory collections which used the default default_values to set the ems_id to set it to the manager being targeted, not the manager owning the inventory_collection.


agrare:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/152 fixed this by moving all this logic into a common place in core, but not sure we want to backport https://github.com/ManageIQ/manageiq/pull/21099 also so this is a more targeted fix

jaywcarman (approved):
I applied the patch and verified this fixes the PowerVS issues discussed in Gitter. Thanks again @agrare !

Issue #122, opened 2020-11-23
Add VPC VM provisioning


jaredm-ibm:
Add provisioning support to the VPC plugin.

- [x] Understand the provisioning process.
- [x] Successfully walk a dummy provision through the workflow.
- [x] Ensure all required data is stored in persister.
  - [x] Store UUID of SSH key in DB.
  - [x] Use cloud_volume_types for volume profiles.
  - [x] Add availability_zone relation to subnets.
  - [x] Add resource groups
- [x] Customize the dialogs to reflect the VPC requirements.
  - [x] Add general fields
  - [x] Only allow subnet selection when both zone & vpc have been selected.
  - [x] Add resource group selection.
  - [x] Add cloud volume selection.
  - [x] Add cloud volume creation.
  - [x] Add new cloud volume validation.
- [x] Build the API payload.
  - [x] Basic instance.
  - [x] w/ new & existing volumes
  - [x] Resource groups 
- [x] Perform full provision.
  - [x] Generate VM with default resource groups.
  - [x] Generate VM with volume attachment.
  - [x] Generate VM with multiple volume attachments.
  - [x] Generate VM with new volume attachment.
  - [x] Generate VM with multiple new volume attachments.
  - [x] Generate VM with selectable resource groups.
- [x] Final dialog scenarios
  - [x] VM Name can only be 2 lower-case characters, followed by lower-case characters, numbers or dash.
  - [x] Unselected resource group is not in payload.
  - [x] Zone & VPC must be selected to populate subnets.
  - [x] Zone must be populated to populate volumes.
  - [x] Volume name can only be 2 lower-case characters, followed by lower-case characters, numbers or dash.
  - [x] Volume size can only be numbers between 10 & 2000.
- [x] Final provision scenarios
  - [x] Image ID is UUID of image
  - [x] Resource group can be nil.
  - [x] Naming conventions are followed.
  - [x] Multiple volumes can be attached.
  - [x] Multiple volumes can be created and then attached.

A modified example payload using pseudocode:
```
    {
      :name                      => get_option(:vm_target_name),
      :keys                      => [{:id => get_option(:guest_access_key_pair)}],
      :profile                   => {:name => get_option_last(:provision_type)},
      :image                     => {:id => vm_image[:ems_ref]},
      :zone                      => {:name => get_option_last(:placement_availability_zone)},
      :vpc                       => {:id => get_option(:cloud_network)},
      :primary_network_interface => {
[...]

PR #198, opened 2021-04-30
Vpc provision

jaredm-ibm:
# Issue
Provide an interface for provisioning a VPC instance.
Main issue is https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/122.

# Priority
* Mostly self-contained.

# Implementation
* Logger - Mixin that will switch loggers depending on build type.
  * For development use the $ibm_cloud_log
  * For production use the _log
  * Methods mimic the two loggers, but adds a standard  `class.method` to each message.
  * For log_backtrace there is an optional contextual msg, to help with troubleshooting, and the option of logging without raise.
* Provision Workflow - UI for provision.
  * Each section of the UI is broken into their own mixin. This keeps the file length shorter and allows for grouping of like methods.
  * The majority of the work is done in fields.rb. These methods standarise the creation of hashes for the UI to process.
  * Volumes is a unique case and requires manual parsing of the returned form details.
  * Standard validation methods are unavailable because they aren't MiqDialog fields, so an exception is thrown on validation error.
  * The UI will first process the standard validation. If it fails those errors will be surfaced.
  * The UI will then process the volumes validation. If it fails those errors will be surfaced.
* Provision - Actual provisioning steps.
  * Broke cloning into cloning and payload.
  * Cloning holds the steps to do and validate the provision.
  * Payload creates the Hash.
  * StateMachine handles the flow.

# Not covered
* See issue for not in scope.

# Specs
* Manual testing see issue for results.
* I'll look into what can be spec tested later.

# Common files
* No common files changed


jaredm-ibm:
Fixed rubocop errors via force-push.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+# Encapsulate logging methods into a single mixin.
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::LoggingMixin
+  # A logger wrapper that can be used as a general interface during dev or production.
+  # During development use the defined $ibm_cloud_log which prints to ibm_cloud.log.
+  # In production use the defined _log which prints to general evm log and should be captured in containers.
```
agrare, line 7: Anything logged to `$ibm_cloud_log` will be printed to stdout in containers since it is a multicast logger

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+# Encapsulate logging methods into a single mixin.
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::LoggingMixin
```
agrare, line 4: Hey @jaredm-ibm this looks like a change which should be extracted and discussed in its own PR

I think it isn't necessary due to how broadcast logging is handled in containers and appliances but I don't want to hold up the provisioning PR while we discuss that.

If there are other issues with `_log` that you're working around it'd be great to try to solve those generally for everyone.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'json'
+
+# Provide all configurable aspects of the actual provision operation.
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision::Cloning
+  # Called during provision. Prints message to log file.
+  # @param clone_options [Hash] Options used for clone operation.
+  # @return [Boolean] The result of the log message.
+  def log_clone_options(clone_options)
```
agrare, line 10: I'm surprised there isn't a generic version of this method in the base `ManageIQ::Providers::CloudManager::Provision::StateMachine` class, it could look very much like what you have here.

It's not good that the base class calls a method that isn't defined at all, not even one that raises an unimplemented exception

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
```diff
@@ -0,0 +1,61 @@
...
+  rescue => e
+    logger(__method__).log_backtrace(e, :context_msg => 'Error while logging clone options for VPC provision.')
+  end
+
+  # Send the final product to the Cloud provider.
+  # @param clone_options [Hash] Payload to send to provider.
+  # @raise [MiqException::MiqProvisionError] An error was returned by the SDK.
+  # @return [String] The ID of the new instance.
+  def start_clone(clone_options)
+    logger(__method__).debug("Json for VPC provision task. #{JSON.dump(clone_options)}")
```
agrare, line 22: The `Vmdb::LogProxy` class dynamically determines the calling method with `#caller_locations` so if you use that you shouldn't have to pass in `__method__`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
jaredm-ibm: When I was trying to debug that issue with the VM state failing, none of the log messages to ibm_cloud_log were showing up. So I did this.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
```diff
@@ -0,0 +1,61 @@
...
+    live_status = instance[:status]
+    return false, _('IBM VPC instance provision has no status present.') if live_status.nil?
+
+    status = live_status.downcase
+    if status == 'failed'
+      error_msg = _('An error occurred while provisioning the IBM VPC instance. Instance has failed status. Check the instance in cloud.ibm.com for more information.')
+      raise MiqException::MiqProvisionError, error_msg
+    elsif status == 'running'
+      return true, _('The IBM VPC instance has been provisioned and has a running status.')
+    elsif %w[pausing pending restarting resuming starting stopping].include?(status)
```
agrare, line 50: This looks like a good candidate for `case/when` instead of `if/elsif/else`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
agrare: I commented out the 
```
     private_class_method def self.create_container_logger
-      return unless ENV["CONTAINER"]
-
       require "manageiq/loggers/container"
       ManageIQ::Loggers::Container.new
     end
```

as a test and:
```
> $ibm_cloud_log.info("hello world")
{"@timestamp":"2021-04-30T11:35:21.416641 ","pid":170242,"tid":"1773c","level":"info","message":"hello world
```

is printed to stdout.

jaredm-ibm: At the moment the error flow uses this implementation. I don't want to rethink it right now. If a better solution comes along then this can be ripped out and rethought.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: Dunno, I took the method from the Power implementation and reworked it.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/options_helper.rb
```diff
@@ -0,0 +1,13 @@
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision::OptionsHelper
+  # The ID for this EMS.
+  # @return [Integer]
+  def cloud_instance_id
+    source.ext_management_system.uid_ems
```
agrare, line 5: Are you trying to return the `:id` or the `:uid_ems` here?  The comment indicates you want the `:id` but this is returning the `:uid_ems` string.

PowerVS sets `:uid_ems` to the pcloud_guid but I don't believe VPC sets this to anything.

agrare: Actually is the `#cloud_instance_id` method used anywhere here?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: I'm not seeing duplication in my logs, and so I don't think that is true.

```
[----] I, [2021-04-01T15:19:41.884962 #51274:3fdf4aadae08]  INFO -- : ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow.parse_new_volumes_fields Values are {:initial_pass=>true, :src_vm_id=>[36, "ibm-centos-7-0-minimal-amd64-1"], :miq_request_dialog_name=>"miq_provision_vpc_dialogs", :current_tab_key=>:volumes, :ip_addr=>nil, :user_script_text=>nil, :provision_type=>[1, "bx2-2x8"], :vm_name=>"avdd", :placement_auto=>[false, 0], :placement_availability_zone=>[1, "us-east-1"], :cloud_network=>["r014-0fa2acc6-2a41-4f2b-9c89-bcea07cdcbc3", "b-vpc-washington-dc"], :cloud_subnet=>["0757-ef523a2f-5356-42ff-8a78-9325509465b9", "b-subneet-washington-dc-1"], :resource_group=>nil, :storage_type=>["general-purpose", "general-purpose"], :guest_access_key_pair=>["r014-14c6c21a-a33d-4323-863b-439561483102", "b-goo-washington-dc-key"], :cloud_volumes=>[nil, nil], :volume_name=>nil, :volume_resource_group=>"default", :volume_size=>nil, :volume_profile=>"general-purpose", :vm_tags=>[], :src_vm_nics=>nil, :src_vm_lans=>[], :src_ems_id=>[2, "Test"], :schedule_time=>nil, :volume_name_1=>"abc", :volume_resource_group_1=>"abc", :volume_profile_1=>"abc", :volume_size_1=>"20"}
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare: Is this essentially a duplicate of line 22 [[ref]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198/files#diff-ce698cd249242f93ff57c4152d96a14261fffb66e561b128da86f060bbc78414R22)

It looks like they both log `JSON.dump(clone_options)`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/options_helper.rb
jaredm-ibm: Dunno, that is another copy paste from Power.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/payload.rb
```diff
@@ -0,0 +1,96 @@
...
+  rescue => e
+    logger(__method__).log_backtrace(e)
+  end
+
+  # Add a resource group if it is selected. If not selected then VPC will use the default resource_group.
+  # @param payload [Hash] Payload for provision.
+  # @return [Hash] Payload for provision with optional resource_group set.
+  def resource_group(payload)
+    resource_group = get_option(:resource_group)
+    payload[:resource_group] = {:id => resource_group} unless resource_group.nil?
```
agrare, line 39: If this is going to modify the argument the method should be `resource_group!`

jaredm-ibm: It's flow control not something that gets called by itself. I did have them as private methods but if I'm able to write specs on this I want to be able call these methods in isolation.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/template.rb
```diff
@@ -1,2 +1,28 @@
...
+    else
+      unsupported_reason_add(:provisioning, _('not connected to ems'))
+    end
+  end
+
+  # Get a new CloudTool VPC object.
+  # @param connection [ManageIQ::Providers::IbmCloud::CloudTools::Vpc] An existing connection.
+  # @return [ManageIQ::Providers::IbmCloud::CloudTools::Vpc] The passed in connection or a new one.
+  def provider_object(connection = nil)
+    connection || ext_management_system.connect
```
agrare, line 19: Yeah see this isn't what `provider_object` should be doing, if you want the ems connection you can just use `with_provider_connection`.

If you do want the template object you'll have to do the equivalent of:
```
  # Used in with_provider_object to scope SDK to this instance.
  def provider_object(vpc)
    vpc.instances.instance(ems_ref)
  end
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
```diff
@@ -73,7 +78,8 @@ def self.provider_region_options
   end
 
   LABEL_MAPPING_ENTITIES = {
-    "VmIBM" => "ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm"
+    "VmIBM"   => "ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm",
+    "VmImage" => "ManageIQ::Providers::IbmCloud::VPC::CloudManager::Template"
```
agrare, line 82: Should be `ImageIBM`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/options_helper.rb
agrare: Okay this isn't needed for VPC and should be removed then

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: Sort of except that this one always goes to _log and is info. The other goes to logger and is debug. In production they should log the same thing. In development it ensures that all debug logging goes to $ibm_cloud_log. Which means that the developer doesn't have to tail both evm and ibm_cloud to see what is going on.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/template.rb
jaredm-ibm: Another copy/paste from Power

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare:
```
p = ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision.new
>> p.logger("foobar").info("hello")
```

evm.log
```
INFO -- : MIQ(ManageIQ::Providers::IbmCloud::VPC::CloudManager::LoggingMixin::LoggingWrapper#info) ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision.foobar hello
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
```diff
@@ -0,0 +1,85 @@
...
+    # @param msg [String] The message string.
+    # @return [String]
+    def format_message(msg)
+      "#{@method_name} #{msg}"
+    end
+
+    # Determine which logger to use for this call.
+    # @return [Vmdb::Loggers::ProviderSdkLogger, Vmdb::LogProxy]
+    def parent_logger
+      development? ? $ibm_cloud_log : _log
```
agrare, line 71: This throws an exception if `development?` returns false

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: True, evm log does print out what it thinks is the calling method. Whereas LoggingWrapper has the expected method.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
jaredm-ibm: Really? I thought I tested that. No, I meant _log. Like in `_log.info("IBM SERVER PROVISIONING OPTIONS: #{json_clone_options}")`. Could it be that the REPL doesn't have it set, but a running server does? I think, II found I could only test it as a server and not from the rails console.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
Fryguy, line 4: The purpose of the logs and various methods is as follows:

- provider specific log (e.g. `$ibm_cloud_log`) - This purpose of this log is to have a separate logger for the low-level operations that are specific to the provider.  Typically this is just the request/response, and in many providers this logger is given directly to the SDK library for them to log to.  This allows a user to tune the verbosity of the "native" logger independently of the application logger.
- `$log` - This is the primary logger for the application.  Since this PR is about provisioning I would expect all provisioning things to go through this logger (except the low-level native calls).
- `_log` - This is a proxy method for `$log` provided by the [Vmdb::LogProxy](https://github.com/ManageIQ/manageiq/blob/4e7b8e882d6cc6ab224dfea770a3129438d5a82d/lib/vmdb/logging.rb#L4), and its only purpose is to prepend code location information into the `$log` entry in the format `MIQ(Class.method_name) - `.  Having the consistent format allows for readability and for diagnostics when we post-process the logs.  Ideally this would live in $log directly and we wouldn't need this logger, but Ruby doesn't have the capability to know the calling class and method names via the backtrace, so we needed this unfortunate proxy.

Additionally, on appliances, all logs are broadcast to journald and on containers/podified all logs are broadcast to STDOUT in JSON <sup>[[ref]](https://github.com/ManageIQ/manageiq/blob/4e7b8e882d6cc6ab224dfea770a3129438d5a82d/lib/vmdb/loggers.rb#L34-L38)</sup>.

So, given that, part of this class feels unnecessary because it's effectively wrapping a wrapper which already prepends code location information.  The other half seems to deal with the development complexity of having to watch logs in multiple files, but that feels like it can be handled in a simpler way.

I think this class should be removed in favor of using `$ibm_cloud_log` directly for native SDK calls and either `$log` or `_log` for the rest of the logging calls.

For the multiple logs problem I'd recommend doing nothing code-wise and locally just doing `tail -f log/evm.log log/ibm_cloud.log` to blend the logs together.  Alternatively we can probably change the [`$ibm_cloud_log` initialization](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/17faf6d2fd99e35059b4dc7637ade4861927c9fb/lib/manageiq/providers/ibm_cloud/engine.rb#L22-L24) to also broadcast to `$log` much in the same way we do for the container logger, though we'll have to make sure we aren't "double-logging".  On that note, if this is something that we think would benefit all providers, we can also just do that in the core logger initialization directly.  @agrare Thoughts on this suggestion since you are familiar with the other providers?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
```diff
@@ -0,0 +1,61 @@
...
+      nil # NoOp: Let the case fall to the end and use the default status.
+    else # If we get here then the status is unknown. We should log the oddity and then let the provision try again.
+      warn_msg = "Unknown IBM VPC instance status received from the cloud API: '#{status}'"
+      method_log.warn(warn_msg)
+    end
+
+    return false, _('The IBM VPC instance is being provisioned.')
+  rescue => e
```
Fryguy, line 58: This doesn't re-raise the error, so what will happen (particularly with the `raise MiqException::MiqProvisionError` up above which I'd expect to bubble up)

jaredm-ibm: 2 different uses:
1. A job calls `start_clone` to do the provision.  It completes and returns to the job.
2. The job then uses `do_clone_task_check`  to check the progress of the provision.

The job logic requires different returns.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: That may work. I didn't try that. It can be updated in a follow up PR.  At the moment the current implementation is known to work.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
jaredm-ibm: I don't mind removing the switch as long as it is verified that the provider logging will show up in pods. As I said the only reason I did that was because I couldn't find my logging messages in pods. The other part, I'm keeping in. The logging is much simpler and standardized.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: Oh sorry, where you asking about the `log_backtrace`?. The method reraises after logging the exception unless told not to.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
Fryguy: > verified that the provider logging will show up in pods.

```ruby
> CONTAINER=true bin/rails c
...
[1] pry(main)> $ibm_cloud_log.info("hi")
{"@timestamp":"2021-04-30T14:28:00.880548 ","pid":14728,"tid":"9164","level":"info","message":"hi"}
```

> The other part, I'm keeping in. The logging is much simpler and standardized.

Not sure I follow.  From what I can tell it's not following the standardization provided by the LogProxy, because it builds an [alternative string](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198/files#diff-f7024315333d99e0f6101d4f8bdcef8ec7e6c162cf350548f6bf27b26fa2b452R83).

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: Json dump is so that I can copy the values into a text editor and run a json formatter on it. Much easier to verify and deal with.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
Fryguy: Not sure I follow.  If these methods don't bubble up their failures, what will happen?

Typically loggers return `true` (or a truthy value) when they log something, so this method will actually return `true` on a failure, which I think will break the flows.  Looks like we might need some failure case specs?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/volumes.rb
```diff
@@ -0,0 +1,140 @@
...
+    validate_volumes(values[:new_volumes])
+  end
+
+  # Select keys that have one of the volume dialog keys followed by a digit.
+  # @param values [Hash<Symbol, String>] Hash of all the keys and values submitted in the form.
+  # @return [Array<Hash{Symbol => String, Integer}>]
+  def create_new_volumes_array(values)
+    new_volumes = []
+    # Regex to place the field name in the field index and count in the count index.
+    keys_regex = /(?<field>#{volume_dialog_keys.join('|')})_(?<count>\d+)/
```
Fryguy, line 49: I'm concerned there might be values in the keys that are Regex tokens (if not now, perhaps in the future).  I'm thinking we should be defensive here with something like:

```suggestion
    keys_regex = /(?<field>#{volume_dialog_keys..map { |k| Regexp.escape(k) }.join('|')})_(?<count>\d+)/
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/volumes.rb
```diff
@@ -0,0 +1,140 @@
...
+      reg = key.match(keys_regex)
+      next if reg.nil?
+
+      count = reg[:count].to_i
+      new_volumes[count] ||= {}
+
+      field = reg[:field].to_sym
+
+      # Use integer if value starts and ends with digits.
+      new_volumes[count][field] = value.match?('^\d+$') ? value.to_i : value
```
Fryguy, line 63: Super-minor, but prefer a regex here over a string with a regex in it.  Performance-wise, the former is significantly faster (though admittedly that's unlikely an actual problem here, but it's more for code consistency)

Fryguy: Same elsewhere, such as on line 107

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/volumes.rb
```diff
@@ -0,0 +1,140 @@
...
+    error_array = []
+    new_volumes.each_with_index do |item, zero_index|
+      index = zero_index + 1
+      validate_volume_required(index, item, error_array)
+      validate_volume_name(index, item[:volume_name], error_array)
+      validate_volume_size(index, item[:volume_size], error_array)
+      validate_volume_profile(index, item[:volume_profile], error_array)
+    end
+    e_msg = _("New volumes has the following problems: %{error}") % {:error => error_array.join(", ")}
+    raise e_msg unless error_array.length.zero?
```
Fryguy, line 86: No need to build the error message if there is no error to raise

```suggestion
    unless error_array.empty?
      e_msg = _("New volumes has the following problems: %{error}") % {:error => error_array.join(", ")}
      raise e_msg 
    end
```

Additionally, prefer `.empty?` over `.length.zero?`  Alternatively, prefer avoiding a double negative (`unless ary.empty?`) with a positive (`if array.any?`)

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/volumes.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  # Validate that volume_name and volume_size are populated. Add a string explaining each violation to error_array.
+  # @param item [Hash<Symbol, String>] A new_volume hash.
+  # @param error_array [Array<String>] An array of validation error strings.
+  # @return [void]
+  def validate_volume_required(index, item, error_array)
+    logger(__method__).debug("Item is #{item}")
+    %i[volume_name volume_size].each do |req|
+      e_str = _("Volume %{index} %{field} is required.") % {:index => index, :field => req.to_s}
+      error_array.append(e_str) unless item.key?(req)
```
Fryguy, line 97: Similar comment here about avoiding building the error string.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
Fryguy: Yeah I can see that.  Does that mean we'd do this in production too, since it's only a development convenience?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
```diff
@@ -0,0 +1,85 @@
...
+    # @param exception [Exception] The exception object.
+    # @param context_msg [String] A message to be printed as an error that provides context for the exception.
+    # @param re_raise [Boolean] Whether to raise the original error again.
+    # @param send_return [] The object to send as a return value.
+    # @return [void]
+    def log_backtrace(exception, context_msg: '', re_raise: true, send_return: nil)
+      parent_logger.error(format_message(context_msg)) unless context_msg.length.zero?
+
+      parent_logger.log_backtrace(exception)
+      raise if re_raise
```
Fryguy, line 54: This is unintuitive.  The log_backtrace method from the base VMDBLogger already has the semantics that it just logs the backtrace (as per the name) and nothing else.  Adding in this raise changes those semantics and makes it complicated for the reader (hence my eventual [comment below](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198/files#r624065699)).

Typically, log_backtrace is only used in places where we don't want to re-raise but need contextual information.  Otherwise letter the errors buble up is often sufficient.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
Fryguy: I commented [over there](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198/files#r624117724)  (didn't realize you replied here when I noticed that, and didn't realize it was basically the same conversation thread).

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/volumes.rb
jaredm-ibm: I read the same thing when I was figuring this out. I prefer match because it is googable for people not familiar with the `=~` which aids in maintenance. I'm also not comfortable with it and don't want to do a bunch of testing for something so minor.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb
agrare: > Could it be that the REPL doesn't have it set, but a running server does? I think, if I remember correctly, I found I could only test it as a server and not from the rails console.

Yeah if you want to call `_log` you have to `include Vmdb::Logging`  Often times this is done in one of the base classes for you but since you're defining your own `LoggingWrapper` class it is up to you to include this.

Even if this worked with the full server running that indicates there is a dependency load ordering issue which will come back to bite us.

> Is there a log method? Where does that go?

```
=> #<Method: main.log() /home/grare/adam/.gem/gems/yard-0.9.26/lib/yard/globals.rb:20>
>> log
=> #<YARD::Logger:0x0000564a04a7ff90 @level=2, @progname=nil, @default_formatter=#<Logger::Formatter:0x0000564a04a7fef0 @datetime_format=nil>, @formatter=#<Method: YARD::Logger#format_log(sev, _time, _prog, msg) /home/grare/adam/.gem/gems/yard-0.9.26/lib/yard/logging.rb:201>, @logdev=#<IO:<STDOUT>>, @show_backtraces=true, @show_progress=false, @warned=false, @progress_indicator=0, @mutex=#<Thread::Mutex:0x0000564a04a7fd88>, @progress_msg=nil, @progress_last_update=2021-05-04 09:00:25.637672949 -0400>
```

So that's not what we want here


Running in production:
```
RAILS_ENV=production rails c
>> vpc_provision = ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision.new
>> vpc_provision.logger("foo").info("hello, world")
        2: from /home/grare/adam/src/manageiq/manageiq-providers-ibm_cloud/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb:20:in `info'
        1: from /home/grare/adam/src/manageiq/manageiq-providers-ibm_cloud/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/logging_mixin.rb:71:in `parent_logger'
NameError (undefined local variable or method `_log' for #<ManageIQ::Providers::IbmCloud::VPC::CloudManager::LoggingMixin::LoggingWrapper:0x00005624ae5dc6d0>)
Did you mean?  log
```

And if we include `Vmdb::Logger`:
```
   class LoggingWrapper
+    include Vmdb::Logging
+
     # Define a short-lived object for this method.
```

```
RAILS_ENV=production rails c
>> vpc_provision = ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision.new
>> vpc_provision.logger("foo").info("hello, world")
=> true
```

agrare:
Okay so taking a step back, as I understand it and correct me if I'm wrong the `LoggingWrapper` was added because logs sent to `$ibm_cloud_log` weren't being sent to STDOUT.

I tested this out [[here]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198#discussion_r623974519) in development but just to be 100% sure I connected to a full-up production podified environment and sure enough:
```
oc rsh orchestrator-7f8bc45687-4gqlt /bin/bash
[root@orchestrator-7f8bc45687-4gqlt /]# vmdb
[root@orchestrator-7f8bc45687-4gqlt vmdb]# source ./container_env 
[root@orchestrator-7f8bc45687-4gqlt vmdb]# rails c
Loading production environment (Rails 6.0.3.6)
irb(main):001:0> $ibm_cloud_log.info("hello, world")
{"@timestamp":"2021-05-04T12:33:08.756904 ","hostname":"orchestrator-7f8bc45687-4gqlt","pid":59082,"tid":"2add16f71958","level":"info","message":"hello, world"}
```

So if this was the reason for the "log-switcher" I believe we can drop that.  This should reduce some complexity and also resolves the issue with `_log` blowing up in due to `Vmdb::Logging` not being included in the `LoggingWrapper` class and the double-caller_locations in production mode.

Using the `LoggingMixin` versus `_log` is up to you, I think `_log` does 90% of what the `LoggingMixin` does but that's your call.

jaredm-ibm:
To get this in, I'm going to remove all of my enhancements. I'm removing:
* My logging enhancements and all logging that uses them.
* My additional verification of the add volumes.

This means that:
* On exception the provision form will throw a useless error and break the cancel button. The user then has to figure out a way to get out of the form on their own.
* On exception the actual error may be swallowed by subsequent errors.
* If the customer isn't very careful in what they put into the add volume text fields they will be able to submit a provision that will fail when it is submitted to VPC. It is only then that they will know a mistake was made but they will not know what it is.

agrare:
@jaredm-ibm I think we might be missing each others points, maybe a video chat would help clear things up?

jaredm-ibm:
I'm removing everything because at this point I don't know how to get it in. It seems that all the comments are: 
* Either in my enhancements 
* In things that I copied and pasted from Power, which I expected should be correct since they are already merged.

So the solution to getting back to zero is just to remove all the code with comments and be done with it. The custom code is put there to work around issues I found while developing this feature. Telling me it is redundant, without asking why I did it is insulting. So since you didn't ask, I'm just telling you at the end where your code has problems and will let you & Jason figure out the ManageIQ way of doing it.

agrare:
@jaredm-ibm I'm trying to move forward not backward here and I hope you'll join me.

> I'm removing everything because at this point I don't know how to get it in

This is what I'm trying to say with https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/198#issuecomment-831931179
We can't merge it as is because it will throw an exception on production, so if you drop the LoggingWrapper I'm good with it.

The methods that are coped from power that are power specific (aka the cloud_instance_id) can be removed because they aren't needed for vpc.

The provider_object methods on power are incorrect, were missed during that review (no one ever said we're perfect) and we should fix those rather than continuing to add incorrect code.  We're not saying it is your fault that those are copied, quite the opposite.

Other than those issues this LGTM, up to you if you want to tweak what you have or throw it all away.

jaredm-ibm:
I agree with you on the log switch. I never liked it, but thought it useless to have this error catching if it didn't show up in production. I can't see how it would blow up in production as _log is used liberally within the code, but whatever as mentioned before, if the standard log shows up in production, then you have no argument from me to remove it. I also didn't like using the Config so I get to remove that as well, and will be much happier.

Yeah, I've already removed the unused method and did a quick test, just haven't checked it in . I'm still not 100% comfortable with the change since I haven't given the time to ensure nothing breaks. In my opinion the incorrect methods should be issues since the PR has already been tested in its current form and this is non-trivial change. There is nothing stopping prioritizing the issue to be the next one. Either way it allows for proper testing of the change and not some mad-dash to get the PR in. When I change code based on PR feedback, I always make stupid mistakes, or think of scenarios after merge where the change could break. We tabled our discussion for the CloudTool string validation. The next PR to cloud tools included the discussed fixes.

jaredm-ibm:
Oh the other issue you should know about, which I haven't solved at all except to do my best to validate everything. Is that the IBM VPC SDK removes the field that failed from the error exception message. I don't know why they did that, maybe to preemptively guard against data leakage. In any case, I had to use my original VPC SDK to verify the payload JSON and get the actual error messages. So you might want to raise an issue upstream to fix that before the payloads change.

agrare:
> In my opinion the incorrect methods should be issues since the PR has already been tested in its current form and this is non-trivial change

No problem at all on my end if you want to tackle some of this in a follow-up

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'ipaddr'
```
chessbyte, line 3: Suggest moving this require closer to where it is used around line 64.

jaredm-ibm:
In #201 I removed the log switch and revalidated the error flow. The exercise found a couple of edge cases and standardized the error handling.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
jaredm-ibm: See #203

jaredm-ibm:
I think all the agreed changes are in. I created #205 for future work on correcting `with_provider_object`.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare: Now that you aren't logging to `_log` this isn't an issue anymore

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare: I agree a logging method re-raising an exception is highly unexpected (at least within our code base), think we should reevaluate in a follow-up.  Personally I'd take an extra line of code in the exception handler as long as it is obvious what is being done if you don't already know how this particular `log_backtrace` method behaves.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: The intended target for the default message is still _log. For debugging the target is $ibm_cloud_log to provide an order of events in one place. Essentially, during normal operation _log logs the operation which is good because all tthe other messages are there. During debugging $ibm_cloud_log has a record which is good because all the exceptions are there.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
jaredm-ibm: Huh? This method is to poll the provision. There are 3 states 
1. false = keep polling
2. true = provision complete
3. Unhandled exception = error but I don't know what to do with it. Fail the provision and log the error.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare: @jaredm-ibm I'm not sure what you mean, I'm stating that now that you are no longer passing a formatted message that includes the calling class+method to `_log` that this no longer will double-prefix messages which is what my comment is about.

If I'm wrong about that and LoggingWrapper actually still is logging to `_log` then please correct this in a follow-up

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/cloning.rb
agrare: We're not about this method specifically, but rather the overall behavior of `log_backtrace` in the `LoggingMixin`

jaredm-ibm: The behaviour is the same but more exact, which is what you want when you are debugging. In many cases during my testing the exception was printed to the screen or otherwise swallowed up because of subsequent error handling. When debugging not having the exact error and where it came from wastes time. So if you are in favour of wasting people's time I can't help you. I can ensure that my code is easy to debug.

Fryguy:
Backported to `morphy` in commit d0114befdb488e2d05be3e66d1d0381dae2013dd.

```text
commit d0114befdb488e2d05be3e66d1d0381dae2013dd
Date:   Tue May 11 10:19:57 2021 -0400

    Merge pull request #198 from jaredm-ibm/vpc_provision
    
    Vpc provision
    
    (cherry picked from commit 5f311520f24f01036fe0734aa1d2773614dc5180)
```

Issue #199, opened 2021-05-03
Attach floating IPs to their VM instances.


jaredm-ibm:
This is issue is to close #99 & #95

- [x] Figure out how FloatingIPs work in VPC Cloud.
- [x] Get the REST result for a VM with a floating IP.
- [x] Determine how to persist floating IP details.
- [x] Add persister configuration.
- [x] Add test for floating IP.

jaredm-ibm:
Floating IPs were actually being persisted already. The relationships to vms and network_ports didn't exist.

Issue #95, opened 2020-10-27
Public/floating IP addresses are not associated with their respective instances


chadh1313:
Would expect this to appear in the instances list and instance details page

![image](https://user-images.githubusercontent.com/3876941/97311988-5ef23200-183b-11eb-92b1-8847e21eed71.png)

![image](https://user-images.githubusercontent.com/3876941/97312045-6dd8e480-183b-11eb-8571-98373dcdd994.png)

chessbyte:
@jaredm-ibm if you can confirm that this works with #200, then please close this issue.

jaredm-ibm:
Yes, I need to capture screen shots then I will close.

jaredm-ibm:
# Instances list shows both IPs

<img width="522" alt="Screen Shot 2021-05-07 at 10 22 04 AM" src="https://user-images.githubusercontent.com/60901094/117464231-74e55b80-af1e-11eb-9d3d-b245a34464a6.png">

# details page shows both IPs

<img width="378" alt="Screen Shot 2021-05-07 at 10 25 21 AM" src="https://user-images.githubusercontent.com/60901094/117464395-9ba39200-af1e-11eb-983c-fd3ebeff9e49.png">

PR #200, opened 2021-05-04
Floating ip to instance

### LINKED ISSUE #99: IBM Cloud VPC - missing details, relationships with Floating IP addresses
1) In IBM Cloud VPC Network Manager, select the Floating IP list is missing details: `Fixed Address`, `Instance name`, `Network Port name`

image

As all these public IP addresses are associated with a private IP on a port/instance, these relationships should be possible.

2) The Network Manager provider overview page shows a count of Floating IPs in the relationships

image

However, when select a particular cloud network, that hosts the instances using these addresses, the Floating IP count is zero (or some lesser number).  Would think that since the network manager knows the ports, private IP addresses, etc. the associated Public/Floating IPs would be part of the individual network overview page.

image

Note: this behavior is also observed with the AWS provider, so either this is working as designed, or is a defect with that provider as well.

jaredm-ibm:
# Issue
Floating ips with attached instances are not being related to their VMs. This PR is being tracked by #199 and should close #99 & #95

# Priority
* Chad would like this backported as it breaks the CP4MCM UI and prevents some agent from being deployed.

# Implementation
* Add method to find the instance id if the `:target` key is available.
* Add vm & network port relationships to persister
* Populate fixed_ip_address field in persister.

# Not covered
* No relationship is created for floating ips attached to network routers. VPC Plugin doesn't have the infrastructure for it yet.
* In manual testing adding them as-is doesn't create any issues with the relationships.

# Specs
* Add a `assert_specific_floating_ip` method which tests all populated fields for a live attachment.
* When rerecording there will need to be a floating IP attached to a VM to have this pass. It's not optimal.
* We need a better infrastructure for this test. Using VCRs isn't good for scenario testing.
* Moved `assert_specific_resource_group` to be below `assert_specific_vm` so that all the methods that will need to be changed on rerecord in a different environment are at the top.

# Common files
* In `spec_helper` Add a response header filter to `vpc_sanitizer`. This will eliminate a number of VCR changes that are not used by the code.
* In `spec_helper` Add tags api to vpc_sanitzer to filter out saving the bearer token and response headers.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -253,14 +253,35 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
     end
   end
 
+  # Persist all floating ips in VPC Cloud.
+  # @return [void]
   def floating_ips
     collector.floating_ips.each do |ip|
+      instance_id = floating_ip_instance_id(ip)
       persister.floating_ips.build(
-        :ems_ref => ip[:id],
-        :address => ip[:address],
-        :status  => ip[:status]
+        :vm               => persister.vms.lazy_find(instance_id),
+        :network_port     => persister.network_ports.lazy_find(ip&.dig(:target, :id)),
+        :ems_ref          => ip[:id],
+        :address          => ip[:address],
+        :status           => ip[:status],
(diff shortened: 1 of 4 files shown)


agrare (approved):

Fryguy:
Backported via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/216

agrare:
Included in morphy via #216

PR #202, opened 2021-05-05
Fix the dynamic secret cassette placeholder code

agrare:
This was supposed to be `.each_key` not `.keys` and was preventing proper secret replacement when re-recording cassettes

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -57,19 +57,16 @@ VCR.configure do |config|
   config.ignore_hosts('codeclimate.com') if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
 
-  # Used to replace the API Key with the placeholder in the saved VCR.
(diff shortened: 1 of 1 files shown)


jaywcarman (approved):
LGTM

jaywcarman:
@iv1111 we'll need to update #159 to include a similar line for `secrets.ibm_cloud_object_storage`

spec/spec_helper.rb
```diff
@@ -57,19 +57,16 @@ def vpc_sanitizer(interaction)
   config.ignore_hosts('codeclimate.com') if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
 
-  # Used to replace the API Key with the placeholder in the saved VCR.
-  config.define_cassette_placeholder('IBM_CLOUD_VPC_API_KEY') { Rails.application.secrets.ibm_cloud_vpc[:api_key] }
```
agrare, line 61: This was added after the fact to workaround [[this bug]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/202/files#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94L69)

PR #208, opened 2021-05-12
Cloudtool verify method existence

### LINKED ISSUE #206: IBM VPC Add method check to cloud tool.
The cloud tool methods request and collection take a method argument. This should be checked to ensure that the method exists before using it. If it doesn't a suitable error message is provided.

jaredm-ibm:
# Issue
Closes #206 

* Since these methods use callbacks it can be easy to pass in methods that are not a client method. This will do a check and throw an error before attempting to make the call. 
* All methods that are potentially paginated begin 'list'. This will ensure that non-paginated methods are not passed in accidentally.

# Priority
* No dependencies

# Implementation
* Check if client responds to the method before calling it.
* In collection check that the passed in method begins with list.

# Not covered
* n/a

# Specs
* Add tests to ensure that the correct Exception is thrown in both cases.

# Common files
* Rework spec_helper so that response headers for iam requests are also filtered.

### DIFF
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/sdk/branch.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/sdk/branch.rb
@@ -71,6 +71,8 @@ module ManageIQ
             end
 
             def send_request(call_back, **kwargs)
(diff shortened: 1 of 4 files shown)


agrare (approved):

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/276

Issue #180, opened 2021-04-01
Enabling cross cloud-provider image import.


iv1111:
Location: UI, API
Form: Add

current type: 
API state:
Need info

PM Input:
Comments: 

This currently open [RFE 7675](https://github.com/ManageIQ/manageiq-ui-classic/issues/7675) is **blocking** us from proceeding.

Additionally, we need to decide on architecture of the API handler in the manageiq-api, namely 
how we implement the "supports?" method for validating whether a given workflow is supported.
ManageIQ/manageiq-api#1021

Ems common

imported status: NOT ASSIGNED

### ManageIQ base, API, and UI
- [x]  Documentation describing steps necessary in order to try out the image import workflow (https://github.com/ManageIQ/manageiq-documentation/pull/1618)
- [x] new POST API method allowing import of images from one provider into another (ManageIQ/manageiq-api#1021)
- [x] adding support for importing images between providers (ManageIQ/manageiq#21106)
- [x] adding 'import_image' API-feature to miq_product_features.yml (ManageIQ/manageiq#21112)
- [x] Allow the addition of standalone object storage managers (ManageIQ/manageiq-ui-classic#7675)
    - [x] Combine Block/Object Mixins into StorageManager (ManageIQ/manageiq#21213)
    - [x] Update API product features (ManageIQ/manageiq-api#1041)
    - [x] Schema migration for product features (ManageIQ/manageiq-schema#581)
    - [x] Combine block and object storage and refactor storage submenu (ManageIQ/manageiq-ui-classic#7746)
 

### PowerVC Provider
- [x] New IBM PowerVC provider ([ManageIQ/manageiq-providers-ibm_power_vc](https://github.com/ManageIQ/manageiq-providers-ibm_power_vc))
- [x] PowerVC decorator (ManageIQ/manageiq-decorators#49) (needs a better icon)
- [x] Restructured registration form (ManageIQ/manageiq-providers-ibm_power_vc#15)

### IBM Cloud Provider
- [x] Image Import Functionality for PowerVS <-- PowerVC (ManageIQ/manageiq-providers-ibm_cloud#215)
- [x]  Add new IBM Cloud-Object-Storage provider (ManageIQ/manageiq-providers-ibm_cloud#159)

----

We are working on an interesting use-case of allowing users to import images from one cloud provider into another.
An example of how this might look like can be seen below:

[...]

PR #215, opened 2021-05-25
Image Import Functionality for PowerVS <-- PowerVC scenario.

iv1111:
- UI Image-Import button on the Provider View
- UI Image-Import form doing data lookup & sending import request
- Import Ansible playbook implementation (including a corresp. role)
- Import Workflow implementation, polling and data cleanup included

For bigger picture and dependencies see this issue: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/180


iv1111:
@agrare Here is our plugin related PowerVS <--- PowerVC import PR. It's by no means clean yet (error handling, code style, some "must-have" functionality missing) and will be polished first.

jaywcarman:
I rebased this onto the latest #159 to pick up @agrare's commits. I still need to take care of linter warnings (testing first, since some of my previous linter "fixes" broke functionality).

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -92,6 +89,10 @@ def self.params_for_create
     }.freeze
   end
 
+  def queue_name_for_ems_refresh
+    "object"
+  end
```
agrare, line 94: Curious what you're trying to solve with this as this is almost certainly incorrect

iv1111: We had a problem that _queue_name_for_ems_refresh_ was non-existent for the object storages, so we had to define it explicitly in our class. I will check now if it has been solved in the parent manager.

agrare: `app/models/manageiq/providers/storage_manager.rb:    delegate :queue_name_for_ems_refresh, :to => :parent_manager` should cover this for you

iv1111: Ok, sounds good, then I'll remove the method.

agrare: If you didn't have a parent manager you would want `def queue_name_for_ems_refresh; queue_name: end`

iv1111: COS is a stand-alone manager, so it might exist without a parent unless the abstract parent manager that we wanted to achieve is already in place.

agrare: :+1:  okay so you want this to be `queue_name` (which is what is in the base ExtManagementSystem) instead of `"object"`

agrare:
cc @kavyanekkalapu can you assist reviewing the js changes here?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,111 @@
...
+      :pre_execute       => {'pre_execute'        => 'execute'},
+      :execute           => {'execute'            => 'running'},
+      :poll_runner       => {'running'            => 'running'},
+      :post_execute      => {'running'            => 'post_execute_poll'},
+      :post_execute_poll => {'post_execute_poll'  => 'post_execute_poll'},
+      :finish            => {'*'                  => 'finished'},
+      :abort_job         => {'*'                  => 'aborting'},
+      :cancel            => {'*'                  => 'canceling'},
+      :error             => {'*'                  => '*'}
+    }
```
agrare, line 17: Since most of this state transition is the same as the one from the base class this method could be simplified to
```
def load_transitions
  super.merge(:post_execute => {'running' => 'post_execute_poll'}, :post_execute_poll => {'post_execute_poll' => 'post_execute_poll'})
end
```

So that you don't overwrite the whole set of state transitions.  This means less likely to break if we change something in core

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def post_execute
```
agrare, line 20: If you're going to overwrite the post_execute method you should still call `cleanup_git_repository` which is done by the core method

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,111 @@
...
+    ems.with_provider_connection(:service => 'PCloudImagesApi') do |api|
+      response = api.pcloud_cloudinstances_images_post(ems.uid_ems, body, {})
+      context[:task_id] = response.taskref.task_id
+
+      started_on = Time.now.utc
+      update!(:context => context, :started_on => started_on)
+      miq_task.update!(:started_on => started_on)
+    end
+
+    route_signal(:post_execute_poll, 'importing image into PVS', 'running')
```
agrare, line 41: You only need `route_signal` if you depend on filesystem state or otherwise require being on the same pod in podified.  It looks like `post_execute_poll` would be just fine with `queue_signal`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,111 @@
...
+        status = 'error'
+      else
+        signal = :error
+        msg = 'PVS API has responded with a non-standard status for a cloud-task: ' + response.status
+        status = 'error'
+        _log.warn(msg)
+      end
+    end
+
+    route_signal(signal, msg, status)
```
agrare, line 78: Same here

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,111 @@
...
+      :source        => 'url',
+      :imageName     => options[:miq_img][:name],
+      :osType        => options[:miq_img][:os],
+      :diskType      => options[:diskType],
+      :imageFilename => "#{options[:session_id]}.ova",
+      **options[:cos_pvs_creds]
+    }
+
+    ems.with_provider_connection(:service => 'PCloudImagesApi') do |api|
+      response = api.pcloud_cloudinstances_images_post(ems.uid_ems, body, {})
```
agrare, line 33: What is the playbook doing?  Could this post and the subsequent poll be handled there?  That way the core poll_runner state would just work

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: Got it, I will change now.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: Technically it can be and I had the implementation for it somewhere at the early stages but our team voted against such design. One reason being that I'd be duplicating REST calls in my Ansible / Python code, which are already available in our ruby gem and thus should be reused. Another reason is that we'd have to unnecessarily provide PowerVS credentials to the Ansible scripts running on a powervc node (which can be thus avoided).

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
chessbyte: easiest way is to do your own thing and call `super`

iv1111: @chessbyte Hmm, yeah, especially if the base class implementation ever get's changed in the future.

agrare: The issue with that is the `post_execute` method in core calls `queue_signal(:finish)` at the end where here he needs to `queue_signal(:post_execute_poll)`

iv1111: good catch

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: done

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: done

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -47,6 +49,87 @@ def raw_delete_image
...
+    return endp.hostname, auth.userid, auth.password
+  end
+
+  def self.image_ems_ref(bucket_id)
+    MiqTemplate.find(bucket_id).uid_ems
+  end
+
+  def self.bucket_name(bucket_id)
+    CloudObjectStoreContainer.find(bucket_id).name
+  end
```
agrare, line 87: Most of these methods don't make any sense as public class methods on a `CloudManager::Template`.  They should be private helper methods (maybe grouped in their own module).

iv1111: Thanks! I'll take care of that.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -47,6 +49,87 @@ def raw_delete_image
...
+
+  def self.raw_import_image(ext_management_system, options = {})
+    session_id = DateTime.now.strftime('%Q')
+
+    location, user, password = node_creds(options['src_provider_id'])
+    hosts = ["[powervc]\n#{location}\n[powervc:vars]\nansible_connection=ssh\nansible_user=#{user}\nansible_ssh_pass=#{password}"]
+
+    cos_credentials = cos_creds(options['obj_storage_id'])
```
agrare, line 95: Think this would be more readable if you did something like `instance_id, api_key = cos_creds()` instead of `cos_credentials[0], cos_credentials[1]`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -47,6 +49,87 @@ def raw_delete_image
...
+    bucket = bucket_name(options['bucket_id'])
+
+    diskType = CloudVolumeType.find(options['disk_type_id']).name
+
+    cos_ans_creds = {:resource_instance_id => cos_credentials[0], :apikey => cos_credentials[1], :bucket_name => bucket, :url_endpoint => cos_credentials[3]}
+    cos_pvs_creds = {:region => cos_credentials[2], :bucketName => bucket, :accessKey => cos_credentials[4], :secretKey => cos_credentials[5]}
+
+    encr_cos_creds, encr_cos_key, encr_cos_iv = encrypt_with_aes(cos_ans_creds)
```
agrare, line 103: Ansible runner has handling of credential files already, not sure that encrypting the string then passing the decryption key in the same extra_vars is dramatically more secure :shrug:

iv1111: The ansible playbook turns the creds into a file on the powervc node while the "key" and "iv" are turned into environmental variables for the python process we use on the powervc side. This won't let a simple user decrypt the credentials since the user doesn't have access to the environment of the python process. You can see it here: content/ansible_runner/roles/image-copy-pvc-to-pvs/defaults/main.yml. Not sure I can achieve the sane functionality using the built-in "credentials" option.

iv1111: See also the following three lines: https://github.com/iv1111/manageiq-providers-ibm_cloud/blob/739f5f36f2aad71364dbe8229442399fba3f2434/content/ansible_runner/roles/image-copy-pvc-to-pvs/files/upload_ova_to_bucket.py#L16

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
iv1111: Sure, I will change.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: I've changed according to Adam's suggestion

kavyanekkalapu:
> cc @kavyanekkalapu can you assist reviewing the js changes here?

@agrare i just got the email :) , i will check today , even github.com/notification is showing empty page 👎

app/javascript/components/import-image-form.jsx
```diff
@@ -0,0 +1,100 @@
+import React, { useMemo, useState } from "react";
+import PropTypes from "prop-types";
+import { connect } from "react-redux";
+import MiqFormRenderer from '@@ddf';
+import createSchema from './import-image-form.schema.js';
+
+const BLANK = {value: '-1', label: ''};
+
+const fetchProviders = (kind) => {
```
kavyanekkalapu, line 9: Is it possible to move all these functions to components/import-image-form/helpers.js file?

app/javascript/components/import-image-form.jsx
```diff
@@ -0,0 +1,100 @@
...
+  const buckets   = useMemo(() => fetchBuckets(storage), [storage]);
+
+  const initialize = (formOptions) => {
+    dispatch({ type: "FormButtons.init",        payload: { newRecord: true, pristine: true } });
+    dispatch({ type: "FormButtons.customLabel", payload: __('Import') });
+    dispatch({ type: 'FormButtons.callbacks',   payload: { addClicked: () => formOptions.submit() }});
+  };
+
+  const submitValues = () => {
+      if ([provider, image, storage, bucket].some(item => parseInt(item) === -1))
```
kavyanekkalapu, line 83: I really couldn't understand why we need to store values in these variables, these will be available in  values.src_provider_id,values.src_image_id....., You can refer https://github.com/ManageIQ/manageiq-ui-classic/tree/master/app/javascript/components/subnet-form or any other form for reference.

app/javascript/components/import-image-form.jsx
```diff
@@ -0,0 +1,100 @@
...
+      {
+          window.add_flash("Request ignored due to incomplete input", 'warning');
```
kavyanekkalapu, line 85: You can add      ` isRequired: true,validate: [{ type: validatorTypes.REQUIRED }],` to schema , ifuser did not enter these fields, this will automatically disable submit button .  I don't we need to a flash message here.

app/javascript/components/import-image-form.schema.js
```diff
@@ -0,0 +1,76 @@
...
+  fields: [
+    {
+      component: componentTypes.SELECT,
+      name:"src_provider_id",
+      id:"src_provider_id",
+      initializeOnMount: true,
+      isRequired: true,
+      label: "Choose source provider",
+      loadOptions: () => providers,
+      onChange: (value) => {setProvider(value); setImage('-1')},
```
kavyanekkalapu, line 15: I really couldn't understand why we need to set values for provider and below fields, these will be available in values.src_provider_id,values.src_image_id....., You can refer https://github.com/ManageIQ/manageiq-ui-classic/tree/master/app/javascript/components/subnet-form or any other form for reference.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
iv1111: @agrare I've grouped  the helper functions into own module. However, I'm note sure how to include them as private static methods into my class without doing things like "class << self". Do you have a suggestion?

iv1111: It's just they're called from "self.raw_import_image" so I can't have them as instance methods.

agrare: The typical way of doing this is to use `included` from `ActiveSupport::Concern`

Something like this gets you a private class method included from a module
```
module TestMixin
  extend ActiveSupport::Concern

  included do
    private_class_method def self.foo
      :tada
    end 
  end 
end
```

iv1111: I did as you suggested here: app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import.rb but I'm getting `[["status", "Error"], ["message", "undefined method `node_creds' ` for some reason. Probably something is still missing.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
iv1111: done

app/javascript/components/import-image-form.jsx
iv1111: These contain API calls in them, so as discussed during the telco I'd leave them as ist only use string constants instead of directly encoding URLs

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
iv1111: Done. I've checked by trying to call the private class methods using Ruby console and it fails as expected:
`NoMethodError (private method `node_creds' called for #<Class:0x000000000884dc00>)`

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -206,4 +204,12 @@ def self.raw_connect(region, endpoint, access_key, secret_key)
 
     Aws.const_get(:S3)::Resource.new(options).client
   end
+
+  def remove_object(bucket_name, object_name)
+    connect.delete_object({:bucket => bucket_name, :key => object_name})
```
agrare, line 209: If the only argument is a hash you don't need the `{}` around it: `connect.delete_object(:bucket => bucket_name, :key => object_name)`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -0,0 +1,99 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ImageImportWorkflow < ManageIQ::Providers::AnsiblePlaybookWorkflow
+  def load_transitions
+    super.merge(:post_execute => {'running' => 'post_execute_poll'}, :post_execute_poll => {'post_execute_poll' => 'post_execute_poll'})
```
agrare, line 3: Maybe it is just because I'm used to seeing these on multiple lines, but I find this harder to read

Can you split up like:
```ruby
super.merge(
  :post_execute      => {...},
  :post_execute_poll => {...}
)
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -47,7 +51,88 @@ def raw_delete_image
     _log.error("image=[#{name}], error: #{e}")
   end
 
+  def self.raw_import_image(ext_management_system, options = {})
+    session_id = DateTime.now.strftime('%Q')
```
agrare, line 55: Does this session_id have to match some other system?  If you just want a unique id something like `SecureRandom.uuid` would be better

agrare:
This looks almost ready

app/javascript/components/import-image-form.jsx
iv1111: I have changed the code and it submits "values" directly + two other json elements that do not come from the UI form / user choice.

app/javascript/components/import-image-form.jsx
iv1111: I have removed this code and use my own custom validator as the standard "required" one is not sufficient in my case. Now the "Submit" button is enabled only if the required fields pass the validation test and this code above is removed.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
iv1111: done

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: done

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
iv1111: done

app/javascript/components/import-image-form.schema.js
iv1111: I've refactored this part doing it in a similar fashion as in the manageiq-ui-classic module. I still had to set the state in the onChange handler's of two fields though. The reason being is that the next fields depend on the choice done in previous ones. They're dynamic and hence the chosen values are handled not only in the "submit" function where the "values" field is available.

iv1111:
> This looks almost ready

Sounds good. I'd still need @kavyanekkalapu OK on the above open conversation. If all is good then I'll run rubocop autocorrect and squash all commits into one.

app/javascript/components/import-image-form/import-image-form.schema.js
```diff
@@ -0,0 +1,82 @@
...
+
+const createSchema = (providers, images, storages, buckets, diskTypes, state, setState) => ({
+  fields: [
+    {
+      component: componentTypes.SELECT,
+      initializeOnMount: true,
+      clearOnUnmount: true,
+      name: 'src_provider_id',
+      id: 'src_provider_id',
+      label: 'Choose source provider',
```
kavyanekkalapu, line 13: @iv1111 labels need to be translated, Just do __('Choose source provider') or you can refer any other form in ui-classic for examples.

app/javascript/components/import-image-form.schema.js
iv1111: done as well, as discussed during the telco

app/javascript/components/import-image-form/import-image-form.schema.js
iv1111: done

kavyanekkalapu:
Looks good from UI side 👍

iv1111:
@agrare Kavya gave her Ok regarding the UI part (above). I've squashed my latest commits and tested one more time. 
I'm removing the WIP. Could you merge this one? Thx.

agrare (approved):

jaywcarman:
Nice work @iv1111 @agrare and @kavyanekkalapu! It's awesome to see this merged! 🎉

agrare:
@iv1111 realize we still need to do the throttling but could you record a short demo of this feature for our ManageIQ youtube channel and the next MIQ sprint review?

iv1111:
> @iv1111 realize we still need to do the throttling but could you record a short demo of this feature for our ManageIQ youtube channel and the next MIQ sprint review?

Sure, I will send you the vid on Tuesday.

iv1111:
Here are some screenshots. Video shared with @agrare.

----

Image Import Button

Image Import Form

PR #216, opened 2021-05-27
[LASKER] Floating ip to instance

agrare:
Manual backport of https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/200


agrare:
I took the VCR from the current `upstream/master` branch, dropped the two collections that we aren't doing (resource groups and volume types) so that we don't get unused http interaction failures, and then just had to update a couple of counts in the existing spec tests.

PR #218, opened 2021-06-10
VPC Targeted Refresh for VMs

nasark:
## Task
- Implemented targeted refresh for VMs
- added scaffolding for targeted refresh of other resources to be implemented in the future

## Results
In this scenario, `nasar-instance` is the name of my VM instance that is being refreshed

```
[----] I, [2021-06-10T15:42:42.221603 #42185:b180]  INFO -- : MIQ(ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher#refresh) EMS: [nasar-vpc], id: [2] Refreshing targets for EMS...
[----] I, [2021-06-10T15:42:42.221678 #42185:b180]  INFO -- : MIQ(ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher#refresh) EMS: [nasar-vpc], id: [2]   InventoryRefresh::TargetCollection [Collection of 1 targets] id [Collection of targets with id: ["nasar-instance"]]
[----] I, [2021-06-10T15:42:45.368297 #42185:b180]  INFO -- : MIQ(ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher#refresh_targets_for_ems) EMS: [nasar-vpc], id: [2] Refreshing target InventoryRefresh::TargetCollection [Collection of 2 targets] id [Collection of targets with id: ["nasar-instance", {:ems_ref=>"02r7_cc596521-f706-4888-b1be-a3aafd2034ea"}]]...
[----] I, [2021-06-10T15:42:57.486071 #42185:b180]  INFO -- : EMS: [nasar-vpc], id: [2] Saving EMS Inventory...
[----] I, [2021-06-10T15:42:57.672808 #42185:b180]  INFO -- : EMS: [nasar-vpc], id: [2] Saving EMS Inventory...Complete
[----] I, [2021-06-10T15:42:57.672891 #42185:b180]  INFO -- : MIQ(ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher#refresh_targets_for_ems) EMS: [nasar-vpc], id: [2] Refreshing target InventoryRefresh::TargetCollection [Collection of 2 targets] id [Collection of targets with id: ["nasar-instance", {:ems_ref=>"02r7_cc596521-f706-4888-b1be-a3aafd2034ea"}]]...Complete
[...]

PR #219, opened 2021-06-11
Make the target_collection class name more generic

agrare:
Rather than hard-coding VPC::TargetCollection use the target.manager.class to build the class name

Follow-up to: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/218#discussion_r649511782

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21273

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory.rb
@@ -3,10 +3,10 @@ class ManageIQ::Providers::IbmCloud::Inventory < ManageIQ::Providers::Inventory
   require_nested :Parser
   require_nested :Persister
 
-  def self.parsed_manager_name(_ems, target)
+  def self.parsed_manager_name(ems, target)
     case target
     when InventoryRefresh::TargetCollection
-      'VPC::TargetCollection'
+      "#{ManageIQ::Providers::Inflector.manager_type(ems.class)}::TargetCollection"
     else
       super
     end


app/models/manageiq/providers/ibm_cloud/inventory.rb
```diff
@@ -6,7 +6,8 @@ class ManageIQ::Providers::IbmCloud::Inventory < ManageIQ::Providers::Inventory
   def self.parsed_manager_name(target)
     case target
     when InventoryRefresh::TargetCollection
-      'VPC::TargetCollection'
+      ems_klass = target.instance_variable_get(:@manager).class
```
agrare, line 9: NOTE we should add `:manager` as an `attr_reader` to `InventoryRefresh::TargetCollection`, PR for this shortly

kbrock: were you going to add manager in the 21272 mentioned above? Or you thinking about adding that at another time?

agrare: We haven't built the collector instance yet, this is one step above that.  However it might be easier to just pass in `ems` here https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory.rb#L93

agrare: Okay I went that direction with this, now depends on https://github.com/ManageIQ/manageiq/pull/21273

agrare:
Backported to `morphy` in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/273

PR #229, opened 2021-06-24
Simplify VCR refresher spec by not duplicating ems

agrare:
There was a thought that duplicating the ems variable improved the speed of the specs, but running both side by side indicates that at best the non-duplicated version is (insignificantly) faster.  Likely it makes no difference at all and makes the spec harder to read.

PR #230, opened 2021-06-24
Fix missing VPC Flavor subclass

agrare:
There was no subclass for this one so the type was nil

Depends:
- [x] https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/229

Follow-up:
* Schema migration to set the type properly


Fryguy (approved):
LGTM - waiting on the schema changes

agrare:
Oh I forgot to link it back here, already in https://github.com/ManageIQ/manageiq-schema/pull/583

PR #241, opened 2021-07-07
Fix PowerVS VM 'cpu_type'

jaywcarman:
The PowerVS VM 'cpu_type' was not being set when the VM was provisioned from a "stock" image.


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -30,6 +30,10 @@ def parse
     sshkeys
   end
 
+  def architecture(image_id)
+    img_to_arch[image_id] ||= collector.image(image_id).specifications.architecture
```
agrare, line 34: @jaywcarman I'm assuming the issue that some instances are based on images which no longer are returned by `GET /images` is that right?

Since multiple instances could be provisioned from the same image this could result in the same API call being issues for the same image multiple times.

I never liked the `img_to_arch` hash, I think we should encapsulate this in the collector like:
```
def image(image_id)
  return images[image_id] if images[image_id] # NOTE this assumes each item returned by `GET /images` is equal to the item returned by `GET /images/:id`

  images[image_id] = images_api.pcloud_cloudinstances_images_get(cloud_instance_id, image_id)
end

def image_architecture(image_id)
  images[image_id]&.specifications&.architecture
end
```

If `GET /images` isn't just an array of what `GET /images/:id` would return then we'll have to keep the `@images` and whatever each `image()` call would return separate.

jaywcarman: > NOTE this assumes each item returned by `GET /images` is equal to the item returned by `GET /images/:id`

@agrare The image objects returned by `GET /images` ([pcloud-cloudinstances-images-getall](https://cloud.ibm.com/apidocs/power-cloud#pcloud-cloudinstances-images-getall)) contain a subset of the attributes returned by `GET /images/:id` ([pcloud-cloudinstances-images-get](https://cloud.ibm.com/apidocs/power-cloud#pcloud-cloudinstances-images-get)), _so_ I think we're safe to go with your idea to put all of the logic directly in the collector and minimize API calls. That seems to be an all-around better approach, so I'll push another patch up soon.

That said, in the existing patch I thought since we memoized the `img_to_arch` hash [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb#L18) that the `collector.image(image_id).specifications.architecture` would only occur the first time that `img_to_arch[image_id]` was `nil`? Was I mistaken?

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -19,14 +19,28 @@ def pvm_instance(instance_id)
     pvm_instances_api.pcloud_pvminstances_get(cloud_instance_id, instance_id)
   end
 
-  def image(img_id)
-    images_api.pcloud_cloudinstances_images_get(cloud_instance_id, img_id)
+  def image(image_id)
+    image = images.detect { |img| img.image_id == image_id }
```
jaywcarman, line 23: I'm not sure if the `detect` is the best lookup method here, but it will return `nil` if the image already already been fetched and saved to `images` by the `GET /images` call.

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -19,14 +19,28 @@ def pvm_instance(instance_id)
...
+    end
+    image
   end
 
   def images
     @images ||= images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images || []
   end
 
+  def image_architecture(image_id)
+    image = image(image_id)
```
jaywcarman, line 36: There are `image_id`s that are not returned by the `GET /images` call (I believe these are copies of "catalog" image and are associated with a single VM and can't be used for provisioning... :shrug:).

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
agrare: Better than looping through the array each time would be to store the images as a hash with the image_id as the hash key:
```
def images_by_id
  @images_by_id ||= images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images.index_by(&:image_id)
end

def images
  images_by_id.values
end
```

agrare (approved):

Fryguy:
Backported to `morphy` in commit 5880eb3ea0b080268f00f4a3cdd9ea79bb9cd481.

```text
commit 5880eb3ea0b080268f00f4a3cdd9ea79bb9cd481
Date:   Sat Jul 17 13:31:16 2021 -0400

    Merge pull request #241 from jaywcarman/fix/cpu_type
    
    Fix PowerVS VM 'cpu_type'
    
    (cherry picked from commit 3733e07b05ba66bac0df12f8980e0305fd41db9b)
```

PR #247, opened 2021-07-26
Image Import Dispatcher for IBM Cloud

iv1111:
* allowing one import session per PowerVC <--- PowerVS import pair
* depends on https://github.com/ManageIQ/manageiq/pull/21340

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
@@ -1,0 +1,29 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ImageImportWorkflow::ImportDispatcher < ::Job::Dispatcher
+  def dispatch
+    jobs_by_ems, = Benchmark.realtime_block(:pending_import_jobs) { pending_import_jobs }
+
+    jobs_by_ems.each do |ems_id, jobs|
+      free = true # TODO: here determine if there is a free slot
+      do_dispatch(job, ems_id) if free
+    end
+  end
+
+  def pending_import_jobs
+    pending_jobs.each_with_object(Hash.new { |h, k| h[k] = [] }) do |job, h|
+      h[job.options[:ems_id]] << job
+    end
+  end
+
+  def do_dispatch(job, ems_id)
+    MiqQueue.put_unless_exists(
+      :args        => [:start],
+      :class_name  => "Job",
+      :instance_id => job.id,
+      :method_name => "signal",
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => "smartstate",
+      :task_id     => job.guid,
+      :zone        => job.zone
+    )
+  end
+end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
@@ -3,26 +3,25 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ImageImp
   def dispatch
     pending, = Benchmark.realtime_block(:pending_import_jobs) { pending_import_jobs }
     running, = Benchmark.realtime_block(:running_import_jobs) { running_import_jobs }
-    busy_src_ids = running.map { |_, dst| dst.map { |job| job[:options][:src_provider_id] } }.flatten
+    busy_src_ids = running.flat_map{ |_, dst| dst.map { |job| job.options[:src_provider_id] }}.uniq
 
     pending.each do |dst_id, pending_jobs|
-      return if running[dst_id].present?
-      job = pending_jobs.detect { |x| busy_src_ids.exclude?(x[:options][:src_provider_id]) }
-      busy_src_ids << job[:options][:src_provider_id]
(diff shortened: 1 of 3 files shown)


iv1111:
@agrare Adding a custom dispatcher similar to https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/441/files.
Code runs but the list of pending jobs is empty:

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
```diff
@@ -0,0 +1,40 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ImageImportWorkflow::ImportDispatcher < ::Job::Dispatcher
+
+  def dispatch
+    pending, = Benchmark.realtime_block(:pending_import_jobs) { pending_import_jobs }
+    running, = Benchmark.realtime_block(:running_import_jobs) { running_import_jobs }
+    busy_src_ids = running.map { |_, dst| dst.map { |job| job[:options][:src_provider_id] } }.flatten
```
agrare, line 6: Minor, but `:options` is a column so you can do `job.options[:src_provider_id]`

agrare: `.map {}.flatten` can be just `running.flat_map`

agrare: You might want to `uniq` this set so:
`busy_src_ids = running.flat_map { |_, job| job.options[:src_provider_id] }.uniq`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
```diff
@@ -0,0 +1,40 @@
...
+      job = pending_jobs.detect { |x| busy_src_ids.exclude?(x[:options][:src_provider_id]) }
+      busy_src_ids << job[:options][:src_provider_id]
+      do_dispatch(job) unless job.nil?
+    end
+  end
+
+  def running_import_jobs
+    running_jobs.each_with_object(Hash.new { |h, k| h[k] = [] }) do |job, h|
+      h[job.options[:ems_id]] << job
+    end
```
agrare, line 19: Looks like you re-wrote `.group_by`, try:

```ruby
running_job.group_by { |job| job.options[:ems_id] }
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def do_dispatch(job)
+    MiqQueue.put_unless_exists(
+      :args        => [:start],
+      :class_name  => "Job",
+      :instance_id => job.id,
+      :method_name => "signal",
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => "smartstate",
```
agrare, line 35: Does this need to be on the `"smartstate"` role?  You can probably leave this out and let it run on the "generic"/default role

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow/import_dispatcher.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+  def pending_import_jobs
+    pending_jobs.each_with_object(Hash.new { |h, k| h[k] = [] }) do |job, h|
+      h[job.options[:ems_id]] << job
+    end
+  end
+
+  def do_dispatch(job)
```
agrare, line 28: Note to @agrare this looks like it can be in the base `Job::Dispatcher` class

iv1111:
@agrare I've followed your suggestions. One question. Can we be sure that there is only one JobDispatcher for a given job (here _ImageImportWorkflow_) at a given time running? It's important to know in order to guarantee there won't be any race conditions.

agrare:
@iv1111 the schedule worker will only put one instance of this onto the queue

iv1111:
> @iv1111 the schedule worker will only put one instance of this onto the queue

To make my question more specific, is there a scenario in the production environment where the _ImageImportWorkflow::ImportDispatcher.dispatch_ method would be called simultaneously from different threads?

iv1111:
@agrare If it's not the case then I'd have to put back the WIP label and think about using a mutex on a small portion of my code.

agrare:
@iv1111 there should only ever be one dispatcher running either per region.  The schedule worker has a put_unless_exists [[ref]](https://github.com/ManageIQ/manageiq/blob/ivanchuk/app/models/miq_schedule_worker/jobs.rb#L208) which means another won't be queued if one is already on the queue.

iv1111:
As discussed during the telco the scenario described by me can't happen.
Removing the [WIP] labell, we can merge it as soon as the dependent PR is merged.

agrare (approved):

agrare:
@iv1111 as a follow-up to this you'll need to create a schedule entry for this dispatcher, e.g. https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_schedule_worker/jobs.rb#L61-L65

iv1111:
@agrare I've created a corresponding PR: https://github.com/ManageIQ/manageiq/pull/21364

PR #248, opened 2021-07-28
IBM Cloud event collection

nasark:
As part of the work to implement event handling scenarios for VPC, I've completed the logic to collect events from IBM Cloud's event service called IBM Cloud Activity Tracker. A SDK has been generated for this service.

Moving forward the events will need to be parsed and used to queue targeted refreshes.

Note: A service key for IBM Cloud Activity Tracker is required in order to use the SDK (See: https://cloud.ibm.com/docs/activity-tracker?topic=activity-tracker-service_keys). For the time being, I created a record for the service key in my local environment to be passed in. But ideally this can be provided via a form when creating the provider.

Ref: https://cloud.ibm.com/apidocs/activity-tracker#export

Depends on:
* [x] [Add IBM Cloud Activity Tracker gem](https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/pull/60)

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
@@ -1,13 +1,14 @@
 class ManageIQ::Providers::IbmCloud::VPC::CloudManager < ManageIQ::Providers::CloudManager
   require_nested :AuthKeyPair
+  require_nested :EventCatcher
   require_nested :Flavor
-  require_nested :RefreshWorker
-  require_nested :Refresher
+  require_nested :LoggingMixin
   require_nested :Provision
   require_nested :ProvisionWorkflow
+  require_nested :RefreshWorker
+  require_nested :Refresher
   require_nested :Template
   require_nested :Vm
-  require_nested :LoggingMixin
 
   supports :provisioning
 
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
@@ -1,0 +1,33 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher::Stream
+  attr_reader :ems, :stop_polling, :poll_sleep
+
+  def initialize(ems, options = {})
+    @ems = ems
+    @stop_polling = false
+    @from = Time.now.to_i.to_s
+    @poll_sleep = options[:poll_sleep] || 20.seconds
+  end
+
+  def start
+    @stop_polling = false
+    @from = Time.now.to_i.to_s
+  end
+
+  def stop
+    @stop_polling = true
+  end
+
+  def poll
+    loop do
+      @to = Time.now.to_i.to_s
+
+      ems.with_provider_connection(:service => 'events') do |api|
+        events_client = api.sdk_client
+        events = events_client.export(:from => @from, :to => @to)
+        @from = @to
+        break if stop_polling
+      end
+      sleep(poll_sleep)
+    end
+  end
+end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
@@ -9,14 +9,20 @@ module ManageIQ::Providers::IbmCloud::VPC::ManagerMixin
     %w[auth_key]
   end
 
-  # Return a cloudtools vpc object.
+  # Return either a cloudtools vpc or activity tracker object.
   # @param options [Hash] Hash of options. Default to an empty Hash.
-  # @return [ManageIQ::Providers::IbmCloud::CloudTools::Vpc]
+  # @return [ManageIQ::Providers::IbmCloud::CloudTools::Vpc] or
+  # [ManageIQ::Providers::IbmCloud::CloudTools::ActivityTracker]
   def connect(options = {})
     key = authentication_key(options[:auth_type])
     region = options[:provider_region] || provider_region
     sdk = self.class.raw_connect(key)
-    sdk.vpc(:region => region)
+    if options[:service] == 'events'
+      service_key = authentication_key("events")
+      sdk.events(:region => region, :service_key => service_key)
+    else
+      sdk.vpc(:region => region)
+    end
   end
 
   # Same as calling connect.
--- a/lib/manageiq/providers/ibm_cloud/cloud_tool.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tool.rb
@@ -55,6 +55,10 @@ module ManageIQ
           @vpc ||= CloudTools::Vpc.new(:cloudtools => self, :region => region, :version => version, :generation => generation)
         end
 
+        def events(service_key:, region: 'us-east')
+          @events ||= CloudTools::ActivityTracker.new(:cloudtools => self, :region => region, :service_key => service_key)
+        end
+
         # Get a class that accesses the IBM Cloud Resource Manager API.
         # @return [CloudTools::ResourceController] the CloudTools Resource Manager SDK wrapper.
         def resource
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools.rb
@@ -2,7 +2,6 @@
 
 require 'logging'
 require 'ibm_vpc'
-require 'ibm_cloud_activity_tracker'
 
 require_relative 'cloud_tools/sdk'
 dir_name = File.basename(__FILE__).split('.')[0]
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools.rb
@@ -2,6 +2,7 @@
 
 require 'logging'
 require 'ibm_vpc'
+require 'ibm_cloud_activity_tracker'
 
 require_relative 'cloud_tools/sdk'
 dir_name = File.basename(__FILE__).split('.')[0]
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/activity_tracker.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/activity_tracker.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+module ManageIQ
+  module Providers
+    module IbmCloud
+      module CloudTools
+        class ActivityTracker < Sdk::Branch
+          def initialize(cloudtools:, region:, service_key:)
+            super(:cloudtools => cloudtools)
+            service_url = "https://api.#{region}.logging.cloud.ibm.com"
+            @sdk_params = {:service_url => service_url, :service_key => service_key}
+          end
+
+          # Return a new Activty Tracker SDK instance.
+          def sdk_client
+            @sdk_params[:authenticator] = @cloudtools.authenticator
+            @sdk_client ||= IbmCloudActivityTracker::ActivityTrackerApiV1.new(@sdk_params)
+          end
+
+          private
+
+          # Create a generator that removes the need for pagination.
+          # @param call_back [String] The method name to use for pagination.
+          # @param array_key [String] The specific key in the returned array to use.
+          #
+          # @return [Enumerator] Object to page through results.
+          # @yield [Hash] Result of request.
+          def each_resource(call_back, array_key: nil, **kwargs)
+            start = kwargs[:start]
+
+            loop do
+              # Send request.
+              response = request(call_back, :start => start, **kwargs)
+              array_key = get_common(response.keys) if array_key.nil?
+
+              resources = response.fetch(array_key.to_sym)
+
+              resources&.each { |value| yield value }
+
+              # VPC has a next key that holds the next URL.
+              return unless response.key?(:next)
+
+              # The :next data structure is a hash with a href member.
+              next_url = response.dig(:next, :href)
+              return unless next_url
+
+              start = Addressable::URI.parse(next_url).query_values['start']
+            end
+          end
+
+          #  Try to determine
+          def get_common(array_keys)
+            keys = array_keys - [:limit, :first, :total_count, :next]
+            return keys.first unless keys.empty?
+
+            raise "No array key found in #{array_keys}"
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher.rb
@@ -1,0 +1,3 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher < ManageIQ::Providers::BaseManager::EventCatcher
+  require_nested :Runner
+end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/runner.rb
@@ -1,0 +1,21 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher::Runner < ManageIQ::Providers::BaseManager::EventCatcher::Runner
+  def monitor_events
+    event_monitor_handle.start
+    event_monitor_running
+    event_monitor_handle.poll do |event|
+      @queue.enq(event)
+    end
+  ensure
+    stop_event_monitor
+  end
+
+  def stop_event_monitor
+    event_monitor_handle.stop
+  end
+
+  private
+
+  def event_monitor_handle
+    @event_monitor_handle ||= self.class.parent::Stream.new(@ems)
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -8,6 +8,12 @@
           :critical:
           - IBM_CLOUD_POWER_VIRTUAL_SERVERS_instance_power_on
           - IBM_CLOUD_POWER_VIRTUAL_SERVERS_instance_power_off
+  :ems_ibm_cloud_vpc:
+    :event_handling:
+      :event_groups:
+        :addition:
+          :critical:
+          - is.instance.instance.create
 
 :ems_refresh:
   :ibm_vpc:
@@ -24,6 +30,8 @@
     :event_catcher:
       :event_catcher_ibm_cloud_power_virtual_servers:
         :poll: 20.seconds
+      :event_catcher_vpc:
+        :poll: 15.seconds
     :queue_worker_base:
       :ems_refresh_worker:
         :ems_refresh_worker_ibm_cloud_power_virtual_servers: {}
--- a/manageiq-providers-ibm_cloud.gemspec
+++ b/manageiq-providers-ibm_cloud.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "ibm_cloud_power", "~> 1.0"
   spec.add_dependency "ibm_cloud_resource_controller", "~> 2.0"
   spec.add_dependency "ibm-cloud-sdk", "~> 0.1"
+  spec.add_dependency "ibm_cloud_activity_tracker", "~> 0.1"
   spec.add_dependency "ibm_cloud_global_tagging", "~> 0.1"
   spec.add_dependency "ibm_vpc", "~> 0.1"
 
--- a/systemd/manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher.target
+++ b/systemd/manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher.target
@@ -1,0 +1,2 @@
+[Unit]
+PartOf=manageiq.target
--- a/systemd/manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher@.service
+++ b/systemd/manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher@.service
@@ -1,0 +1,13 @@
+[Unit]
+PartOf=manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher.target
+[Install]
+WantedBy=manageiq-providers-ibm_cloud_vpc_cloud_manager_event_catcher.target
+[Service]
+WorkingDirectory=/var/www/miq/vmdb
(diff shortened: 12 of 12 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def stop
+    @stop_polling = true
+  end
+
+  def poll
+    loop do
+      @to = DateTime.current.to_time.to_i.to_s
+
+      ems.with_provider_connection(:service => 'ActivityTrackerApiV1', :provider_region => 'us-south') do |api|
```
nasark, line 25: The region that is being passed in here is the region in which the Activity Tracker instance is located in. If there is a better way to go about this please let me know

lib/manageiq/providers/ibm_cloud/cloud_tools/activity_tracker.rb
```diff
@@ -0,0 +1,66 @@
...
+          def initialize(cloudtools:, region:)
+            super(:cloudtools => cloudtools)
+            service_url = "https://api.#{region}.logging.cloud.ibm.com"
+            @sdk_params = {:service_url => service_url}
+          end
+
+          # Return a new Activty Tracker SDK instance.
+          def sdk_client
+            @sdk_params[:authenticator] = @cloudtools.authenticator
+            service_key = AuthUseridPassword.find_by(:authtype => "events").auth_key
```
nasark, line 17: This is to access the record which was created in my local env to access the service key. The record was created using:
```
vpc = ManageIQ::Providers::IbmCloud::VPC::CloudManager.first
vpc.authentications.create!(:authtype => "events", :auth_key => <service_key>)
```

nasark:
@miq-bot add-labels enhancement

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -16,7 +16,11 @@ def connect(options = {})
     key = authentication_key(options[:auth_type])
     region = options[:provider_region] || provider_region
     sdk = self.class.raw_connect(key)
-    sdk.vpc(:region => region)
+    if options[:service] == 'ActivityTrackerApiV1'
```
agrare, line 19: since we're calling the service `"events"` belong (as in `sdk.events`) how about simply `"events"` as the service name here?

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -16,7 +16,11 @@ def connect(options = {})
...
+      sdk.events(:region => region)
```
agrare, line 20: So this is where we're going to want to pull in the service key since we have the proper provider context, rather than https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/248/files#diff-ecb90bb1257297ad164960093c29c4d725223dbcfd4a9bd36cbbb6f2f30a5ac4R17 which would be the same for all provider instances.

You can do `service_key = authentication_key("events")` here

lib/manageiq/providers/ibm_cloud/cloud_tools/activity_tracker.rb
agrare: So the problem with doing this here is that `AuthUseridPassword.find_by()` would look through the whole "authentications" table not just for this provider type or even this provider.

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
nasark: sure, that makes more sense

nasark:
@agrare Thanks for the feedback, I've made the changes. Let me know if you have any further suggestions.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def stop
+    @stop_polling = true
+  end
+
+  def poll
+    loop do
+      @to = DateTime.current.to_time.to_i.to_s
+
+      ems.with_provider_connection(:service => 'events', :provider_region => 'us-south') do |api|
```
agrare, line 24: We should be able to pull the `provider_region` out of the ems record in the `#connect` method down here https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/248/files#diff-9c1c1fe8a8cf1522f3f23dad6743d18f6c423658360cd877f66c9cb7ac3a1753R21

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -16,7 +16,12 @@ def connect(options = {})
     key = authentication_key(options[:auth_type])
     region = options[:provider_region] || provider_region
     sdk = self.class.raw_connect(key)
-    sdk.vpc(:region => region)
+    if options[:service] == 'events'
+      service_key = authentication_key("events")
+      sdk.events(:region => region, :service_key => service_key)
```
agrare, line 21: So if you need the provider region here, you can get that by doing `sdk.events(:region => region, :service_key => service_key, :provider_region => provider_region)` since its an attribute of the ext_management_system

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
agrare: Oh, do you know if the Activity Tracking isn't available on each region?  Usually our provider is added for a specific region so you can just use that.

lib/manageiq/providers/ibm_cloud/cloud_tool.rb
```diff
@@ -55,6 +55,10 @@ def vpc(region: 'us-east', version: '2021-01-01', generation: 2)
           @vpc ||= CloudTools::Vpc.new(:cloudtools => self, :region => region, :version => version, :generation => generation)
         end
 
+        def events(region: 'us-south', service_key:)
```
agrare, line 58: We should probably default to the same region as vpc above: `def vpc(region: 'us-east',` as long as the service is available in that region

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
nasark: It is available in the following regions: https://cloud.ibm.com/docs/activity-tracker?topic=activity-tracker-endpoints, which I think is all the regions? In that case should we assume that the activity tracker instance is provisioned in the same region as the provider?

nasark:
The Activity Tracker service is available in the same regions as the provider and so the provider's region will be used when interfacing with the Activity Tracker SDK

https://cloud.ibm.com/docs/activity-tracker?topic=activity-tracker-endpoints

agrare:
Looking good, so travis is failing because you have a worker which doesn't have a systemd service file.  You can copy another event catcher like https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/systemd/manageiq-providers-amazon_cloud_manager_event_catcher%40.service and the target https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/systemd/manageiq-providers-amazon_cloud_manager_event_catcher.target

nasark:
It seems like my gem isn't getting picked up although I've pushed it up here: https://rubygems.org/gems/ibm_cloud_activity_tracker

agrare:
@nasark you need to add it to the [[gemspec]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/manageiq-providers-ibm_cloud.gemspec#L21-L26)

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
```diff
@@ -8,6 +8,7 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager < ManageIQ::Providers::Cl
   require_nested :Template
   require_nested :Vm
   require_nested :LoggingMixin
+  require_nested :EventCatcher
```
Fryguy, line 11: alphabetical? (not sure if this was ordered at the end on purpose or not)

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,34 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher::Stream
+  attr_reader :ems, :stop_polling, :poll_sleep
+
+  def initialize(ems, options = {})
+    @ems = ems
+    @stop_polling = false
+    @from = DateTime.current.to_time.to_i.to_s
```
Fryguy, line 7: Maybe not for this PR, if this was copied from others, but why are we using `DateTime.current.to_time` (which comes from ActiveSupport), as opposed to `Time.now` (which is in Ruby core)?  The latter is simpler and gives the same number.

cc @agrare

Fryguy:
```ruby
[6] pry(main)> [DateTime.current.to_time.to_i, Time.now.to_i]
=> [1627569302, 1627569302]
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,34 @@
...
+  def stop
+    @stop_polling = true
+  end
+
+  def poll
+    loop do
+      @to = DateTime.current.to_time.to_i.to_s
+
+      ems.with_provider_connection(:service => 'events') do |api|
+        # Replace with your provider's events API
```
Fryguy, line 25: I think this comment can go away :)

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
Fryguy: Separate question, more for @agrare, but do we have a way to record the last "key" that we ran against?  The reason I'm asking is that if the event catcher goes down, we will lose everything during the down time.  I thought we had a way to record the last "key" that we saw, so we can pick up where we left off.

If not, maybe we should take the last event that was recorded in the table and use that timestamp?

nasark: Ah didn't know there was a simpler way to do it, thank you

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
nasark: It was not on purpose, thanks I will order the other `require_nested` statements in this file as well while I'm at it

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
```diff
@@ -16,7 +16,12 @@ def connect(options = {})
...
+    else
+      sdk.vpc(:region => region)
+    end
```
Fryguy, line 24: Not necessary for this PR, but this feels like it should be broken up into smaller methods and/or the raw_connect method should take the "service" as a parameter (I think that's how we do it for amazon).

@agrare Thoughts on this?  I know in other places you have individual methods like `def vpc`, `def events`, but I can't recall at what level you had those methods.

Fryguy: Also with this addition the comment on line 12 and 14 is now Basmati Rice (https://twitter.com/nzkoz/status/538892801941848064)

Fryguy:
This is great work @nasark ... looking forward to this one!

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
agrare: @Fryguy if the API is expecting the start point to be timestamp based then we can do `ems.ems_events.order(:timestamp).last`

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -0,0 +1,34 @@
...
+        events_client = api.sdk_client
+        events = events_client.export(:from => @from, :to => @to)
```
agrare, line 27: Think you need to `yield events` here, actually the caller is expecting one event at a time so `events.each { |e| yield e }` or take a `&block` like rubocop was recommending and to `events.each(&block)`

nasark: `events` returned from the API is a `IBMCloudSdkCore::DetailedResponse` object where the response is a single string containing all the events that occurred during the `@from -> @to` timeframe. Therefore I am not able to use `.each` here. Let me know if my understanding is incorrect or if there is another way.

i.e.
```
#<IBMCloudSdkCore::DetailedResponse:0x00007fdbfbca21d0 @status=200, @headers={"Server"=>"restify", "Content-Security-Policy"=>"frame-ancestors 'none'", "Strict-Transport-Security"=>"max-age=31536000; includeSubDomains", "Date"=>"Thu, 29 Jul 2021 15:59:47 GMT", "Connection"=>"close", "Transfer-Encoding"=>"chunked"}, @result="{\"_account\":\"c6e1d98061\",\"_cluster\":\"ld86\",\"_host\":\"containers-kubernetes\",\"_ingester\":\"supertenant\",\"_logtype\":\"json\",\"_file\":\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\",\"_line\":\"{\\\"action\\\":\\\"containers-kubernetes.cluster-ingress-status.get\\\",\\\"correlationId\\\":\\\"ff9f6161-9c2d-4cba-b6df-81581d05d580\\\",\\\"dataEvent\\\":true,\\\"eventTime\\\":\\\"2021-07-29T15:41:52.29+0000\\\",\\\"initiator\\\":{\\\"id\\\":\\\"crn:v1:bluemix:public:containers-kubernetes\\\",\\\"name\\\":\\\"armada-alb-api\\\",\\\"typeURI\\\":\\\"service/security/account/service\\\",\\\"host\\\":{\\\"address\\\":\\\"\\\",\\\"addressType\\\":\\\"\\\"},\\\"credential\\\":{\\\"type\\\":\\\"\\\"}},\\\"logSourceCRN\\\":\\\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\\\",\\\"message\\\":\\\"Kubernetes Service: get cluster-ingress-status c31pbtiw0gfedg1bna4g\\\",\\\"observer\\\":{\\\"name\\\":\\\"ActivityTracker\\\"},\\\"outcome\\\":\\\"success\\\",\\\"reason\\\":{\\\"reasonCode\\\":200,\\\"reasonType\\\":\\\"OK\\\",\\\"reasonForFailure\\\":\\\"\\\"},\\\"resourceGroupID\\\":\\\"crn:v1:bluemix:public:resource-controller::a/c56c9a268d23e1b339ac14774358133c::resource-group:345c433098294722ba52d9039133e8cf\\\",\\\"saveServiceCopy\\\":true,\\\"severity\\\":\\\"normal\\\",\\\"target\\\":{\\\"id\\\":\\\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\\\",\\\"name\\\":\\\"c31pbtiw0gfedg1bna4g\\\",\\\"typeURI\\\":\\\"containers-kubernetes/cluster\\\",\\\"host\\\":{},\\\"resourceGroupId\\\":\\\"crn:v1:bluemix:public:resource-controller::a/c56c9a268d23e1b339ac14774358133c::resource-group:345c433098294722ba52d9039133e8cf\\\"},\\\"id\\\":\\\"ff9f6161-9c2d-4cba-b6df-81581d05d580\\\"}\",\"_rawline\":null,\"level\":\"normal\",\"_ts\":1627574361281,\"_platform\":\"containers-kubernetes\",\"_app\":\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\",\"_ip\":\"166.9.14.3\",\"__key\":\"logline:c6e1d98061:crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g:::containers-kubernetes:ld86\",\"_bid\":\"852a8e56-6e9a-4c7b-8671-c6b3ef2cabe5:1302:ld86\",\"action\":\"containers-kubernetes.cluster-ingress-status.get\",\"correlationId\":\"ff9f6161-9c2d-4cba-b6df-81581d05d580\",\"dataEvent\":true,\"eventTime\":\"2021-07-29T15:41:52.29+0000\",\"logSourceCRN\":\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\",\"message\":\"Kubernetes Service: get cluster-ingress-status c31pbtiw0gfedg1bna4g\",\"outcome\":\"success\",\"resourceGroupID\":\"crn:v1:bluemix:public:resource-controller::a/c56c9a268d23e1b339ac14774358133c::resource-group:345c433098294722ba52d9039133e8cf\",\"saveServiceCopy\":true,\"severity\":\"normal\",\"id\":\"ff9f6161-9c2d-4cba-b6df-81581d05d580\",\"_id\":\"1373946466414157825\",\"initiator\":{\"id\":\"crn:v1:bluemix:public:containers-kubernetes\",\"name\":\"armada-alb-api\",\"typeURI\":\"service/security/account/service\",\"host\":{\"address\":\"\",\"addressType\":\"\"},\"credential\":{\"type\":\"\"}},\"observer\":{\"name\":\"ActivityTracker\"},\"reason\":{\"reasonCode\":200,\"reasonType\":\"OK\",\"reasonForFailure\":\"\"},\"target\":{\"id\":\"crn:v1:bluemix:public:containers-kubernetes:us-east:a/c56c9a268d23e1b339ac14774358133c:c31pbtiw0gfedg1bna4g::\",\"name\":\"c31pbtiw0gfedg1bna4g\",\"typeURI\":\"containers-kubernetes/cluster\",\"resourceGroupId\":\"crn:v1:bluemix:public:resource-controller::a/c56c9a268d23e1b339ac14774358133c::resource-group:345c433098294722ba52d9039133e8cf\",\"host\":{}}}\n">
```

lib/manageiq/providers/ibm_cloud/cloud_tools.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'logging'
 require 'ibm_vpc'
+require 'ibm_cloud_activity_tracker'
```
nasark, line 5: Travis is failing with: 
```
Failure/Error: require 'ibm_cloud_activity_tracker'
LoadError:
  cannot load such file -- ibm_cloud_activity_tracker
# ./lib/manageiq/providers/ibm_cloud/cloud_tools.rb:5:in `<top (required)>'
# ./spec/models/manageiq/providers/ibm_cloud/vpc/cloud_tools_spec.rb:3:in `<top (required)>'
```

@agrare Do I need to require the gem here if its already listed as a dependency in the gemspec?

app/models/manageiq/providers/ibm_cloud/vpc/manager_mixin.rb
agrare: Passing the service in as an option and switching on it is pretty common, you're thinking of the google cloud collector where I was caching the various service connections.

We could bump this logic up to the `.raw_connect` method if that would help but at this point that is only used by this method and `verify_credentials` which doesn't need any specific service (just checking the IAM token is valid)

agrare (approved):

agrare:
Looks like the cloudtools api wrapper for ibm cloud activity tracker, the worker class, and the stream thread are all here.  Great first step we can iterate on

Fryguy:
@nasark A conflict occurred during the backport of this pull request to `morphy`.

Conflict details:

```diff
diff --cc app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
index 4fb6a8b,3f8d1a1..0000000
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager.rb
@@@ -1,12 -1,14 +1,19 @@@
  class ManageIQ::Providers::IbmCloud::VPC::CloudManager < ManageIQ::Providers::CloudManager
    require_nested :AuthKeyPair
++<<<<<<< HEAD
 +  require_nested :RefreshWorker
 +  require_nested :Refresher
++=======
+   require_nested :EventCatcher
+   require_nested :Flavor
+   require_nested :LoggingMixin
++>>>>>>> 66a5620... Merge pull request #248 from nasark/vpc_event_handling
    require_nested :Provision
    require_nested :ProvisionWorkflow
+   require_nested :RefreshWorker
+   require_nested :Refresher
    require_nested :Template
    require_nested :Vm
-   require_nested :LoggingMixin
  
    supports :provisioning
  
diff --cc lib/manageiq/providers/ibm_cloud/cloud_tool.rb
index 3fd1b6e,91cd310..0000000
--- a/lib/manageiq/providers/ibm_cloud/cloud_tool.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tool.rb
@@@ -55,6 -55,16 +55,19 @@@ module ManageI
            @vpc ||= CloudTools::Vpc.new(:cloudtools => self, :region => region, :version => version, :generation => generation)
          end
  
++<<<<<<< HEAD
++=======
+         def events(service_key:, region: 'us-east')
+           @events ||= CloudTools::ActivityTracker.new(:cloudtools => self, :region => region, :service_key => service_key)
+         end
+ 
+         # Get a class that accesses the IBM Cloud Resource Manager API.
+         # @return [CloudTools::ResourceController] the CloudTools Resource Manager SDK wrapper.
+         def resource
+           @resource ||= CloudTools::ResourceController.new(:cloudtools => self)
+         end
+ 
++>>>>>>> 66a5620... Merge pull request #248 from nasark/vpc_event_handling
          private
  
          # Set a new logger
```

Fryguy:
Backported to `morphy` via https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/271

PR #249, opened 2021-07-28
disabling warning dialog on redirect from image import form

iv1111:
- depends on https://github.com/ManageIQ/manageiq-ui-classic/pull/7799

### DIFF
--- a/app/javascript/components/import-image-form/import-image-form.jsx
+++ b/app/javascript/components/import-image-form/import-image-form.jsx
@@ -80,7 +80,7 @@ const ImportImageForm = ({ dispatch }) => {
     };
 
     const onCancel = () => {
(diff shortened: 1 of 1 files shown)


iv1111:
@kavyanekkalapu I have tried out the solution and it does work. Thanks!  
Using _dispatch_ in the close-callback this way did the trick:

`ManageIQ.redux.store.dispatch({ type: 'FormButtons.reset' })`

Could you merge your PR so that we can merge this one too.

iv1111:
Dependency is resolved, I guess we can merge this one.

agrare:
@kavyanekkalapu can you review?

kavyanekkalapu (approved):

PR #252, opened 2021-08-05
Added EventParser class

nasark:
- added the EventParser class which will create the relevant schemas and add the necessary details when a `create instance` event is captured (to be used when implementing functionality to trigger targeted refreshes)
- modified event collection API call to return events that are hosted by "is". This means that events will be filtered so that we only receive VPC related events (see: https://cloud.ibm.com/docs/vpc?topic=vpc-at-events)
- modified event collection API call to return a hash array of events

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
@@ -1,0 +1,27 @@
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    event_hash = {
+      :event_type => event["action"],
+      :source     => "IBM_CLOUD_VPC",
+      :ems_id     => ems_id,
+      :ems_ref    => event["_id"],
+      :timestamp  => event["eventTime"],
+      :full_data  => event
+    }
+
+    case event_hash[:event_type]
+    when 'is.instance.instance.create'
+      parse_vm_event!(event, event_hash)
+    end
+
+    event_hash
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    return if event['responseData'].nil?
+
+    event_hash[:vm_uid_ems] = event.dig('responseData', 'id')
+    event_hash[:vm_ems_ref] = event.dig('responseData', 'id')
+    event_hash[:vm_name]    = event.dig('responseData', 'name')
+  end
+end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/runner.rb
@@ -13,6 +13,11 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher::Runner < M
     event_monitor_handle.stop
   end
 
+  def queue_event(event)
+    event_hash = ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser.event_to_hash(event, @cfg[:ems_id])
+    EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
+  end
+
   private
 
   def event_monitor_handle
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    return if event['responseData'].nil?
+
+    event_hash[:timestamp] =  event["eventTime"]
+    event_hash[:vm_uid_ems] = event.dig('responseData', 'id')
+    event_hash[:vm_ems_ref] = event.dig('responseData', 'id')
+    event_hash[:vm_name]    = event.dig('responseData', 'name')
+  end
```
nasark, line 25: Let me know if we should add anything else here, the following is available:
```
event['responseData'].keys

["bandwidth", "created_at", "crn", "disks", "href", "id", "memory", "name", "resource_type", "startable", "status", "status_reasons", "boot_volume_attachment", "image", "network_interfaces", "primary_network_interface", "profile", "resource_group", "vcpu", "volume_attachments", "vpc", "zone"]
```

nasark:
@miq-bot add-labels enhancement

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,26 @@
...
+      parse_vm_event!(event, event_hash)
+    end
+
+    event_hash
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    return if event['responseData'].nil?
+
+    event_hash[:timestamp] =  event["eventTime"]
```
agrare, line 21: We need timestamp in the general event case, can this be moved up [[here]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/252/files#diff-d705b2bd25d0350d720927209af1307766723f626713b08dd66f14a1b000c4b0R7) or does eventTime only show up for `'is.instance.instance.create'` type events?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,26 @@
+module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    event_hash = {
+      :event_type => event["action"],
+      :source     => "IBM_CLOUD_VPC",
+      :ems_ref    => ems_id,
```
agrare, line 6: `:ems_ref` should be the unique identifier of _this event_

Does the `event` have some sort of unique id?

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
nasark: sure, `eventTime` is available for all events

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
nasark: Yes there seems to be a unique identifier:
```
> event["_id"]
"1376197173894295555"
```

Is there any use for `@cfg[:ems_id]` in this method?

agrare: > Is there any use for @cfg[:ems_id] in this method?

Yes the event table has an `:ems_id` column so it would be good to set that

agrare (approved):

Fryguy:
Backported to `morphy` in commit 32e9b426d42e87e084cebee1a9c74d6d6f08c4cb.

```text
commit 32e9b426d42e87e084cebee1a9c74d6d6f08c4cb
Date:   Thu Aug 5 12:37:35 2021 -0400

    Merge pull request #252 from nasark/vpc_event_handling
    
    Added EventParser class
    
    (cherry picked from commit 4237f0414d47dcb2d6799c18fb5f845a913d4c8a)
```

PR #253, opened 2021-08-10
Validator task for a PowerVC node in PVC <-- PVS import workflow

iv1111:
- validator will run before the import workflow
& fail if the PowerVC node set-up is incomplete

- performing validation via ansible


iv1111:
@agrare removing WIP

agrare (approved):

PR #256, opened 2021-08-14
Fix PowerVS image collector to handle elusive image IDs

jaywcarman:
Two cases that these commits address:

1) When an image ID is completely missing, instead of breaking the refresh we can continue on. The result is that the `cpu_type` will be `nil`.

2) Add an API call to the stock image catalog. Depending on how you provision a VM, the associated image reference could be to this stock catalog, which is not returned in the regular list of images for a PowerVS instance.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -24,14 +24,14 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
       images_by_id[image_id] ||= images_api.pcloud_cloudinstances_images_get(cloud_instance_id, image_id)
     rescue IbmCloudPower::ApiError => err
       error_message = JSON.parse(err.response_body)["description"]
-      _log.warn("ImageID not found: #{error_message}")
+      _log.debug("ImageID not found: #{error_message}")
     end
 
     begin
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -20,7 +20,20 @@ def pvm_instance(instance_id)
   end
 
   def image(image_id)
-    images_by_id[image_id] ||= images_api.pcloud_cloudinstances_images_get(cloud_instance_id, image_id)
+    begin
+      images_by_id[image_id] ||= images_api.pcloud_cloudinstances_images_get(cloud_instance_id, image_id)
+    rescue IbmCloudPower::ApiError => err
+      error_message = JSON.parse(err.response_body)["description"]
+      _log.warn("ImageID not found: #{error_message}")
```
agrare, line 27: Unless it is actually an issue if an image is a stock image we probably don't want to be logging a warning in this case since if it is found by https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/256/files#diff-d244b92097e187e59c19aceb5f90f45f0de129bb130fa3d8d0cd1682a2e686e4R31 everything will be successful but a warning will be logged every time.

jaywcarman: I switched to `debug` level, since the message is already parsed out and was helpful in my dev environment.

agrare (approved):

PR #257, opened 2021-08-16
Remove unnecessary PowerVS volume API calls

jaywcarman:
We're collecting all volumes through a single API call and then making an additional API call for each volume. All of the information needed in the parser is available from the first "get all" call.


agrare (approved):

agrare:
@jaywcarman you can either re-record the VCR or delete the API calls that aren't used anymore

PR #259, opened 2021-08-18
VPC event targeted refresh for VMs

nasark:
- added event-based targeted refresh for VMs

Depends on:
* [x] [Add IBM Cloud VPC VM event handler](https://github.com/ManageIQ/manageiq-content/pull/689)

Ref:
- https://cloud.ibm.com/docs/vpc?topic=vpc-at-events#events-compute

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
@@ -2,7 +2,7 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
   def self.event_to_hash(event, ems_id)
     event_hash = {
       :event_type => event["action"],
-      :source     => "IBM_CLOUD_VPC",
+      :source     => "IBMCloud-VPC",
       :ems_id     => ems_id,
       :ems_ref    => event["_id"],
       :timestamp  => event["eventTime"],
@@ -10,7 +10,7 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
     }
 
     case event_hash[:event_type]
-    when 'is.instance.instance.create'
+    when /^is\.instance/
       parse_vm_event!(event, event_hash)
     end
 
@@ -18,10 +18,10 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
   end
 
   def self.parse_vm_event!(event, event_hash)
-    return if event['responseData'].nil?
+    vm_ref = event.dig('target', 'id').match(/::instance:(?<instance_id>\S+)$/)
+    return if vm_ref.nil?
 
-    event_hash[:vm_uid_ems] = event.dig('responseData', 'id')
-    event_hash[:vm_ems_ref] = event.dig('responseData', 'id')
-    event_hash[:vm_name]    = event.dig('responseData', 'name')
+    event_hash[:vm_uid_ems] = vm_ref.named_captures["instance_id"]
+    event_hash[:vm_ems_ref] = vm_ref.named_captures["instance_id"]
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_target_parser.rb
@@ -1,0 +1,28 @@
+class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventTargetParser
+  attr_reader :ems_event
+
+  def initialize(ems_event)
+    @ems_event = ems_event
+  end
+
+  def parse
+    parse_ems_event_targets(ems_event)
+  end
+
+  def parse_ems_event_targets(event)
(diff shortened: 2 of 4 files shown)


nasark:
@miq-bot add-labels enhancement

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def parse_ems_event_targets(event)
+    target_collection = InventoryRefresh::TargetCollection.new(
+      :manager => event.ext_management_system,
+      :event   => event
+    )
+
+    case event[:event_type]
+    when 'is.instance.instance.create'
```
agrare, line 19: @nasark we should probably make this more general, `when /^is\.instance/` should cover all vm events

nasark: @agrare so we want a targeted refresh for all VM events?

nasark: when creating an instance, that usually fires off multiple `is.instance` events and so that would mean multiple targeted refreshes would be kicked off for the same vm in a short timeframe

agrare: Yeah that's fine, we group multiple targets into a single refresh so if they come in in rapid succession then they will be refreshed together

nasark: for example, when an instance is created on IBM Cloud, the event service fires off `is.instance.instance.create`, `is.instance.instance.start`, and multiple `is.instance.instance.read` events in quick succession

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
```diff
@@ -2,26 +2,25 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
...
 
   def self.parse_vm_event!(event, event_hash)
-    return if event['responseData'].nil?
+    return if event.dig('target', 'id').partition('instance:').last == ""
 
-    event_hash[:vm_uid_ems] = event.dig('responseData', 'id')
-    event_hash[:vm_ems_ref] = event.dig('responseData', 'id')
-    event_hash[:vm_name]    = event.dig('responseData', 'name')
+    event_hash[:vm_uid_ems] = event.dig('target', 'id').partition('instance:').last
+    event_hash[:vm_ems_ref] = event.dig('target', 'id').partition('instance:').last
```
nasark, line 24: This had to be changed because some instance events did not contain the vm ref and those that did contain the ref did not have it readily available in `responseData` as was the case for `is.instance.instance.create`. Therefore the ref is now obtained by parsing it from a longer target ID.

i.e. 
```
crn:v1:bluemix:public:is:us-south-2:a/c56c9a268d23e1b339ac14774358133c::instance:0727_3fa23c25-3e03-46f3-a503-d459953816ef
```

If there is a better way of going about this let me know

agrare: I think a regular expression would be safer here:
```
.match(/::instance:(?<instance_id>\S+)$/).named_captures["instance_id"]
```

agrare: You might want to pull this out into a method like
```
instance_id = parse_event_instance_id(event)
return if instance_id.nil?
```

agrare (approved):

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_parser.rb
```diff
@@ -2,26 +2,26 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventParser
...
+    when /^is\.instance/
       parse_vm_event!(event, event_hash)
     end
 
     event_hash
   end
 
   def self.parse_vm_event!(event, event_hash)
-    return if event['responseData'].nil?
+    vm_ref = event.dig('target', 'id').match(/::instance:(?<instance_id>\S+)$/)
```
agrare, line 21: Super minor and can be done in a follow-up, but you can do `.match()&.named_captures&.dig("instance_id")` that way you don't have `vm_ref.named_captures["instance_id"]` in multiple places below

Fryguy:
Backported to `morphy` in commit b4d93ad2d23eef3f81d3e2eac0a8fdd1d2c14460.

```text
commit b4d93ad2d23eef3f81d3e2eac0a8fdd1d2c14460
Date:   Thu Aug 19 16:31:13 2021 -0400

    Merge pull request #259 from nasark/vpc_event_handling
    
    VPC event targeted refresh for VMs
    
    (cherry picked from commit 8ab4eb54635175ad5d324578ec5ab22de50e0b58)
```

PR #261, opened 2021-08-23
passing encryption key, initialization and cyphertext using Authentication object

iv1111:
via Authentications object


iv1111:
@agrare here is the corresp. PR. I tested it and it does indeed remove the Auth. object on manager "destroy". Needs more testing, not removing WIP yet.

iv1111:
Depends on: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/261

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -22,6 +22,12 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
           :inverse_of  => :parent_manager,
           :dependent   => :destroy
 
+  has_one :import_auth,
```
agrare, line 25: recommend `has_many` here, even if at first you only want to enable one-at-a-time in the future if you want to make that configurable you'll have to change this

iv1111: changed in code, now testing

PR #265, opened 2021-08-29
Refactor SDK connections

nasark:
In the case of adding other SDKs for inventory collection, this was done so that we no longer need to re-authenticate for a connection other than VPC since we will now be using the same `CloudTool` object. In the near future this will be relevant for cloud database collection

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
@@ -6,6 +6,7 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::EventCatcher::Stream
     @stop_polling = false
     @from = Time.now.to_i
     @poll_sleep = options[:poll_sleep] || 20.seconds
+    @service_key = ems.authentication_key("events")
   end
 
   def start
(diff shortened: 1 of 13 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
```diff
@@ -21,8 +22,8 @@ def poll
     loop do
       @to = Time.now.to_i
 
-      ems.with_provider_connection(:service => 'events') do |api|
-        events_client = api.sdk_client
+      ems.with_provider_connection do |connect|
+        events_client = connect.events(:region => ems.provider_region, :service_key => @service_key).sdk_client
```
nasark, line 26: Looking for suggestions here, it would be ideal if we didn't have to re-auth for every poll

nasark:
@miq-bot add-labels refactoring

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb
agrare: Yeah agreed, better to pull the connection out of the loop like
```ruby
events_client = ems.connect.events(:region => ems.provider_region, :service_key => @service_key).sdk_client
loop do
  ...
end
```

nasark: @agrare thanks, I've updated with your suggestion

agrare (approved):

Fryguy:
Backported to `morphy` in commit 7aca0f5d0af3f1c6866e33dc0c23fefe22f1bb8f.

```text
commit 7aca0f5d0af3f1c6866e33dc0c23fefe22f1bb8f
Date:   Mon Aug 30 12:11:14 2021 -0400

    Merge pull request #265 from nasark/refactor_sdk_connections
    
    Refactor SDK connections
    
    (cherry picked from commit 6acc72adac2fc405b9f0db893074d85f907e2db4)
```

PR #266, opened 2021-09-01
all the changes for the placement group

Kuldip-Nanda:
All the changes for the placement group.
Now placement group is available in advanced settings.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -53,6 +53,18 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
     images_by_id.values
   end
 
+  def placement_groups_by_id
+    @placement_groups_by_id ||= placement_groups_api.pcloud_placementgroups_getall(cloud_instance_id).placement_groups.index_by(&:id)
+  end
+
+  def placement_group(placement_group_id)
+    placement_groups_by_id[placement_group_id] ||= placement_groups_api.pcloud_placementgroups_get(cloud_instance_id, placement_group_id)
+  rescue IbmCloudPower::ApiError => err
+    error_message = JSON.parse(err.response_body)["description"]
+    _log.debug("Placement group with id #{placement_group_id} does not exist: #{error_message}")
+    nil
+  end
+
   def image_architecture(image_id)
     image = image(image_id)
     architecture = image&.specifications&.architecture
@@ -118,6 +130,10 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
 
   def pcloud_tenant_id
     cloud_manager.pcloud_tenant_id(connection)
+  end
+
+  def placement_groups_api
+    @placement_groups_api ||= IbmCloudPower::PCloudPlacementGroupsApi.new(connection)
   end
 
   def images_api
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -119,6 +119,18 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
         :read_only    => true
       )
 
+      # saving placement_group
+      placement_group = collector.placement_group(instance.placement_group)
+      value = "#{placement_group.name}[#{placement_group.policy}]" unless placement_group.nil?
+      persister.vms_and_templates_advanced_settings.build(
+        :resource     => ps_vmi,
+        :name         => 'placement_group',
+        :display_name => _('Placement Group'),
+        :description  => _('The placement group of the server'),
+        :value        => value,
+        :read_only    => true
+      )
+
       ldesc = software_licenses_description(instance.software_licenses)
       ldesc.present? && persister.vms_and_templates_advanced_settings.build(
         :resource     => ps_vmi,
--- a/content/reports/1000_IBM Power - Virtual Machines/010_PVMs_ Hardware.yaml
+++ b/content/reports/1000_IBM Power - Virtual Machines/010_PVMs_ Hardware.yaml
@@ -7,9 +7,9 @@ conditions: !ruby/object:MiqExpression
     INCLUDES:
       field: Vm.ext_management_system-emstype_description
       value: IBM Power
-  context_type: 
-  col_details: 
-  ruby: 
+  context_type:
+  col_details:
+  ruby:
 updated_on: 2021-07-28 18:04:00.000000 Z
 order: Ascending
 graph:
@@ -26,6 +26,7 @@ col_order:
 - hardware.cpu_total_cores
 - hardware.disks.device_type
 - processor_pin_policy
+- placement_group
 - hardware.num_disks
 - hardware.disks.size
 - ipaddresses
@@ -57,6 +58,7 @@ cols:
 - entitled_processors
 - processor_share_type
 - processor_pin_policy
+- placement_group
 - ipaddresses
 - mac_addresses
 template_type: report
@@ -71,6 +73,7 @@ headers:
 - vCPUs
 - Disk Tier
 - Pin Policy
+- Placement Group
 - "# of Disks"
 - Disk Size (GB)
 - IP Addresses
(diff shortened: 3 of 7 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -53,6 +53,18 @@ def images
...
 
+  def placement_groups_by_id
+    @placement_groups_by_id ||= placement_groups_api.pcloud_placementgroups_getall(cloud_instance_id).placement_groups.index_by(&:id)
+  end
+
+  def placement_group(placement_group_id)
+    placement_groups_by_id[placement_group_id] ||= placement_groups_api.pcloud_placementgroups_get(cloud_instance_id, placement_group_id)
+  rescue IbmCloudPower::ApiError => err
+    error_message = JSON.parse(err.response_body)["description"]
+    _log.debug("Placement Group ID not found in stock catalog: #{error_message}")
```
jaywcarman, line 64: The "stock catalog" is a specific to images, so this is a little confusing. Maybe just "Placement group ID not found"?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -31,6 +31,9 @@ def parse
   end
 
   def pvm_instances
+    # collect all the placement group names to be consumed by each instance
+    collector.placement_groups_by_id
```
jaywcarman, line 35: Is this necessary?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -119,6 +122,18 @@ def pvm_instances
         :read_only    => true
       )
 
+      # saving placement_group
+      placement_group = collector.placement_group(instance.placement_group)
+      value = "#{placement_group.name}[#{placement_group.policy}]" unless placement_group.nil?
```
jaywcarman, line 127: If `placement_group.nil?` does `value == nil`? or is it left unitialized?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -154,7 +169,9 @@ def images
   end
 
   def volumes
-    collector.volumes.each do |vol|
+    collector.volumes.each do |volume_ref|
```
jaywcarman, line 172: Seems like these changes to the volume parser are unintentional? I think you might need to rebase to include #257

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -16,16 +16,17 @@ def prepare_for_clone_task
...
-      specs['key_pair_name'] = chosen_key_pair unless chosen_key_pair == 'None'
-      specs['storage_type']  = get_option_last(:storage_type)
-      specs['sys_type']      = get_option_last(:sys_type)
+      specs['server_name']     = get_option(:vm_target_name)
+      specs['memory']          = get_option_last(:vm_memory).to_i
+      specs['migratable']      = get_option_last(:migratable) == 1
+      specs['processors']      = get_option_last(:entitled_processors).to_f
+      specs['proc_type']       = get_option_last(:instance_type)
+      specs['pin_policy']      = get_option_last(:pin_policy)
+      specs['placement_group'] = get_option_last(:placement_group)
```
jaywcarman, line 25: We'll need to update the provisioning form to include `placement_group` before we can add it to the tasks here. Let's save the provisioning changes for a later PR.

content/reports/1000_IBM Power - Virtual Machines/010_PVMs_ Hardware.yaml
```diff
@@ -26,6 +26,7 @@ col_order:
 - hardware.cpu_total_cores
 - hardware.disks.device_type
 - processor_pin_policy
+- placement_group
```
jaywcarman, line 29: 👍🏻 Thanks for updating the reports to include this!

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: Yes

jaywcarman: :) haha. Let me expand on my question.

On line 126 you call `placement_group = collector.placement_group(instance.placement_group)`, which will call the `collector.placement_groups_by_id`. Since `placement_groups_by_id` is memoized it won't cost any extra API calls, but I don't think this line is necessary. Is there something I'm missing?

Kuldip-Nanda: Sure Jay. Let me recheck the code segment after removing the line.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: not sure how it got through. I may rebase it again and see.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
Kuldip-Nanda: sure, I can remove ut from here

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: `value == nil`  will be true

jaywcarman: Great, thanks for the explanation... makes sense and I learned something new about Ruby :)

agrare: If `placement_group` is `nil` do you want to build the advanced setting at all?  Seems confusing to have the advanced setting with a value of `nil`

What about
```ruby
placement_group = collector.placement_group(instance.placement_group)
if placement_group
    persister.vms_and_templates_advanced_settings.build(
      :resource     => ps_vmi,
      :name         => 'placement_group',
      :display_name => _('Placement Group'),
      :description  => _('The placement group of the server'),
      :value        => "#{placement_group.name}[#{placement_group.policy}]",
      :read_only    => true
    )
end
```

jaywcarman: @agrare I thought adding an Placement group = nil entry into the advanced settings was a good idea (to differentiate between not collecting placement group vs. no placement is assigned to a VM). I was also thinking about reporting, but I'm not completely sure what the behavior is - could we filter a report of VMs without placement groups assigned in the same way that we could with placement groups = nil?

agrare: I just noticed that your other optional advanced setting didn't build if it was nil down on `+134` [[ref]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/266/files#diff-f985d1bd514a044797c4062d1d8d94ff3da33e97f6bfa8d182955661184e7d45R133)

agrare: But if you'd prefer to create it with a nil value that works too

jaywcarman: If both ways work (sounds like they do) then I don't feel strongly either. @Kuldip-Nanda you can make the call if you'd like since it's your patch.

Kuldip-Nanda: Let it remain Adam. This is a powerful feature. If it is not set it will show the customer something he can use.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
jaywcarman: @Kuldip-Nanda these changes to the volume parsing still need to be removed. After that and removing the `[WIP]` from the PR title I think this is ready to merge!

jaywcarman (approved):
LGTM

PR #267, opened 2021-09-02
parsing and displaying ansible fail message

iv1111:


iv1111:
Related issue was detected and solved my MIQ team: https://github.com/ManageIQ/manageiq/issues/21411

iv1111:
two more areas left for status propagation: python OVA image upload script and post ansible-playbook code (image transfer, cleanup code, etc)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -20,87 +51,119 @@ def post_execute
...
     ems = ExtManagementSystem.find(options[:ems_id])
 
     msg    = nil
     status = nil
     signal = nil
+    max_retries = 16
 
     ems.with_provider_connection(:service => 'PCloudTasksApi') do |api|
       response = api.pcloud_tasks_get(context[:task_id])
+      raise MiqException, 'HTTP response object is malformed, remote API error' if response&.status.nil?
```
agrare, line 89: This seems like an odd thing to handle unless you actually hit it, when would the API call not return _any_ http status?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -20,87 +51,119 @@ def post_execute
...
+                     JSON.parse(e.response_body)["description"] if e.response_body
+                   end
+                 end
+
+      messsage ||= e.message
+
+      retr = context[:retry].to_i
+      if retr < max_retries
+        msg = "failed to retrieve status for a cloud-task: '#{trunc_err_msg(messsage)}'; try number: #{retr + 1} of #{max_retries}"
+        signal = :post_execute_poll
```
agrare, line 121: If you're going to loop on a retry you probably want to add a queue delay so it doesn't just spin, e.g. add `:deliver_on => Time.now.utc + 1.minute` to the `queue_signal` method opts

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -20,87 +51,119 @@ def post_execute
...
     end
 
+    set_status(msg, status)
     queue_signal(signal, msg, status)
   end
 
   def post_poll_cleanup
     ems = ExtManagementSystem.find(options[:ems_id])
-    ems.remove_import_auth(options[:import_creds_id])
+    ems&.remove_import_auth(options[:import_creds_id])
```
agrare, line 139: `.find(id)` will raise a `RecordNotFound` exception if the record cannot be found so you don't have to worry about `nil` being returned.  That's if you use `.find_by(:id => id)`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -20,87 +51,119 @@ def post_execute
...
 
     return if options[:keep_ova] == true
+
     cos = ExtManagementSystem.find(options[:cos_id])
-    cos.remove_object(options[:bucket_name], options[:session_id] + '.ova')
+
+    begin
+      cos&.remove_object(options[:bucket_name], "#{options[:session_id]}.ova")
```
agrare, line 146: Same comment here

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: Since we don't control the remote site and can't be completely sure if the response is well formed I though to guard against nullptr here. However, this is probably a job a response validator from the SDK client gem. I will remove.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -20,87 +51,119 @@ def post_execute
...
+        msg = "importing image into image registry has failed with an error: '#{trunc_err_msg(response.status)}'"
+        signal = :abort
+        status = 'error'
+        _log.error(response.status)
+      else
+        raise MiqException, "unexpected status - '#{response.status}'"
+      end
+    rescue => e
+      messsage = if e.kind_of?(IbmCloudPower::ApiError)
+                   suppress(StandardError) do
```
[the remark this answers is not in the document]
iv1111: If it's an error coming from the API the idea is to try parse the error description as a JSON object locating the description field containing short error summary and if it's not possible then just output the message as is. So there is no more logic in that block other than trying to parse a JSON object.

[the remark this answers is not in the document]
iv1111: I will move the API error parsing logic to a separate method, should be a bit cleaner this way.

Issue #147, opened 2021-02-18
[Najdorf] IBM Cloud PowerVS Provider Enhancements


jaywcarman:
Power Virtual Servers and PowerVC provider features targeted for Morphy Release:

- [x] PowerVC -> PowerVS cross cloud-provider image import #180
- [x] New IBM Cloud Object Storage provider #159 
- [x] Add event handler (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/279)
- [x] Collect snapshots (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/313)
- [x] Snapshots
  - [x] Create snapshots https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/325
  - [x] Delete snapshots https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/327
  - [x] Revert to snapshot https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/331
  - [x] Calculate snapshot size https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/332

agrare:
Split out issues which are being backported to Morphy in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/269

jaywcarman:
Couple of updates to comment on:
- Remove `Enable chargeback rate assignments based on Power related attributes (e.g. entitled processors).` We'll defer this to a future release. There are still some reporting "enhancements" I'd like to make for existing reports (consistent names, tweaks, etc.).
- Add `VM affinity for multi-VM create`. PowerVS allow us to specify affinity/anti-affinity when creating VMs. We'd like to add this to the existing VM provision workflow.

PR #279, opened 2021-09-14
PowerVS event catcher and parser

jaywcarman:
I'm working to implement event catching/parsing/handling for the PowerVS cloud manager.

Depends on:
- [ ] ManageIQ/manageiq-content#690

https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/147

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
@@ -14,7 +14,6 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
   end
 
   def queue_event(event)
-    log_prefix = "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCatcher::Runner"
     _log.info("#{log_prefix} Caught event [#{event[:eventID]}]")
     event_hash = ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventParser.event_to_hash(event, @cfg[:ems_id])
     EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
@@ -14,30 +14,15 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
   end
 
   def queue_event(event)
-    _log.info("#{log_prefix} Caught event [#{event[:id]}]")
-    event_hash = event_to_hash(event, @cfg[:ems_id])
+    log_prefix = "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCatcher::Runner"
+    _log.info("#{log_prefix} Caught event [#{event[:eventID]}]")
+    event_hash = ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventParser.event_to_hash(event, @cfg[:ems_id])
     EmsEvent.add_queue('add', @cfg[:ems_id], event_hash)
   end
 
   private
 
-  def event_to_hash(event, ems_id)
-    {
-      :event_type => "IBM_CLOUD_VIRTUAL_SERVERS#{event[:name]}",
-      :source     => 'IBM_CLOUD_VIRTUAL_SERVERS',
-      :timestamp  => event[:timestamp],
-      :vm_ems_ref => event[:vm_ems_ref],
-      :full_data  => event,
-      :ems_id     => ems_id
-    }
-  end
-
   def event_monitor_handle
-    @event_monitor_handle ||= begin
-      self.class.parent::Stream.new(
-        @ems,
-        :poll_sleep => worker_settings[:poll]
-      )
-    end
+    @event_monitor_handle ||= self.class.parent::Stream.new(@ems)
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
@@ -18,17 +18,17 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
   def poll
     from_time = Time.now.utc.to_i
 
+    pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
     loop do
-      ems.with_provider_connection(:service => 'PCloudEventsApi') do |api|
-        # Could also use IBM Cloud Activity Tracker like the VPC provider
-        events = api.pcloud_events_getsince(ems.uid_ems, from_time).events
-        from_time = Time.now.utc.to_i
-
-        break if stop_polling
-
-        events.each { |event| yield event.to_hash }
+      begin
+        events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
+      # TODO: Only rescue token experation exception
+      rescue Exception => e
+        pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
       end
-
+      from_time = Time.now.utc.to_i
+      events.each { |event| yield event.to_hash }
+      break if stop_polling
       sleep(poll_sleep)
     end
   end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
@@ -4,7 +4,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
   def initialize(ems, options = {})
     @ems = ems
     @stop_polling = false
-    @poll_sleep = options[:poll_sleep] || 20.seconds
+    @poll_sleep = options[:poll_sleep] || 30.seconds
   end
 
   def start
@@ -25,10 +25,11 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
       events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
 
       from_time = Time.now.utc.to_i
-      events.each { |event| yield event.to_hash }
-      break if stop_polling
 
       sleep(poll_sleep)
+
+      events.each { |event| yield event.to_hash }
+      break if stop_polling
     rescue IbmCloudPower::ApiError => e
       raise unless e.code == 403 && retry_connection
 
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
@@ -18,18 +18,22 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
   def poll
     from_time = Time.now.utc.to_i
 
-    pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
     loop do
-      begin
-        events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
-      # TODO: Only rescue token experation exception
-      rescue Exception => e
-        pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
-      end
+      pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
+
+      retry_connection = true
+      events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
+
       from_time = Time.now.utc.to_i
       events.each { |event| yield event.to_hash }
       break if stop_polling
+
       sleep(poll_sleep)
+    rescue IbmCloudPower::ApiError => e
+      raise unless e.code == 403 && retry_connection
+
+      retry_connection = false
+      retry
     end
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
@@ -1,10 +1,8 @@
 class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCatcher::Stream
-  class ProviderUnreachable < ManageIQ::Providers::BaseManager::EventCatcher::Runner::TemporaryFailure
-  end
+  attr_reader :ems, :stop_polling, :poll_sleep
 
   def initialize(ems, options = {})
     @ems = ems
-    @last_activity = nil
     @stop_polling = false
     @poll_sleep = options[:poll_sleep] || 20.seconds
   end
@@ -17,32 +15,21 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
     @stop_polling = true
   end
 
-  def fake_events
-    [
-      OpenStruct.new(:name       => %w[instance_power_on instance_power_off].sample,
-                     :id         => Time.zone.now.to_i,
-                     :timestamp  => Time.zone.now,
-                     :vm_ems_ref => [1, 2].sample),
-      OpenStruct.new(:name       => %w[instance_power_on instance_power_off].sample,
-                     :id         => Time.zone.now.to_i + 1,
-                     :timestamp  => Time.zone.now,
-                     :vm_ems_ref => [1, 2].sample),
-    ]
-  end
-
   def poll
-    @ems.with_provider_connection do |_connection|
-      catch(:stop_polling) do
-        loop do
-          fake_events.each do |activity|
-            throw :stop_polling if @stop_polling
-            yield activity.to_h
-          end
-          sleep @poll_sleep
-        end
-      rescue => exception
-        raise ProviderUnreachable, exception.message
+    from_time = Time.now.utc.to_i
+
+    loop do
+      ems.with_provider_connection(:service => "PCloudEventsApi") do |api|
+        # Could also use IBM Cloud Activity Tracker like the VPC provider
+        events = api.pcloud_events_getsince(ems.uid_ems, from_time).events
+        from_time = Time.now.utc.to_i
+
+        break if stop_polling
+
+        events.each { |event| yield event.to_hash }
       end
+
+      sleep(poll_sleep)
     end
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
@@ -19,7 +19,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCat
     from_time = Time.now.utc.to_i
 
     loop do
-      ems.with_provider_connection(:service => "PCloudEventsApi") do |api|
+      ems.with_provider_connection(:service => 'PCloudEventsApi') do |api|
         # Could also use IBM Cloud Activity Tracker like the VPC provider
         events = api.pcloud_events_getsince(ems.uid_ems, from_time).events
         from_time = Time.now.utc.to_i
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
@@ -1,6 +1,5 @@
 module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventParser
   def self.event_to_hash(event, ems_id)
-    ems = ExtManagementSystem.find(ems_id)
     event_hash = {
       :event_type => "#{event[:resource]}.#{event[:action]}",
       :source     => "IBMCloud-PowerVS",
@@ -22,6 +21,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventPa
 
   def self.parse_vm_event!(event, event_hash)
     pvm_instance_name = event[:message].split('\'')[1]
+    ems = ExtManagementSystem.find(event_hash[:ems_id])
     vm = ems.vms.find_by(:name => pvm_instance_name)
     return if vm.nil?
 
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
@@ -1,5 +1,6 @@
 module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventParser
   def self.event_to_hash(event, ems_id)
+    ems = ExtManagementSystem.find(ems_id)
     event_hash = {
       :event_type => "#{event[:resource]}.#{event[:action]}",
       :source     => "IBMCloud-PowerVS",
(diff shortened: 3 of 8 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
```diff
@@ -17,32 +15,27 @@ def stop
...
-          sleep @poll_sleep
-        end
-      rescue => exception
-        raise ProviderUnreachable, exception.message
+    from_time = Time.now.utc.to_i
+
+    loop do
+      ems.with_provider_connection(:service => "PCloudEventsApi") do |api|
+        # Could also use IBM Cloud Activity Tracker like the VPC provider
+        events = api.pcloud_events_getsince(cloud_instance_id, from_time).events
```
jaywcarman, line 24: Link to the PowerVS Events API:
https://cloud.ibm.com/apidocs/power-cloud#pcloud-events-getsince

Unfortunately, in my testing I'm always getting an empty `metadata` object, which is supposed to include the VM instance ID. I'm following up with the API devs. If this doesn't work the options would be to parse the VM instance name from the `message` string, or use the IBM Cloud Activity Tracker API like VPC does.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
```diff
@@ -14,30 +14,15 @@ def monitor_events
   end
 
   def queue_event(event)
-    _log.info("#{log_prefix} Caught event [#{event[:id]}]")
-    event_hash = event_to_hash(event, @cfg[:ems_id])
+    log_prefix = "ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventCatcher::Runner"
```
agrare, line 17: `log_prefix` is a common method in the base class so you don't need to override, also if you just want to print the class name that is already part of the `_log` formatter.

Here is an example from the vmware event catcher:
`_log.info("#{log_prefix} Starting Event Monitor Thread")`
yields:
`INFO -- evm: MIQ(ManageIQ::Providers::Vmware::InfraManager::EventCatcher::Runner#start_event_monitor) EMS [vc.rb.nj.grare.com] as [root] Starting Event Monitor Thread`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
```diff
@@ -17,32 +15,21 @@ def stop
...
-        raise ProviderUnreachable, exception.message
+    from_time = Time.now.utc.to_i
+
+    loop do
+      ems.with_provider_connection(:service => 'PCloudEventsApi') do |api|
+        # Could also use IBM Cloud Activity Tracker like the VPC provider
+        events = api.pcloud_events_getsince(ems.uid_ems, from_time).events
+        from_time = Time.now.utc.to_i
+
+        break if stop_polling
```
agrare, line 27: I think you need this break outside of the `with_provider_connection` block for this to exit cleanly

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
```diff
@@ -17,32 +15,21 @@ def stop
...
-            yield activity.to_h
-          end
-          sleep @poll_sleep
-        end
-      rescue => exception
-        raise ProviderUnreachable, exception.message
+    from_time = Time.now.utc.to_i
+
+    loop do
+      ems.with_provider_connection(:service => 'PCloudEventsApi') do |api|
```
agrare, line 22: Might be more efficient to setup a connection outside of the loop rather than re-connect every time:
```
pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
loop do
  events = pcloud_events_api.....
end
```

You might need to rescue a stale bearer token and retry but it should reduce the number of API calls to the IAM service

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,32 @@
...
+    when /^pvm-instance/
+      parse_vm_event!(event, event_hash)
+    end
+
+    event_hash
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    pvm_instance_name = event[:message].split('\'')[1]
+    vm = Vm.find_by(:name => pvm_instance_name)
```
agrare, line 23: VM name is not unique globally, should be `ext_management_system.{vms/vms_and_templates}.find_by(:name => pvm_instance_name)`

You can do the `ems = ExtManagementSystem.find(ems_id)` up in `self.event_to_hash` and pass it down

agrare:
Looks awesome @jaywcarman just a couple of comments

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
agrare: Also, the event catcher will finish processing events that are added to its internal queue before stopping (to prevent "in-flight" loss of events) so you probably want to yield the events to the caller then check for `stop_polling`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
jaywcarman: Decided to go with parsing out the `message` strings.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/runner.rb
jaywcarman: Thanks! No need to override so I removed it in 2be2afa

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
jaywcarman: LIke this 04bb108 ?

If so then should the VPC `EventCatcher::Stream` also switch to yield before `break if stop_polling`? https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/event_catcher/stream.rb#L35-L37

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
jaywcarman: I refactored per your suggestion, but still need to lookup the correct Exception to rescue for a stale token (and test).

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
jaywcarman: Do you mean include it in the `event_hash` hash? I tried adding an `:ems` key with the value set to the appropriate ems object, but got the following error:
```
[----] I, [2021-11-11T13:50:48.932606 #3797567:f33964]  INFO -- evm: MIQ(MiqQueue#deliver) Message id: [1626], Delivering...                                                                                                                                  
[----] E, [2021-11-11T13:50:50.845219 #3797567:f33964] ERROR -- evm: MIQ(MiqQueue#deliver) Message id: [1626], Error: [unknown attribute 'ems' for EmsEvent.]                                                                                                 
[----] E, [2021-11-11T13:50:50.845390 #3797567:f33964] ERROR -- evm: [ActiveModel::UnknownAttributeError]: unknown attribute 'ems' for EmsEvent.  Method:[block (2 levels) in <class:LogProxy>]                                                               
[----] E, [2021-11-11T13:50:50.845554 #3797567:f33964] ERROR -- evm: /home/jwcarman/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0/gems/activemodel-6.0.4.1/lib/active_model/attribute_assignment.rb:52:in `_assign_attribute'
```

Instead I grabbed the `:ems_id` from the `event_hash` and add the lookup to `self.parse_vm_event`. This works, but will have to run the query every time in parses a vm event.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
agrare: Yes, and yes :wink:

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
agrare: My guess is `IbmCloudPower::ApiError` with `err.code => 401` but yeah, good to test it to confirm

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
agrare: No sorry, suggesting something like this:
```
def self.event_to_hash(event, ems_id)
  ems = ExtManagementSystem.find(ems_id)
  ...
      parse_vm_event!(ems, event, event_hash)
end

def self.parse_vm_event!(ems, event, event_hash)
  ems.vms.find_by(:name => pvm_instance_name)
```

What you have https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/279/files#diff-94f8045f8f587fbabedbe1ade93e3282f34b17895fd96a490f4a095aa5d697edR24 works too but if we end up having more `parse_{resource}_event` methods we might have to do that query in each one

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
jaywcarman: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/304

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_catcher/stream.rb
```diff
@@ -17,32 +15,21 @@ def stop
...
+      begin
+        events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
+      # TODO: Only rescue token experation exception
+      rescue Exception => e
+        pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
       end
+      from_time = Time.now.utc.to_i
+      events.each { |event| yield event.to_hash }
+      break if stop_polling
+      sleep(poll_sleep)
```
agrare, line 32: You can use `retry` and it'll go back to the top of the current block, should shave you having to call ems.connect in two different places.

Try:
```suggestion
    loop do
      pcloud_events_api = ems.connect(:service => "PCloudEventsApi")
      
      retry_connection = true

      events = pcloud_events_api.pcloud_events_getsince(@ems.uid_ems, from_time).events
      
      from_time = Time.now.utc.to_i
      events.each { |event| yield event.to_hash }
      break if stop_polling
      sleep(poll_sleep)
    rescue IbmCloudPower::ApiError => e
      raise unless e.code == 403 && retry_connection
      retry_connection = false
      retry
    end
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+    ems = ExtManagementSystem.find(event_hash[:ems_id])
+    vm = ems.vms.find_by(:name => pvm_instance_name)
+
+    if vm.nil? && ['update', 'create'].include?(event['type'])
+
+      # TODO: If event type is 'pvm-instance.update' and the VM name was
+      #       changed then only the new name is given in the event message :(
+      #       Should we refresh the entire EMS on this event? Same question
+      #       for a 'pvm-instance.create'.
+      event_hash
```
jaywcarman, line 33: I'm still trying to figure out how to queue up a full EMS refresh here.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      return
+    end
+
+    sleep(60)
```
jaywcarman, line 38: I hate to add calls to `sleep`... but the PowerVS back-end can be kind of slow. When testing MIQ would usually refresh the VM before the state had changed. I'm open to suggestions for a more elegant solution.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
agrare: You could handle this in your `EventTargetParser`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
agrare: Interesting so they raise an event about a state change before a `GET` would return the new state?  Is this like `pvm.instance.start` event and the state goes from stopped to starting but not running yet, or does it still return stopped?

jaywcarman: > Interesting so they raise an event about a state change before a `GET` would return the new state?

Correct. The event seems to be raised as soon as a state change request is made.

Here's an example using cli tools:
```
rdr-miq-mon01 > pcloud GET /pcloud/v1/cloud-instances/${cloud_instance_id}/events 'from_time=1638199159' | jq .
{
  "events": []
}
rdr-miq-mon01 > ic pi in c2c2a435-2cb9-4cfd-b5b3-3345dc01201b --json | jq .status                              
"ACTIVE"
rdr-miq-mon01 > ic pi instance-stop c2c2a435-2cb9-4cfd-b5b3-3345dc01201b         
Performing action stop for instance c2c2a435-2cb9-4cfd-b5b3-3345dc01201b under account IBM Power Cloud Development as user jwcarman@us.ibm.com...
OK
Action stop complete for instance c2c2a435-2cb9-4cfd-b5b3-3345dc01201b.
rdr-miq-mon01 > pcloud GET /pcloud/v1/cloud-instances/${cloud_instance_id}/events 'from_time=1638199159' | jq .
{
  "events": [
    {
      "action": "stop",
      "eventID": "80bf0d39-0ab5-4e30-af2c-9a18d33bbc46",
      "level": "info",
      "message": "Power virtual server instance 'rdr-rhacm-build-newname' stop action successfully completed.",
      "metadata": {},
      "resource": "pvm-instance",
      "time": "2021-11-29T15:20:52.777Z",
      "timestamp": 1638199252,
      "user": {
        "email": "jwcarman@us.ibm.com",
        "name": "Unknown Name",
        "userID": "IBMid-2700063YVN"
      }
    }
  ]
}
rdr-miq-mon01 > ic pi in c2c2a435-2cb9-4cfd-b5b3-3345dc01201b --json | jq .status                   
"ACTIVE"
rdr-miq-mon01 > sleep 60 && ic pi in c2c2a435-2cb9-4cfd-b5b3-3345dc01201b --json | jq .status
"SHUTOFF"
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,34 @@
...
+  def parse
+    target_collection = InventoryRefresh::TargetCollection.new(
+      :manager => ems_event.ext_management_system,
+      :event   => ems_event
+    )
+
+    case ems_event[:event_type]
+    when /^pvm-instance\.create/
+      target_collection.add_target(
+        :association => :ext_management_system,
```
jaywcarman, line 17: @agrare I'm still struggling to target the entire ems (haven't been able to find a similar situation elsewhere in MIQ).

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
```diff
@@ -0,0 +1,34 @@
...
+        :manager_ref => {:ems_ref => ems_event.ext_management_system}
+      )
+    when /^pvm-instance\.update/
+      target_collection.add_target(
+        :association => :ext_management_system,
```
jaywcarman, line 22: Is there a way to target all existing VMs?

@vfebvre has a similar question [here](https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/22/files#diff-fe5f4bef291f246bd6808eda3d3c8d1d8eceb696015b491ff70d1f7784c66a65R43).

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
agrare: That is unfortunate, really don't want to slow down the event catcher at all.  If we have to sleep somewhere I think this isn't the place, if someone powers on 10 vms in short order it'll be an 10 minutes before the last one is even queued for the event handler.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,41 @@
...
+    event_hash
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    pvm_instance_name = event[:message].split('\'')[1]
+    ems = ExtManagementSystem.find(event_hash[:ems_id])
+    vm = ems.vms.find_by(:name => pvm_instance_name)
+
+    if vm.nil?
+      if ['update', 'create'].include?(event['type'])
```
agrare, line 28: Could we do this check before doing the query to find the vm?  The event catcher side is really intended to not run any DB queries to maintain high event throughput, that is why all of the queries are done by the event_handler once the events have been caught.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,41 @@
...
+      parse_vm_event!(event, event_hash)
+    end
+
+    event_hash
+  end
+
+  def self.parse_vm_event!(event, event_hash)
+    pvm_instance_name = event[:message].split('\'')[1]
+    ems = ExtManagementSystem.find(event_hash[:ems_id])
+    vm = ems.vms.find_by(:name => pvm_instance_name)
```
agrare, line 25: Also you could turn this into one query like `VmOrTemplate.find_by(:ems_id => event_hash[:ems_id], :name => pvm_instance_name)`

(find_by assumes the name is unique within a provider instance)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
agrare: Yeah I don't believe any other `EventTargetParser`s do this let me play around a bit and let you know how you'd go about doing this

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
agrare: I think the way to go here would be to give a target that is just the `:association => :vms` with no `manager_ref`.

I don't know if this is possible at the `InventoryRefresh::Target` level, but assuming it is you could handle this case in your targeted_collector.  We do set the targeted_scope based on the manager_refresh_targets so you might have to be careful with that.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
jaywcarman: I think we can use the sleep in the `EventCatcher::Stream` to force a delay (and allow it to be see via settings): f1bda4d

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
jaywcarman: Thanks @agrare - implemented here: 572010c

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
jaywcarman: For now I targeted the entire EMS (572010c), but in the future would be nice to only refresh VMs (or better yet, if the PowerVS event API will improve to allow us to target a single VM!)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
jaywcarman: [done](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/279/commits/572010c76a2f29bf7681667c5bad8bdb1fb28b85#diff-94f8045f8f587fbabedbe1ade93e3282f34b17895fd96a490f4a095aa5d697edR28-R31)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
jaywcarman: [done](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/279/commits/572010c76a2f29bf7681667c5bad8bdb1fb28b85#diff-94f8045f8f587fbabedbe1ade93e3282f34b17895fd96a490f4a095aa5d697edR23)

jaywcarman:
> **app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb**
> 
> * [ ]  warning - [Line 14](https://github.com/jaywcarman/manageiq-providers-ibm_cloud/blob/f1bda4d5a7f58298265a48496bff755f7a822b5c/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb#L14), Col 5 - [Lint/DuplicateBranch](https://docs.rubocop.org/rubocop/1.13/cops_lint.html#lintduplicatebranch) - Duplicate branch body detected.

This is on purpose, with the intention of refactoring the `when /^pvm-instance\.update/` branch in the future to do a more targeted refresh scope.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,42 @@
+module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventParser
+  def self.event_to_hash(event, ems_id)
+    event_hash = {
+      :event_type => "#{event[:resource]}.#{event[:action]}",
+      :source     => "IBMCloud-PowerVS",
+      :ems_id     => ems_id,
+      :ems_ref    => event[:eventID],
+      :timestamp  => event[:time],
+      :full_data  => event,
+      :username   => event[:user][:email]
```
agrare, line 10: Not sure if `event[:user]` can ever be nil, but if it can this will fail

`event.dig(:user, :email)` is a good way to handle this condition

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+  def self.parse_vm_event!(event, event_hash)
+    if ['update', 'create'].include?(event['type'])
+      event_hash
+    else
+      pvm_instance_name = event[:message].split('\'')[1]
+
+      vm = VmOrTemplate.find_by(
+        :ems_id => event_hash[:ems_id],
+        :name   => pvm_instance_name
+      )
```
agrare, line 31: Might want to add a comment here indicating that name is unique within a powervs provider instance because usually find by name would not work

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      if vm.nil?
+        return
+      end
```
agrare, line 35: Just a style thing but usually this is written as `return if vm.nil?`

agrare (approved):
Looks good to me just really minor things that we can do in a follow-up

PR #286, opened 2021-09-24
VPC Rake task for refresh setup

nasark:
- Added rake task to initialize environment and generate VCRs to be used by `refresher_spec`
- changes were made to `refresher_spec` in order to accommodate for the task

Steps:
1. Create resources
2. Generate VCRs
3. Cleanup resources

Prerequisites (included in comments):
- Log in with IBM Cloud CLI i.e. `ibmcloud login -sso`
- Place IBM Cloud API key in secrets config

Notes:
- Improvements are greatly appreciated
- Cloud databases can take 5-10 minutes to provision


nasark:
@miq-bot add-labels technical debt

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
+namespace :spec do
+  require 'ibm_cloud_sdk_core'
+  require 'ibm_cloud_resource_controller'
+  require 'ibm_vpc'
+
+  base_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.config.root.to_s, 'spec')
```
agrare, line 6: You can leave this as a Pathname which makes it a little cleaner to build the paths later:

```
base_dir = ManageIQ::Providers::IbmCloud::Engine.root.join("spec")
```

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+  cass_dir = File.join(base_dir, 'vcr_cassettes/manageiq/providers/ibm_cloud')
```
agrare, line 7: Then this can be
```
cass_dir = base_dir.join("vcr_cassettes/manageiq/providers/ibm_cloud")
```

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+  spec_dir = File.join(base_dir, 'models/manageiq/providers/ibm_cloud')
+
+  # Prerequisites:
+  # 1. Log in with IBM Cloud CLI i.e. `ibmcloud login -sso`
+  # 2. Place IBM Cloud API key in secrets config
+  desc "Generate refresher VCR"
+  task :refresh_init => :environment do
+    Dir["#{cass_dir}/vpc/cloud_manager/refresher*.yml"].each do |file|
```
agrare, line 15: `Pathname` will let you `glob()` and return an array of pathnames so you can:
```
cass_dir.glob("vpc/cloud_manager/**/*.yml").each(&:delete)
```

(NOTE `&:method` passed to an enumerator will call the method on each item in the enumeration so this is the same as saying `.each { |file| file.delete }`

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+      FileUtils.rm(file, :verbose => true)
+    end
+    Dir["#{cass_dir}/vpc/cloud_manager/refresher/vm_target.yml"].each do |file|
+      FileUtils.rm(file, :verbose => true)
+    end
+
+    apikey = YAML.load_file("config/secrets.yml")["test"]["ibm_cloud_vpc"]["api_key"]
```
agrare, line 22: Maybe `.dig("test", "ibm_cloud_vpc", "api_key")` and raise a helpful error if it is `nil`? (if a key in the path is missing this will just raise a `NoMethodError` exception

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+    token = IBMCloudSdkCore::IamAuthenticator.new(:apikey => apikey)
+                                             .instance_variable_get(:@token_manager).access_token
```
agrare, line 24: Does creating an `IamAuthenticator` to get the access_token buy us much over going to the `IAMTokenManager` directly?  I think this could be done with `IBMCloudSdkCore::IAMTokenManager.new(:apikey => apikey).access_token`

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+    auth = IBMCloudSdkCore::BearerTokenAuthenticator.new(:bearer_token => token)
+
+    connection = IbmVpc::VpcV1.new({:authenticator => auth})
```
agrare, line 27: If the only argument is a hash you don't need to put `{}` around it, just `IbmVpc::VpcV1.new(:authenticator => auth)` should work.

You really only need to wrap the hash arg in `{}` if there are two separate hash-type arguments and you need to indicate which one is which like: `def foo(opts = {}, more_opts = {})` for that you'd need to call it like `foo({:a => "a"}, {:b => "b"})`

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+      default_resource_group_id = `ibmcloud resource group default --id | tr -d '\n'`
+      cloud_db_id = resource_controller.create_resource_instance(
+        :name             => 'rake-db',
+        :target           => 'bluemix-us-south',
+        :resource_group   => default_resource_group_id,
+        :resource_plan_id => 'databases-for-postgresql-standard'
+      ).result['id']
+      status = resource_controller.get_resource_instance(:id => cloud_db_id).result['state'] until status == 'active'
+    rescue => err
+      puts err
```
agrare, line 41: So if this create fails this will continue on, is that intentional (aka ignore any failures here)?

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+      :zone     => {
+        :name => 'us-south-1'
+      },
+      :name     => 'rake-vol',
+      :capacity => 10
+    }
+    volume_id = connection.create_volume(:volume_prototype => volume_prototype).result['id']
+
+    `ssh-keygen -f temp -t rsa -P ""`
+    ssh_key = `cat temp.pub`
```
agrare, line 69: Backticks are convenient for running complex commands but this is easy enough to do in ruby with `File.read("temp.pub")`

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+    # Create resources
+    begin
+      default_resource_group_id = `ibmcloud resource group default --id | tr -d '\n'`
+      cloud_db_id = resource_controller.create_resource_instance(
+        :name             => 'rake-db',
+        :target           => 'bluemix-us-south',
+        :resource_group   => default_resource_group_id,
+        :resource_plan_id => 'databases-for-postgresql-standard'
+      ).result['id']
+      status = resource_controller.get_resource_instance(:id => cloud_db_id).result['state'] until status == 'active'
```
agrare, line 39: This is really slick but we probably want to incorporate some sleep into the loop so it doesn't "spin" on the api-call

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+    instance_id = connection.create_instance(:instance_prototype => instance_prototype).result['id']
+    status = connection.get_instance(:id => instance_id).result['status'] until status == 'running'
+
+    network_interface_id = connection.list_instance_network_interfaces(:instance_id => instance_id).result["network_interfaces"][0]["id"]
+    floating_ip_prototype = {:name => 'rake-floating-ip', :target => {:id => network_interface_id}}
+    floating_ip_id = connection.create_floating_ip(:floating_ip_prototype => floating_ip_prototype).result['id']
+    status = connection.get_floating_ip(:id => floating_ip_id).result['status'] until status == 'available'
+
+    # Generate VCRs
+    spec_file = Dir["#{spec_dir}/**/refresher_spec.rb"].first
```
agrare, line 122: Do you want to only run the `vpc` specs (since this is vpc.rake)?  Maybe `spec_dir` should be `base_dir.join("models/manageiq/providers/ibm_cloud/vpc")` ?

agrare: Oh you're just taking the first anyway, could you simply call out the full path to the `vpc/cloud_manager/refresher_spec.rb` then instead of the `Dir[].first`?

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+    `bundle exec rspec #{spec_file} --tag full_refresh`
+
+    instance_prototype[:name] = 'rake-instance2'
+    instance_prototype[:boot_volume_attachment][:volume][:name] = 'rake-instance2-boot'
+    target_instance_id = connection.create_instance(:instance_prototype => instance_prototype).result['id']
+    status = connection.get_instance(:id => target_instance_id).result['status'] until status == 'running'
+
+    `bundle exec rspec #{spec_file} --tag target_vm`
+
+    # Cleanup resources
```
agrare, line 132: It'd probably be a good idea to put the cleanup in an `ensure` so that if any of these raise an exception we don't leave resources hanging around.  Just have to deal nicely with resources not being created yet.

config/secrets.defaults.yml
```diff
@@ -6,6 +6,7 @@ test:
     <<: *ibm_cloud_power_defaults
   ibm_cloud_vpc_defaults: &ibm_cloud_vpc_defaults
     api_key: IBM_CLOUD_VPC_API_KEY
+    provider_region: provider-region
```
agrare, line 9: Does this have to be one of the regions that we declare in the `REGIONS` list?  I think this might fail on the regions validation

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,168 @@
...
+
+  base_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.config.root.to_s, 'spec')
+  cass_dir = File.join(base_dir, 'vcr_cassettes/manageiq/providers/ibm_cloud')
+  spec_dir = File.join(base_dir, 'models/manageiq/providers/ibm_cloud')
+
+  # Prerequisites:
+  # 1. Log in with IBM Cloud CLI i.e. `ibmcloud login -sso`
+  # 2. Place IBM Cloud API key in secrets config
+  desc "Generate refresher VCR"
+  task :refresh_init => :environment do
```
Fryguy, line 14: @agrare Do we have a naming convention for the task that generates a new env?

agrare: Azure and openstack both name these differently so we should stick with one and update the rest
`Azure: rake spec:regenerate:all`
`Openstack: rake vcr:rerecord`
Kubernetes just has a shell script

lib/tasks_private/vpc.rake
nasark: Sounds good, I'll incorporate a `sleep` into the the other API calls as well

config/secrets.defaults.yml
nasark: Yes this would have to be one of the regions in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/db/fixtures/ibm_cloud_vpc_regions.yml with the value of the `:name` key

nasark:
@agrare Thanks for the feedback, I've made the changes, let me know if there is anything else you would like to see

lib/tasks_private/vpc.rake
```diff
@@ -0,0 +1,181 @@
...
+        :resource_group   => default_resource_group_id,
+        :resource_plan_id => 'databases-for-postgresql-standard'
+      ).result['id']
+      status = resource_controller.get_resource_instance(:id => cloud_db_id).result['state']
+      puts "Provisioning cloud database..."
+      until status == 'active'
+        sleep(10)
+        status = resource_controller.get_resource_instance(:id => cloud_db_id).result['state']
+      end
+      puts "Provisioned cloud database"
```
nasark, line 43: I've added these helpful put statements here since it can take a while to provision a cloud database

config/secrets.defaults.yml
nasark: I haven't had this fail so far, where does the regions validation take place?

agrare: Oh interesting it looks like there isn't a provider_region validation on VPC, that's a different issue though.

Most providers have `validates :provider_region, :inclusion => {:in => ->(_region) { ManageIQ::Providers::Amazon::Regions.names }}` which would fail the `create` if the region isn't in the list of "allowed" regions

agrare: That's okay we can come back to this later

lib/tasks_private/vpc.rake
agrare: Don't like `spec:regenerate:all` because we aren't regenerating all specs haha  `vcr:rerecord` works unless @Fryguy has a better idea

lib/tasks_private/vpc.rake
nasark: I've decided to `raise` here instead since if we were to continue https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/refresher_spec.rb#L29 would fail anyway

nasark:
@agrare added VCRs

agrare (approved):

Issue #263, opened 2021-08-25
[Najdorf] IBM Cloud VPC Provider Enhancements


chessbyte:
- [x] Add ems_ref to auth key pairs (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/178)
- [x] Modify/resize VM (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/300)
- [x] Public gateways
- [x] VPN gateways (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/298)
- [x] ACLs (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/297)
- [x] Load balancers
- [x] Metrics Collection support (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/308)

---

- [x] Provision VM from service (catalog item) (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/362)
- [ ] Cloud Database Operations
  - [x] Create https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/366
  - [ ] Update
  - [x] Delete https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/364

PR #300, opened 2021-11-03
Add resize VM functionality

nasark:
- added the ability to resize/reconfigure VMs

Restrictions when resizing VMs from IBM Cloud:
- VM must be in a `stopped` or `stopping` state
- new flavor must be of the same profile family/architecture as old flavor
- unable to resize to a flavor with a disk if current flavor doesn't have a disk

If the user selects a flavor that doesn't meet the above requirements then an error is thrown and displayed via notifications, indicating the incompatibility of the new flavor selected

Depends on:
* [x] [Add error notification for resizing VMs](https://github.com/ManageIQ/manageiq/pull/21543)

Ref:
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/263

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
@@ -182,12 +182,13 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::VPC < ManageIQ::Provider
   def flavors
     collector.flavors.each do |flavor|
       memory = flavor&.dig(:memory, :value)
-      memory_mb = Integer(memory) * 1024 if memory
+      disk = flavor[:disks].first&.dig(:size, :value) || 0
       persister.flavors.build(
         :ems_ref         => flavor[:name],
         :name            => flavor[:name],
         :cpu_total_cores => flavor&.dig(:vcpu_count, :value),
-        :memory          => memory_mb,
+        :memory          => memory&.gigabytes,
+        :root_disk_size  => disk&.gigabytes,
         :enabled         => true
       )
     end
--- a/lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance.rb
+++ b/lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance.rb
@@ -72,6 +72,15 @@ module ManageIQ
               @parent.request(:delete_instance, :id => id)
             end
 
+            def resize(instance, new_flavor_name)
+              @parent.request(:update_instance,
+                              :id             => instance[:id],
+                              :instance_patch => {
+                                :name    => instance[:name],
+                                :profile => {:name => new_flavor_name}
+                              })
+            end
+
             # Wait for the VM instance to be have a started status.
             # @param sleep_time [Integer] The time to sleep between refreshes.
             # @param timeout [Integer] The number of seconds before raising an error.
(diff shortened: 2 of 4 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -182,12 +182,13 @@ def vm_and_template_taggings(resource, tags_inventory_objects)
...
       memory = flavor&.dig(:memory, :value)
-      memory_mb = Integer(memory) * 1024 if memory
+      disk = flavor[:disks].first&.dig(:size, :value) || 0
       persister.flavors.build(
         :ems_ref         => flavor[:name],
         :name            => flavor[:name],
         :cpu_total_cores => flavor&.dig(:vcpu_count, :value),
-        :memory          => memory_mb,
+        :memory          => memory&.gigabytes,
+        :root_disk_size  => disk&.gigabytes,
```
nasark, line 191: needed to add this attribute in order to make the form work due to https://github.com/ManageIQ/manageiq-ui-classic/blob/5e3a5d7858224498f8fa09a01aaf268a4c00693c/app/controllers/mixins/actions/vm_actions/resize.rb#L119

lib/manageiq/providers/ibm_cloud/cloud_tools/vpc_sdk/instance.rb
```diff
@@ -72,6 +72,19 @@ def delete
...
+              @parent.request(:update_instance,
+                              :id             => instance[:id],
+                              :instance_patch => {
+                                :name    => instance[:name],
+                                :profile => {:name => new_flavor_name}
+                              }
+                             )
+            rescue IBMCloudSdkCore::ApiException => e
+              Notification.create(:type => :vm_resize_error, :options => {:message => e.to_s})
+              raise MiqException::MiqProvisionError, e.to_s
```
agrare, line 85: I think this exception handler would be better to live up [[here]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/300/files#diff-5af23959f8142218f0266befea8e232991173bc7a177a5b6a24154a782cc2c2eR117) so this stays more of a pure IBM Cloud interface and keep the MIQ bits local to `app/models`

app/models/manageiq/providers/ibm_cloud/inventory/parser/vpc.rb
```diff
@@ -182,12 +182,13 @@ def vm_and_template_taggings(resource, tags_inventory_objects)
   def flavors
     collector.flavors.each do |flavor|
       memory = flavor&.dig(:memory, :value)
-      memory_mb = Integer(memory) * 1024 if memory
+      disk = flavor[:disks].first&.dig(:size, :value) || 0
```
agrare, line 185: Is there any guarantee that the first disk is the root disk?

nasark: `.first` here is due to the way IBM Cloud has structured their response. `flavor[:disks]` is an array of hashes however, there is only ever a single hash in this array. Flavors that use multiple disks as returned by the API, only use disks that are of the same size and as a result when there are multiple disks it is specified in the response like this:
```
"disks"=>[{
  "quantity"=>{"type"=>"fixed", "value"=>2},
  "size"=>{"type"=>"fixed", "value"=>2640},
  "supported_interface_types"=>{"default"=>"virtio_blk", "type"=>"enum", "values"=>["virtio_blk"]}
  }]
```
I couldn't find a way to determine which disk is the root but `:root_disk_size` would be set accordingly since the disks are of the same size

agrare: Ah okay so even if there are multiple disks there's still only one array entry, got it thanks

agrare (approved):

PR #301, opened 2021-11-08
[WIP] Add support for catalog provisioning VPC instances

agrare:


app/models/aliases/miq_provision_ibm_cloud_vpc_workflow.rb
```diff
@@ -0,0 +1 @@
+::MiqProvisionIbmCloudVpcWorkflow = ::ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow
```
agrare, line 1: The UI calls `MiqProvisionWorkflow.class_from_platform` which does `.classify` on the catalog type.

Fryguy: interesting...might be worth it to start killing that (not for this PR)

agrare: But it also checks that `klass.name.to_s == class_name` which doesn't work for aliases :cry:

agrare: > interesting...might be worth it to start killing that (not for this PR)

Yeah, https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/application_controller/miq_request_methods.rb#L972-L1039 is kind of a mess

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
```diff
@@ -31,7 +37,8 @@ def cloud_subnets(_options = {})
     return {} if zone.nil? || cloud_network.nil?
 
     method_log.debug("availability_zone value is #{zone} && cloud_network is #{cloud_network}")
-    subnets = ar_ems.cloud_subnets.select { |sn| sn[:availability_zone_id] == zone && sn.cloud_network.ems_ref == cloud_network }
+    all_cloud_subnets = resources_for_ui[:ems] ? ar_ems.cloud_subnets : ManageIQ::Providers::IbmCloud::VPC::NetworkManager::CloudSubnet.all
+    subnets = all_cloud_subnets.select { |sn| sn[:availability_zone_id] == zone && sn.cloud_network.ems_ref == cloud_network }
```
agrare, line 41: I'm a little concerned about stuff like this though, have to be careful that this doesn't match on a subnet that is in a different EMS from the selected cloud_network if we only have the ems_ref to go on

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
```diff
@@ -6,7 +6,10 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Gene
   # @param _options [void]
   # @return [Hash] Hash with ems_ref as key and name as value.
   def provision_type_to_profile(_options = {})
-    @provision_type_to_profile ||= index_dropdown(ar_ems.flavors)
+    @provision_type_to_profile ||= begin
+      flavors = resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all
+      index_dropdown(flavors)
+    end
```
Fryguy, line 12:
```suggestion
    flavors = resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all
    @provision_type_to_profile = index_dropdown(flavors)
```

Fryguy: (Similar pattern can be applied to the rest of the places)

agrare: Hm I don't know about that one, that would kind of defeat the purpose of memoizing the result since it'd issue the query everytime

Maybe
```ruby
def provision_type_to_profile
  return @provision_type_to_profile if @provision_type_to_profile.present?

  flavors = resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all
  @provision_type_to_profile = index_dropdown(flavors)
end
```

But I'm not sure that's drastically better than what I have already?

Fryguy:
Closed in favor of #362

PR #302, opened 2021-11-09
Subclass AvailabilityZone under IbmCloud::VPC

agrare:
These did not have a subclass under the VPC namespace and thus fell back to `::AvailabilityZone`


agrare:
Will push up a data migration PR shortly

agrare:
Date migration to fix existing records: https://github.com/ManageIQ/manageiq-schema/pull/617

PR #303, opened 2021-11-09
Replace auth_key_pair_create with create

agrare:
Required for:
* https://github.com/ManageIQ/manageiq-api/pull/1098


kbrock:
is this WIP or is this just WIP because it is dependent?

agrare:
Good catch @kbrock thanks, just missed this one

Fryguy (approved):

PR #304, opened 2021-11-12
Fix VPC event catch poll sequence

jaywcarman:
Fix yield / break sequence (yield first to be sure any events get
queued, then break if stop_polling).

@agrare caught this when reviewing the PowerVS event catcher https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/279#discussion_r747636876


agrare (approved):
:+1: LGTM thanks @jaywcarman !

PR #306, opened 2021-11-24
converting supported_for_create to supports :create

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/21573


agrare:
You need to add `supports :create` to `VPC::CloudManager` and `PowerVirtualServers::CloudManager`

agrare:
Ah crap I didn't see you didn't do VPC/PowerVS yet

PR #307, opened 2021-11-24
Use the Regions class support from core

agrare:


agrare:
This is all VPC specific, I need to find out if the PowerVS side has the same regions, a subset of regions, or something different cc @jaywcarman

Fryguy (approved):

Fryguy:
Even so, we can merge as is, because this is a pure refactoring from what I can see (the regions.yml file moved clean with no changes)

agrare:
Yeah good point, taking out of WIP

jaywcarman:
> This is all VPC specific, I need to find out if the PowerVS side has the same regions, a subset of regions, or something different cc @jaywcarman

PowerVS regions are totally different. Since we require a service instance GUID for provider registration we use the [resource controller API](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L167) to get the [CRN](https://cloud.ibm.com/docs/account?topic=account-crn), parse the ["location"](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L16) and then build a ["region" and endpoint URL](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L42-L48).

The location/region names aren't consistent either (for example, we have two _different_ Dallas regions: `us-south` and `dal12`).

I'm not sure what is the best way to fit this into the core `ManageIQ::Providers::Regions` class.

agrare:
Thanks @jaywcarman 

> I'm not sure what is the best way to fit this into the core ManageIQ::Providers::Regions class.

If it doesn't fit we don't have to force it, the core class was mostly consolidating existing logic in one place so if you're not going to benefit from it no issues.

PR #313, opened 2021-12-17
Adding Snapshots support for Inventory

miyamotoh:
Verified working with a VM in PowerVS with 2 snapshots created (separately via CLI). "Size" info is missing, as the PowerVS API doesn't return it. We may have to make a few extra calls, but at least this PR establishes a baseline to work off of. With these changes, my local server can show the list of available snapshots in the UI as well.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
@@ -25,6 +25,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Persister::PowerVirtualServers <
     add_cloud_collection(:operating_systems)
     add_cloud_collection(:auth_key_pairs)
     add_cloud_collection(:miq_templates)
+    add_cloud_collection(:snapshots)
     add_advanced_settings
   end
 
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -114,6 +114,10 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
     @storage_types ||= storage_capacity_api.pcloud_storagecapacity_types_getall(cloud_instance_id)
   end
 
+  def snapshots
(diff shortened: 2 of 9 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/persister/power_virtual_servers.rb
```diff
@@ -50,4 +51,12 @@ def add_advanced_settings
       )
     end
   end
+
+  def add_snapshots
+    add_collection(cloud, :snapshots) do |builder|
+      builder.add_properties(:model_class => ::Snapshot)
+      builder.add_properties(:parent_inventory_collections => %i[vms miq_templates])
+      builder.add_properties(:complete => !targeted?)
+    end
```
agrare, line 60: These options should be the defaults coming from the core inventory collections: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/infra_manager.rb#L59-L64

agrare:
As for the spec failures, you'll need to re-record the VCR cassette so that the API request to get the snapshots doesn't fail.

@Kuldip-Nanda can you help @miyamotoh record cassettes with your terraform rake task?

miyamotoh:
Thanks @Kuldip-Nanda and @jaywcarman for the help on VCR cassette. I got my recording, passing my local runs, included in the latest force-push.

agrare:
Looks good to me @miyamotoh can you add a spec showing the snapshots are being collected?  Ideally showing the count, the specific snapshot attributes, and the reference to the VM

miyamotoh:
@agrare The latest force-push contains a little logic that verifies snapshots are associated with the VM, correct count, and the EMS specific IDs. Please take another look. Thanks!

agrare (approved):
Looks good to me

chessbyte:
@agrare should this be part of an epic on the ManageIQ roadmap?  also, is there anything to update in the [Capabilities Matrix](https://www.manageiq.org/docs/reference/latest/capabilities_matrix)?

agrare:
@chessbyte it is in the 
`[Najdorf] IBM Cloud PowerVS Provider Enhancements` (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/147) issue which is on the roadmap

jaywcarman:
> also, is there anything to update in the Capabilities Matrix?

This PR is limited to collecting existing snapshots. Only snapshot create/remove is included in the matrix so there's no update. That said, @miyamotoh is going to tackle operations next, so it should be coming soon.

PR #324, opened 2022-01-21
IKS Metrics Capture

nasark:
- added the ability to capture metrics for the IKS container provider
- IBM Cloud Monitoring service uses Sysdig PromQL queries (https://docs.sysdig.com/en/docs/sysdig-monitor/metrics-dictionary/metrics-and-label-mapping/mapping-classic-metrics-with-promql-metrics/)

Screen Shot 2022-01-20 at 5 16 55 PM

Documentation PR:
- https://github.com/ManageIQ/manageiq-documentation/pull/1635

Note:
- There is a bug at the moment related to dashboards in the UI, see https://github.com/ManageIQ/manageiq-ui-classic/issues/8068

Reference:
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/314

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
+++ b/app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
@@ -1,0 +1,105 @@
+class ManageIQ::Providers::IbmCloud::ContainerManager::MetricsCapture::PrometheusCaptureContext < ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::PrometheusCaptureContext
+  def initialize(target, start_time, end_time, interval)
+    @target = target
+    @starts = start_time.to_i.in_milliseconds
+    @ends = end_time.to_i.in_milliseconds if end_time
+    @interval = interval.to_i
+    @tenant = target.try(:container_project).try(:name) || '_system'
+    @ext_management_system = @target.ext_management_system
+    @ts_values = Hash.new { |h, k| h[k] = {} }
+    @metrics = []
+
+    @node_hardware = if @target.respond_to?(:hardware)
+                       @target.hardware
+                     else
+                       @target.try(:container_node).try(:hardware)
+                     end
+
+    @node_cores = @node_hardware.try(:cpu_total_cores)
+    @node_memory = @node_hardware.try(:memory_mb)
+
+    validate_target
+  end
+
+  def collect_node_metrics
+    # set node labels
+    labels = labels_to_s(:kube_node_name => @target.name)
+
+    @metrics = %w(cpu_usage_rate_average mem_usage_absolute_average net_usage_rate_average)
+    collect_metrics_for_labels(labels)
+  end
+
+  def collect_container_metrics
+    # set container labels
+    labels = labels_to_s(
+      :container_name      => @target.name,
+      :kube_pod_name       => @target.container_group.name,
+      :kube_namespace_name => @target.container_project.name,
+    )
+
+    @metrics = %w(cpu_usage_rate_average mem_usage_absolute_average)
+    collect_metrics_for_labels(labels)
+  end
+
+  def collect_group_metrics
+    # set pod labels
+
+    labels = labels_to_s(
+      :kube_pod_name       => @target.name,
+      :kube_namespace_name => @target.container_project.name,
+    )
+
+    @metrics = %w(cpu_usage_rate_average mem_usage_absolute_average net_usage_rate_average)
+    collect_metrics_for_labels(labels)
+  end
+
+  def collect_metrics_for_labels(labels)
+    # promQL field is in pct of cpu cores
+    # miq field is in pct of node cpu
+    cpu_resid = "sum(sysdig_container_cpu_cores_used_percent{#{labels}})"
+    fetch_counters_data(cpu_resid, 'cpu_usage_rate_average')
+
+    # promQL field is in bytes, @node_memory is in mb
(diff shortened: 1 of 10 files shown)


config/settings.yml
```diff
@@ -83,9 +83,12 @@
         :poll: 15.seconds
       :event_catcher_ibm_cloud_iks:
         :poll: 15.seconds
+      :event_catcher_prometheus:
+        :poll: 20.seconds
     :queue_worker_base:
       :ems_metrics_collector_worker:
         :ems_metrics_collector_worker_vpc: {}
+        :ems_metrics_collector_worker_ibm_cloud: {}
```
nasark, line 91: Earlier I had `ems_metrics_collector_worker_iks` here but the Travis build failed and suggested `ems_metrics_collector_worker_ibm_cloud` instead. How is this worker name determined?

nasark:
Ready for review 😄

config/settings.yml
agrare: You need to override the settings_name method in that worker, the default settings_name in `MiqWorker` base doesn't handle child managers well.

E.g. from the IbmCloud::ContainerManager::RefreshWorker:
```
  def self.settings_name
    :ems_refresh_worker_ibm_cloud_iks
  end 
```

Recommend staying with this naming convention and using `:ems_metrics_collector_worker_ibm_cloud_iks`

app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,103 @@
...
+                       @target.hardware
+                     else
+                       @target.try(:container_node).try(:hardware)
+                     end
+
+    @node_cores = @node_hardware.try(:cpu_total_cores)
+    @node_memory = @node_hardware.try(:memory_mb)
+
+    validate_target
+  end
```
agrare, line 22: I might just be missing it but does this differ in any way from https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb#L3-L23 which is included in the parent class https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_capture_context.rb#L4

app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    @ts_values
+  end
+
+  def fetch_counters_data(resource, metric_title, conversion_factor = 1)
+    start_sec = (@starts / 1_000) - @interval
+    end_sec = @ends ? (@ends / 1_000).to_i : Time.now.utc.to_i
+
+    client ||= ManageIQ::Providers::IbmCloud::ContainerManager::MetricsCapture::PrometheusClient.new(@ext_management_system).prometheus_client
```
agrare, line 82: `client` looks like a local variable here so the `||=` will always evaluate the assignment

If you want to only instantiate the prometheus_client once (good idea btw) then something like
```
def prometheus_client
  @prometheus_client ||= ManageIQ::Providers::IbmCloud::ContainerManager::MetricsCapture::PrometheusClient.new(@ext_management_system).prometheus_client
end
```

should do the trick

app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
nasark: That was my first thought however I found that the mixin was returning a `ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture` object even when being called from https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/f921529b891e2673fcc383675c051cb5006bc91e/app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture.rb and that led to https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/capture_context_mixin.rb#L27-L29 calling the parent class methods as opposed to the overridden methods at the IBM Cloud provider level

app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
nasark: There is a slight difference in that the provider level client requires the use of headers to instantiate the client (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/f921529b891e2673fcc383675c051cb5006bc91e/app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_client.rb#L15) whereas the parent class client uses `:credentials` instead (https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes/container_manager/metrics_capture/prometheus_client_mixin.rb#L16)

app/models/manageiq/providers/ibm_cloud/container_manager/metrics_capture/prometheus_capture_context.rb
agrare: Interesting will have to dig into that one later

agrare (approved):

PR #325, opened 2022-01-21
Hooking up Create Snapshot action to backend logic

miyamotoh:
Hooking up "Create a new snapshot for this instance" (on a VM in PowerVS) menu action to the backend logic. 

Also being added a little "name" validation logic that's drawn from API response;
image

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -1,4 +1,6 @@
 class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < ManageIQ::Providers::CloudManager::Vm
+  include_concern 'Operations'
+
   supports :terminate
   supports :reboot_guest do
     unsupported_reason_add(:reboot_guest, _("The VM is not powered on")) unless current_state == "on"
@@ -7,6 +9,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
     unsupported_reason_add(:reset, _("The VM is not powered on")) unless current_state == "on"
   end
   supports :snapshots
+  supports :snapshot_create
 
   supports_not :suspend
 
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/snapshot.rb
```diff
@@ -0,0 +1,2 @@
+class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Snapshot < ::Snapshot
```
agrare, line 1: Snapshots don't currently have a `"type"` column which means they don't support STI and subclassing.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -0,0 +1,17 @@
+module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm::Operations
+    extend ActiveSupport::Concern
```
agrare, line 2: Looks like some inconsistent leading spacing

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -38,6 +41,27 @@ def raw_destroy
     end
   end
 
+  def self.params_for_create_snapshot
```
agrare, line 44: The `params_for_create_snapshot` are expected to be an instance method not a class method.  This likely would cause https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/subcollections/snapshots.rb#L51 to raise an exception

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def raw_create_snapshot(_name, _desc = nil, _memory)
+    #run_command_via_parent(:vm_create_snapshot, :name => name, :desc => desc, :memory => memory)
+    with_provider_connection(:service => 'PCloudPVMInstancesApi') do |api|
+      reqBody = IbmCloudPower::SnapshotCreate.new({
+        :name        => _name,
```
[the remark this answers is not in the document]
miyamotoh: Ruby newbie here. Assume it'd be flagged by `rubocop`, but hand-corrected for now. Thanks!

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/snapshot.rb
miyamotoh: Sorry, but I need a little translation here. Without really knowing what this does and how it helps, I got this idea from: https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/cloud_manager/snapshot.rb
Do I not need this file at all?

agrare: Oh sorry, if a model doesn't use STI then you can't subclass it so you can delete this file.  That VMware cloud manager one looks like a bug.

agrare:
Looking good @miyamotoh

agrare (approved):

PR #327, opened 2022-01-28
Wiring delete a snapshot & all snapshots to backend

miyamotoh:
Hooking up delete a selected snapshot and all snapshots of a VM in PowerVS to the backend. Since there is no "delete all" API on PowerVS, this implementation queries and loops through all known snapshots of the VM with the "delete single snapshot" API. When there is an error, it'd post the error, and keep going for every snapshot in the list.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
@@ -15,4 +15,28 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm::Ope
     create_notification(:vm_snapshot_failure, :error => err.to_s, :snapshot_op => "create")
     raise MiqException::MiqVmSnapshotError, err.to_s
   end
+
+  def raw_remove_snapshot(snapshot_id)
+    with_provider_connection(:service => 'PCloudSnapshotsApi') do |api|
+      snapshot = Snapshot.find(snapshot_id)
+      api.pcloud_cloudinstances_snapshots_delete(cloud_instance_id, snapshot.uid_ems)
+    end
+  rescue => err
+    create_notification(:vm_snapshot_failure, :error => err.to_s, :snapshot_op => "delete")
+    raise MiqException::MiqVmSnapshotError, err.to_s
+  end
+
+  def raw_remove_all_snapshots
+    with_provider_connection(:service => 'PCloudSnapshotsApi') do |api|
+      snapshots.each do |snapshot|
+        api.pcloud_cloudinstances_snapshots_delete(cloud_instance_id, snapshot.uid_ems)
+      rescue => err
+        create_notification(:vm_snapshot_failure, :error => err.to_s, :snapshot_op => "delete")
+        raise MiqException::MiqVmSnapshotError, err.to_s
+      end
+    end
+  rescue => err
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -15,4 +15,30 @@ def raw_create_snapshot(name, desc = nil, _memory = nil)
     create_notification(:vm_snapshot_failure, :error => err.to_s, :snapshot_op => "create")
     raise MiqException::MiqVmSnapshotError, err.to_s
   end
+
+  def raw_remove_snapshot(snapshot_id)
+    with_provider_connection(:service => 'PCloudSnapshotsApi') do |api|
+      snapshot = Snapshot.find_by(:id => snapshot_id)
```
agrare, line 21: So `Model.find_by(:id => id)` will return `nil` if the record cannot be found where `Model.find(id)` will raise a RecordNotFound exception but will guarantee that the object you get back is not `nil`.

Issue being if someone deleted the snapshot from the database after this message was queued but before this is executed then `snapshot` would be `nil` here and the line below would raise an exception.

Recommend either `snapshot = Snapshot.find(snapshot_id)` or keep what you current have but check if `snapshot` is `nil` and handle that condition

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -15,4 +15,30 @@ def raw_create_snapshot(name, desc = nil, _memory = nil)
...
+      api.pcloud_cloudinstances_snapshots_delete(cloud_instance_id, snapshot.uid_ems)
+    end
+  rescue => err
+    create_notification(:vm_snapshot_failure, :error => err.to_s, :snapshot_op => "delete")
+    raise MiqException::MiqVmSnapshotError, err.to_s
+  end
+
+  def raw_remove_all_snapshots
+    with_provider_connection(:service => 'PCloudSnapshotsApi') do |api|
+      vm = Vm.find_by(:uid_ems => ems_ref)
```
agrare, line 31: So since this is an instance method you don't need to do a query to get `vm`, `self` is already the `vm` record.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -15,4 +15,30 @@ def raw_create_snapshot(name, desc = nil, _memory = nil)
...
+      snapshots = Snapshot.where(:vm_or_template_id => vm.id)
+      snapshots.each do |snapshot|
```
agrare, line 33: You should already have a `snapshots` association defined on the vm so this should work without the previous two queries

miyamotoh: Wonderful. This is great. The less code, the better! Tested working without the 2 lines, and force-pushed.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
miyamotoh: Great points. Thanks for the suggested code change. Tested fine and included in the newest push.

agrare (approved):
:+1: LGTM

PR #328, opened 2022-01-31
[WIP] setting default RC file location

iv1111:


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -142,7 +142,9 @@ def validate_delete_image
     def_endp = powervc.endpoint(:default)
     auth = powervc.node_auth
 
-    return def_endp.hostname, auth, node_endp.options
+    rcfile = node_endp&.options
+    default_rcfile = '/opt/ibm/powervc/powervcrc'
```
agrare, line 146: Is this a file that exists on Power based appliances?  What about on podified?

iv1111: it exists on a server where PowerVC is installed

iv1111:
Opening a different PR which includes this change as well https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/337

PR #331, opened 2022-02-01
Hooking up Snapshot restore action to backend

miyamotoh:
With the proposed change in place, I can restore the VM to the selected snapshot, if the VM is shut down in advance. The restore operation would fail, if the VM is up (because I chose not to "force" it), and the returned error message from PowerVS is a bit cryptic. Thus, decided to check the `power_state` proactively and `raise` an error with a better message which the user would find in the Notification panel. It basically works, but I get the same notification 2 or 3 times. Is there a better way that'd generate only one such notification?

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -10,6 +10,9 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   end
   supports :snapshots
   supports :snapshot_create
+  supports :revert_to_snapshot do
+    unsupported_reason_add(:revert_to_snapshot, _("Cannot revert to snapshot while VM is running")) unless current_state == "off"
+  end
   supports :remove_snapshot
   supports :remove_all_snapshots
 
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -10,6 +10,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   end
   supports :snapshots
   supports :snapshot_create
+  supports :revert_to_snapshot
```
agrare, line 13: Instead of https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/331/files#diff-b0ec1540d8e7568e453fb09d326d45f1f9aea6af9fd06c4749fdcddd87c2a987R20 you could check the vm's power state here in a supports block, something like

```
supports :revert_to_snapshot do
  unsupported_reason_add(:revert_to_snapshot, "Cannot revert to snapshot while VM is running") unless power_state == "off"
end
```

miyamotoh: This is great as it grays out the "revert to snapshot" icon with a hover message, nullifying the need for check and worrying about duplicate notifications. Addressed in the newest force-push.

agrare (approved):

PR #332, opened 2022-02-02
Calculate snapshot size by tallying up associated volume sizes

miyamotoh:
Adding logic to come up with the size of Snapshot, by querying for size of each associated volume. Collector may or may not have the volume already stored in the db, and may have to call out the API, but I assume that shouldn't happen every time this logic runs.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -307,12 +307,15 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
 
   def snapshots
     collector.snapshots.each do |snapshot|
+      # API does not return the total size of snapshot so we have to sum the size of the associated volumes
+      total_size = snapshot.volume_snapshots.keys.sum { |vol| collector.volume(vol).size&.gigabytes.to_i }
       persister.snapshots.build(
         :uid            => snapshot.snapshot_id,
         :uid_ems        => snapshot.snapshot_id,
         :ems_ref        => snapshot.snapshot_id,
         :name           => snapshot.name,
         :description    => snapshot.description,
+        :total_size     => total_size,
         :create_time    => snapshot.creation_date,
         :vm_or_template => persister.vms.lazy_find(snapshot.pvm_instance_id)
       )


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -307,12 +307,17 @@ def availability_zones
 
   def snapshots
     collector.snapshots.each do |snapshot|
+      sum = 0 # API not returning size of snapshot, thus have to tally up size of associated volumes
+      snapshot.volume_snapshots.each_key do |vol|
+        sum += collector.volume(vol).size&.gigabytes
+      end
```
Fryguy, line 313: Slightly simpler for readability:

```suggestion
      # API not returning size of snapshot, thus have to tally up size of associated volumes
      snapshot.volume_snapshots.keys.map { |vol| collector.volume(vol).size&.gigabytes.to_i }.sum
```

agrare: I think you can use the `sum` directly as the enumerator method: `total_size = snapshot.volume_snapshots.keys.sum { |vol| collector.volume(vol).size&.gigabytes.to_i }`

Fryguy: Oh yes you are right...this is better...I'll update my suggestion

agrare: Love ruby, I just thought "wouldn't it be cool if you could just .sum the result of the block....let me try that and see if it works"

miyamotoh: Sure, looks simpler, but I get;
```
ManageIQ::Providers::BaseManager::Refresher::PartialRefreshError (undefined local variable or method `sum' for #<ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers::CloudManager:0x0000000023d32cc8>)
```
Could you help this Ruby newbie? 🙂

agrare: @Fryguy's example sets the result to `total_size` not `sum` so you'd have to change below from `:total_size     => sum,` to `:total_size     => total_size`

miyamotoh: I was simply confused by the `sum` reference above being the method or the local variable. Revised code force-pushed. Please review again.

Fryguy: Ah yeah sorry about that...forgot I changed the variable name to something less generic.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -307,12 +307,15 @@ def availability_zones
 
   def snapshots
     collector.snapshots.each do |snapshot|
+      # API not returning size of snapshot, thus have to tally up size of associated volumes
```
agrare, line 310: Just a minor change to the wording here maybe,

```suggestion
      # API does not return the total size of snapshot so we have to sum the size of the associated volumes
```

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
miyamotoh: Latest force-push should have everything discussed above. Please take a look!

agrare:
@miyamotoh are there any snapshots in the vcr cassette?  If so could you add a spec test to check the new total_size property?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
miyamotoh: Updated in the latest force-push.

miyamotoh:
There is snapshot examining logic there, but right now @Kuldip-Nanda is `terraform`'ing it. Will have to wait until that's complete.

agrare (approved):

PR #335, opened 2022-02-04
PowerVS additional events

jaywcarman:
Requires:
- https://github.com/ManageIQ/manageiq-content/pull/702

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
@@ -17,6 +17,8 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventPa
       parse_network_event!(event, event_hash)
     when /^pvm-instance/
       parse_vm_event!(event, event_hash)
+    when /^volume/
+      parse_volume_event!(event, event_hash)
     end
 
     event_hash
@@ -49,5 +51,11 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventPa
       event_hash[:vm_ems_ref] = vm.ems_ref
       event_hash[:vm_name]    = vm.name
     end
+  end
+
+  def self.parse_volume_event!(_event, event_hash)
+    # The PCloudEvent API doesn't provide adequate metadata to parse at this
+    # time. Adding this methods as a placeholder for future updates.
+    event_hash
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
@@ -20,7 +20,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::EventPa
   end
 
   def self.parse_vm_event!(event, event_hash)
-    if ['update', 'create'].include?(event['type'])
+    if ['update', 'create'].include?(event[:action])
       event_hash
     else
       pvm_instance_name = event[:message].split('\'')[1]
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
```diff
@@ -6,21 +6,48 @@ def self.event_to_hash(event, ems_id)
...
       parse_vm_event!(event, event_hash)
+    when /^volume/
+      parse_volume_event!(event, event_hash)
     end
 
     event_hash
   end
 
+  def self.parse_image_event!(event, event_hash)
+    if ['delete'].include?(event[:action])
```
agrare, line 30: I'm guessing there are going to be more actions to be added here in the future?

agrare: Or just staying consistent with the other `parse_*_event!` methods?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
```diff
@@ -9,6 +9,25 @@ def parse
...
+        :manager_ref => {:ems_ref => ems_event[:vm_ems_ref]},
+        :manager_id  => ems_event.ext_management_system.id,
+        :event_id    => ems_event.id
+      )
+    when /^image/
+      if ems_event[:message].include?("successfully imported into image catalog")
+        targets << ems_event.ext_management_system
+      end
+    when /^network/
+      network_manager = ManageIQ::Providers::IbmCloud::PowerVirtualServers::NetworkManager.find_by(:parent_ems_id => ems_event[:ems_id])
```
agrare, line 24: `ems_event` should be an EmsEvent object, which `ext_management_system` is already a loaded association for, so you should be able to do `ems_event.ext_management_system.network_manager` without doing the query manually.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
```diff
@@ -9,6 +9,25 @@ def parse
     targets = []
 
     case ems_event[:event_type]
+    when /^image\.delete/
+      targets << InventoryRefresh::Target.new(
+        :association => :miq_templates,
+        :manager_ref => {:ems_ref => ems_event[:vm_ems_ref]},
```
agrare, line 15: It is minior, but prefer to use the accessor methods like `ems_event.vm_ems_ref` so it is more obvious that this isn't just a hash

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
```diff
@@ -20,6 +39,9 @@ def parse
         :manager_id  => ems_event.ext_management_system.id,
         :event_id    => ems_event.id
       )
+    when /^volume/
+      storage_manager = ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager.find_by(:parent_ems_id => ems_event[:ems_id])
+      targets << storage_manager
```
agrare, line 44: Similar here, you can do `targets << ems_event.ext_management_system.storage_manager`

agrare:
Love seeing this getting filled out more, just a few minor things

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_target_parser.rb
jaywcarman: Nice! I'm glad to get rid of the queries.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/event_parser.rb
jaywcarman: There are quite a few `image` events available today, but unfortunately only `delete` provides enough metadata to trigger a refresh (well, actually I'm parsing the message in the `EventTargetParser` to trigger a refresh when an image import is completed - [ref](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/335/files#diff-d14118ccd17b8c31d6b54be4d044f9d6d9902de875fadefe758a35cae6f62f01R20-R22)). I'm hoping in the future they'll standardize the event responses and we can do more here.

agrare (approved):
Looking good

PR #339, opened 2022-02-08
passing COS sensitive info by id and adding RBAC enabled image-import button

iv1111:


iv1111:
@jaywcarman removed WIP

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -42,13 +42,19 @@ def post_execute
...
       :source        => 'url',
       :imageName     => options[:miq_img][:name],
       :osType        => options[:miq_img][:os],
       :diskType      => options[:diskType],
       :imageFilename => "#{options[:session_id]}.ova",
-      **options[:cos_pvs_creds]
+      :region        => cos_pvs_creds[:region],
+      :bucketName    => cos_pvs_creds[:bucketName],
+      :accessKey     => bearer[:userid],
+      :secretKey     => ManageIQ::Password.try_decrypt(bearer[:password])
```
Fryguy, line 57:
```suggestion
      :accessKey     => bearer.userid.,
      :secretKey     => ManageIQ::Password.try_decrypt(bearer.password)
```

Fryguy: Actually I'm confused by the try_decrypt.  The password should be auto-decrypted here.  Perhaps because you weren't going through the accessor methods?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -42,13 +42,19 @@ def post_execute
 
     ems = ExtManagementSystem.find(options[:ems_id])
 
+    cos_pvs_creds = options[:cos_pvs_creds]
+    bearer = Authentication.find(cos_pvs_creds[:bearer_id])
```
Fryguy, line 46: What happens if it's not found?  This will blow up I think, which is probably good.

What happens if somebody passes a random id here...can they then extract authentications from other providers?

At a minimum it feels like we should be scoping the authentications by ems, so I'd expect something like `ems.authentications.find(...)`, but even so, we should probably only look for specific kinds of authentications, to keep the exposure to the minimum.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -56,12 +56,13 @@ def self.raw_import_image(ext_management_system, options = {})
...
 
     location, node_auth, _ = node_creds(options['src_provider_id'])
-    guid, apikey, region, endpoint, access_key, secret_key = cos_creds(options['obj_storage_id'])
+    guid, apikey, region, endpoint, _, _ = cos_creds(options['obj_storage_id'])
     bucket = bucket_name(options['bucket_id'])
     diskType = CloudVolumeType.find(options['disk_type_id']).name
 
     cos_ans_creds = {:resource_instance_id => guid, :apikey => apikey, :bucket_name => bucket, :url_endpoint => endpoint}
-    cos_pvs_creds = {:region => region, :bucketName => bucket, :accessKey => access_key, :secretKey => secret_key}
+    cos_bearer_id = Authentication.find_by(:resource_id => options['obj_storage_id'], :authtype => 'bearer').id
```
Fryguy, line 64: Similar comment here...this feels like we need more scoping on the query itself, even though this only returns an id.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: Good catch. Spoofing COS id will send COS creds to some PowerVC machine. E.g. one register's a PowerVC provider and chooses COS owned by another user. Then COS creds of that user will be sent to the PowerVC machine. One way to fix that would be associating COS with a given PowerVC provider during provider registration. Re-association should be allowed only if COS creds are revalidated.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
agrare: `resource_id` isn't sufficient to query a polymorphic association, you could hit on a different resource_type with the resource_id you're looking for.

And yeah it is possible the auth record could be deleted after this was queued and before it is processed so it could fail to be found.

iv1111: Ok, I thought the IDs were unique regardless of the polymorphic type. I see now. 

The second part is a valid concern, only don't know of methods to check for existence and and get the record atomically. I guess doing "find(...)" would just throw an exception if the record is not found, which is close to what we need. It is "find!" that would return nil in such case from what I know.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: It isn't decrypted at that stage, I've checked it.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
jaywcarman: > One way to fix that would be associating COS with a given PowerVC provider during provider registration.

I like this idea - during COS registration select PowerVC instances that are allowed to access the credentials.

iv1111:
Related PR from the API module: https://github.com/ManageIQ/manageiq-api/pull/1136
It makes sure current user is allowed to view COS by this ID.

iv1111:
Related: https://github.com/ManageIQ/manageiq/pull/21112

iv1111:
@Fryguy I'm adding my own button with RBAC check instead of the generic one 
located in the `miq-ui-classic` base repo. Hope it's ok to do so. I inherit from the basic button.

iv1111:
@agrare Can we get this one merged if all is Ok as it's independent of other PRs.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -41,14 +41,28 @@ def post_execute
     cleanup_git_repository
 
     ems = ExtManagementSystem.find(options[:ems_id])
+    raise MiqException::Error, _("unable to find ems by this id '#{options[:ems_id]}'") if ems.nil?
```
agrare, line 44: @iv1111 when you use `find(id)` if it cannot find the record it will raise a `RecordNotFound` exception and will never return `nil` so this `raise` is unreachable.

If you want to handle the not-found case yourself you can use `ems = ExtManagementSystem.find_by(:id => options[:ems_id])`  instead

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -41,14 +41,28 @@ def post_execute
...
+
+    cos_data = options[:cos_data]
+
+    region = cos_data[:region]
+    bucket_name = cos_data[:bucketName]
+
+    cos = ExtManagementSystem.find(cos_data[:cos_id])
+    raise MiqException::Error, _("unable to find cloud object storage by this id '#{options['obj_storage_id']}'") if cos.nil?
```
agrare, line 52: Same comment here

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: I'd need a custom handling code in order to form a user friendly message, so done as you suggested, using _find_by_.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
iv1111: Same as above.

iv1111:
@agrare corrections submitted

agrare (approved):

Fryguy:
Backported to `najdorf` in commit 18ce3d2a5daf2c1a5311290420fa40044c5723b6.

```text
commit 18ce3d2a5daf2c1a5311290420fa40044c5723b6
Date:   Thu Mar 3 08:34:47 2022 -0500

    Merge pull request #339 from iv1111/hide_cos_creds
    
    passing COS sensitive info by id and adding RBAC enabled image-import button
    
    (cherry picked from commit 664f0e1967ab660846d1c84dd21e7413c92004d9)
```

PR #340, opened 2022-02-09
updated the vm naming convention to what is defined for PowerVS. Also…

Kuldip-Nanda:
Making the VM naming convention rules same as PowerVS.

### DIFF
--- a/content/miq_dialogs/miq_provision_ibm_powervs_vm_dialog.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_powervs_vm_dialog.yaml
@@ -87,7 +87,8 @@
         :vm_name:
           :description: Instance Name
           :required_method: :validate_regex
-          :required_regex: !ruby/regexp /^[a-zA-Z]+$/
+          :required_regex: !ruby/regexp /^[a-zA-Z0-9\-_]+$/
+          :required_regex_fail_details: Only letters (no accents), numbers, underscores and dashes are allowed.
           :required: true
           :display: :edit
           :data_type: :string


content/miq_dialogs/miq_provision_ibm_powervs_vm_dialog.yaml
```diff
@@ -87,7 +87,8 @@
         :vm_name:
           :description: Instance Name
           :required_method: :validate_regex
-          :required_regex: !ruby/regexp /^[a-zA-Z]+$/
+          :required_regex: !ruby/regexp /^[a-zA-Z0-9\-_]+$/
+          :required_regex_fail_details: Only letters (no accents), numbers, underscores and dashes are allowed.
```
agrare, line 91: I wonder if there is a better way to say "Only letters (no accents)" since other non-ascii chars like umlauts probably also aren't supported.

iv1111: Wouldn't this match strings like e.g. 'ABC----DEF---123'?

iv1111:
@Kuldip-Nanda @agrare maybe this is what we need? `/^(([a-zA-Z0-9])[\-_]?)+[a-zA-Z0-9]+$/`


This would match e.g.: 
```
hello-world-1
hel_lo-world-1
hello-world-1a
he-llo-world-1a
he_llo-wor_ld-1a
hello_world_1a
```

**It will not match e.g.:**
```
hello-world-1Ä
hellö-world-1
-hello-world-1a
_hello-world-1a
hello-world-1a-
hello-world-1a_
hello---world-1a
hello---wor___ld-1a

```
One interesting case is this one:
`hello-_world-123`

Do we have to support that? Because the regexp I provided would exclude that. One could match that in addition, if a slight modification is done to the above one: /^(([a-zA-Z0-9])[\-]?[_]?)+[a-zA-Z0-9]+$/

content/miq_dialogs/miq_provision_ibm_powervs_vm_dialog.yaml
agrare: Oh interesting, I'm okay matching the text string that they have.  I also think we need to allow cool-sunglasses-emoji as instance names though :)

Kuldip-Nanda: let me figure out how to fulfill this request milord...
-Kuldip

agrare (approved):

Fryguy:
Backported to `morphy` in commit 40345c3af83bbb84cb8d11469a35f4341f226aa1.

```text
commit 40345c3af83bbb84cb8d11469a35f4341f226aa1
Date:   Thu Feb 10 10:13:08 2022 -0500

    Merge pull request #340 from Kuldip-Nanda/master
    
    updated the vm naming convention to what is defined for PowerVS. Also…
    
    (cherry picked from commit bdd98be5a30fa8f5b6cc567747f84075502a7d71)
```

PR #341, opened 2022-02-10
Powervs catalog provisioning

jaywcarman:
Depends on:
- https://github.com/ManageIQ/manageiq/pull/21697
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8109


jaywcarman:
This is required to enable provisioning through the catalog. It also adds an "Image" field to our standard VM provisioning dialog, which is unnecessary (the user has already selected an image) but seems OK to me since it doesn't require any additional user interaction (pre-selected image shows up as highlighted/selected).
image

agrare (approved):

PR #329, opened 2022-01-31
new terraformed scripts for test automation

Kuldip-Nanda:

PR #345, opened 2022-02-11
Adding Snapshots to `rake provision/spec` flow

miyamotoh:
Adding support for Snapshots on top of Kuldip's work (#329), along with a few cleanups on TF variables/defaults. With this in place, one would;
1. `bundle exec rake provision:apply` to provision the static resources (eg. VM, volumes, networks)
2. `bundle exec rake provision:snapshot` to take a snapshot of the VM
3. `bundle exec rake spec` which would now examine the snapshot as well as all other things
4. `bundle exec rake provision:destroy`, when you are done with testing, to delete all resources, including the snapshot.

The `provision:snapshot` step would fail;
- if the VM is not in "OK" (=healthy) state, for which you could try rebooting the VM if that doesn't happen after provisioning, or
- if you run it again after a successful run (because the snapshot name is hardcoded)


agrare:
This is _awesome_ @miyamotoh  @Kuldip-Nanda can you review as well especially the terraform bits?

spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb
```diff
@@ -159,6 +159,13 @@ def assert_specific_vm
       expect(vm.advanced_settings.find { |setting| setting['name'] == 'placement_group' }).to have_attributes(
         :value        => nil
       )
+
+      expect(vm.snapshots.count).to eq(1)
+      expect(vm.snapshots[0]).to have_attributes(
```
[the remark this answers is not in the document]
miyamotoh: Addressed in the newest force-push.

.gitignore
```diff
@@ -31,3 +31,6 @@
 /lib/tasks_private/terraform/.terraform
 /lib/tasks_private/terraform/.terraform*
 /lib/tasks_private/terraform/.byebug*
+lib/tasks_private/terraform/snapshot/terraform.*
```
jaywcarman, line 34: We can clean this up later, but all of these terraform ignores (include those before this commit could be combined into a few lines:

```
*.log
.terraform*
terraform.*
```

Kuldip-Nanda: When I did that initially, I just wanted to highlight what I am trying to ignore. That is why I made explicit statements.

Kuldip-Nanda: To truly make the test repeatable,  I do have a couple of changes coming along with
1. The code will remove explicit dependency on the public network name
2. If the volume is already existing, it will not recreate a volume. 
3. Start removing hard count on the VM count (and some others etc). 
These changes may be necessary as our requirements have highlighted some Issues/PRs in Terraform Plugin code.
Righjt now these changes can go as in.

lib/tasks_private/terraform/snapshot/variables.tf
```diff
@@ -0,0 +1,19 @@
+variable "ibmcloud_api_key" {
+    description = "Denotes the IBM Cloud API key to use"
+    type = string
```
Kuldip-Nanda, line 3: These changes are very good improvements on my initial submission.

Kuldip-Nanda: approved

Kuldip-Nanda (approved):

jaywcarman (approved):
Nice work @miyamotoh !

PR #346, opened 2022-02-21
Add ActiveSupport inflection rule for PowerVS

jaywcarman:
Thanks to @Fryguy for telling me about inflections:
https://github.com/ManageIQ/manageiq/pull/21697#issuecomment-1036702555

The IBM Power IaaS is officially named "IBM Power Systems Virtual
Servers". We've morphed this product name into our class name as
"PowerVirtualServers", which breaks the rails convention of using
singular class names.

To fix this, we can add an inflection rule defining the singular of
"PowerVirtualServers" to be itself, "PowerVirtualServers". With this now
singular and classify methods work:

```ruby
irb(main):001:0> 'IbmCloud::PowerVirtualServers'.singularize
=> "IbmCloud::PowerVirtualServers"
irb(main):002:0> 'IbmCloud::PowerVirtualServers'.classify
=> "IbmCloud::PowerVirtualServers"
```

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -1,3 +1,6 @@
 ActiveSupport::Inflector.inflections do |inflect|
   inflect.singular(/PowerVirtualServers$/, 'PowerVirtualServers')
+  inflect.plural(/PowerVirtualServers$/, 'PowerVirtualServers')
+  inflect.singular(/power_virtual_servers$/, 'power_virtual_servers')
+  inflect.plural(/power_virtual_servers$/, 'power_virtual_servers')
 end
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -1,0 +1,3 @@
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.singular(/PowerVirtualServers$/, 'PowerVirtualServers')
+end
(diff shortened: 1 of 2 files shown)


config/initializers/inflections.rb
```diff
@@ -0,0 +1,3 @@
+ActiveSupport::Inflector.inflections do |inflect|
+  inflect.singular(/PowerVirtualServers$/, 'PowerVirtualServers')
```
Fryguy, line 2: There should probably be a singular inflection (what you have) and also a plural inflection (for going the other direction, otherwise Rails will drop the "s" thinking it should singularize.

I recommend adding some specs a la https://github.com/ManageIQ/manageiq/blob/master/spec/initializers/as_inflections_spec.rb#L1-L8

Fryguy: Also I think we need to beef up that core spec 😆

Fryguy:
@jaywcarman Should this go back to najdorf?

jaywcarman:
> @jaywcarman Should this go back to najdorf?

I don't know of any exposure in najdorf so I didn't tag it. That said, it doesn't seem like a bad idea to go ahead and backport if you would like to.

Fryguy:
Backported to `najdorf` in commit 484476685039b57de9559f4b71ba0bb94b2ec0e0.

```text
commit 484476685039b57de9559f4b71ba0bb94b2ec0e0
Date:   Mon Feb 28 10:40:31 2022 -0500

    Merge pull request #346 from jaywcarman/powervs_singular_inflection
    
    Add ActiveSupport inflection rule for PowerVS
    
    (cherry picked from commit 1794aea87f1941bfa32819f5e4a43bb8e44da409)
```

PR #349, opened 2022-03-04
added Fingerprint calculation as it does not come with the get ssh key rest api for powervs

Kuldip-Nanda:


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -248,10 +249,13 @@ def networks
...
   def sshkeys
     collector.sshkeys.each do |tkey|
+      cert = OpenSSL::X509::Certificate.new(tkey.ssh_key + "\n")
+      fingerprint = OpenSSL::Digest::SHA1.hexdigest(cert.to_der).scan(/../).join(':')
       tenant_key = {
         :creationDate => tkey.creation_date,
         :name         => tkey.name,
         :sshKey       => tkey.ssh_key,
+        :fingerprint  => fingerprint
       }
```
Fryguy, line 259: @Kuldip-Nanda Where is this tenant_key actually used?  When it build the auth_key_pair below, it only uses the name, so the rest of these are not needed at all?

Fryguy:
Failures look legit.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
jaywcarman: I don't think tenant_key used anywhere.. possibly a past oversight?

Would be more clear to get rid of tenant_key altogether and just do:
```ruby
      persister.auth_key_pairs.build(           
        :name        => tkey.name,              
        :fingerprint => fingerprint,
        :public_key  => tkey.ssh_key
      )
```

Fryguy: A spec can be added to ensure this is working too.

jaywcarman: Also, I think we need to set an `ems_ref` here. Since the keys are owned at the (PowerVS) tenant level, we end up with duplicate DB entries for ssh keys (1 x PowerVirtualServers::CloudManager even though they're shared). The PowerVS managers have access to the tenant_id [[ref]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L49-L52), we just need something like `:ems_ref     => "#{pcloud_tenant_id}-#{tkey.name}"` here.

agrare:
@Kuldip-Nanda I thought the bug you were fixing was in `create_auth_key_pair`?  This sets fingerprint in refresh but we still need to update the raw_create_key_pair method

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: @jaywcarman  thanks for pointing this out. Let me make this change.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -248,10 +249,13 @@ def networks
 
   def sshkeys
     collector.sshkeys.each do |tkey|
+      cert = OpenSSL::X509::Certificate.new(tkey.ssh_key + "\n")
+      fingerprint = OpenSSL::Digest::SHA1.hexdigest(cert.to_der).scan(/../).join(':')
```
agrare, line 253: `OpenSSL::X509::Certificate` is for an SSL cert where you have an ssh key, recommend

```suggestion
      fingerprint = OpenSSL::Digest::SHA1.hexdigest(tkey.ssh_key)
```

agrare: Yeah but you don't want an X509 certificate so getting that line to not fail isn't very helpful :smile: , it looks like you followed https://stelfox.net/blog/2014/calculating-rsa-key-fingerprints-in-ruby/

> X509 certificates also use a key fingerprint to help identify a certificate's signing authority.

This isn't what you want to do here.  The closer analog to what you want to do is further down, `The next one I got working was the SSH fingerprints thanks to the RFCs mentioned earlier.`

But it is even easier than that, because you already get a formatted public key so you can completely skip the first two:
```
key = OpenSSL::PKey::RSA.new(File.read(path_to_key))
data_string = [7].pack('N') + 'ssh-rsa' + key.public_key.e.to_s(0) + key.public_key.n.to_s(0)
```
^ you get essentially these two "for free' because the API is returning this format to you already

Did you happen try my suggestion?

Fryguy: I think the problem is that only the DER portion of the key is how the fingerprint is calculated, so the to_der is the important part that has to be kept.

So I _think_ this should work:

```suggestion
      ssh_key_der = OpenSSL::PKey::RSA.new(tkey.ssh_key).to_der
      fingerprint = OpenSSL::Digest::SHA1.hexdigest(ssh_key_der).scan(/../).join(':')
```

Fryguy: Unless maybe I'm misunderstand what the structure of the `tkey.ssh_key` is?

agrare: `tkey.ssh_key` is already a standard exported public key string so `OpenSSL::PKey::RSA.new(tkey.ssh_key)` isn't going to work.  Also note that would fail for non RSA keys.

Fryguy: Ah ok, I was misunderstanding how the key was being returned by the API and assumed it was in PEM format.  Though I am now more confused about the DER format and the content of a .pub file.  I thought hexdigest needed the binary format.

agrare: We could use `sshkey` gem https://github.com/bensie/sshkey#fingerprints

agrare: Wow so I completely forgot that I did basically exactly this for GCE back in '15 https://github.com/ManageIQ/manageiq/pull/6109/files#diff-9f8ff9a0bafc734718f6453ef8e5994673ff993d54bb721e5e18ed13a8921b6aR307

Fryguy: Oh then we should definitely use this gem and not roll our own.  @Kuldip-Nanda can you look into doing this the same way that @agrare did?

Fryguy: So is this change as simple as this:

```suggestion
      fingerprint = SSHKey.sha1_fingerprint(tkey.ssh_key)
```

(as well as changing the require to `require 'ssh_key'`, and also adding ssh key to the gemspec?)

Gemfile
```diff
@@ -2,6 +2,7 @@
 # Bundler will treat runtime dependencies like base dependencies, and
 # development dependencies will be added by default to the :development group.
 gemspec
+gem "sshkey",                         "~>1.8.0",           :require => false
```
agrare, line 5: This should go in the gemspec: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/manageiq-providers-ibm_cloud.gemspec#L30

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -248,14 +249,11 @@ def networks
...
-        :creationDate => tkey.creation_date,
-        :name         => tkey.name,
-        :sshKey       => tkey.ssh_key,
-      }
-
-      # save the tenant instance
-      persister.auth_key_pairs.build(:name => tenant_key[:name])
+      persister.auth_key_pairs.build(
+        :name        => tenant_key[:name])
+        :sshKey      => tkey.ssh_key,
```
agrare, line 254: There is no `sshKey` column on the authentications table, I think you want `:public_key` here

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/auth_key_pair.rb
```diff
@@ -1,13 +1,15 @@
...
 
   def self.raw_create_key_pair(ext_management_system, create_options)
     ext_management_system.with_provider_connection(:service => "PCloudTenantsSSHKeysApi") do |api|
       tenant_id = ext_management_system.pcloud_tenant_id(api.api_client)
       ssh_key   = IbmCloudPower::SSHKey.new(
-        :name    => create_options['name'],
-        :ssh_key => create_options['public_key']
+        :name        => create_options['name'],
+        :ssh_key     => create_options['public_key'],
+        :fingerprint => SSHKey.sha1_fingerprint(create_options['public_key'])
```
agrare, line 12: I don't see a `fingerprint` attribute for the `IbmCloudPower::SSHKey` class only name, ssh_key, and creation_date.  I don't think you want to pass the fingerprint in to the `pcloud_tenants_sshkeys_post` call.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/auth_key_pair.rb
```diff
@@ -1,13 +1,15 @@
...
       )
 
       api.pcloud_tenants_sshkeys_post(tenant_id, ssh_key)
```
agrare, line 15: @Kuldip-Nanda part of https://github.com/ManageIQ/manageiq/pull/21759 is to return a hash with the appropriate attributes needed to create an auth_key_pair record.

See https://github.com/ManageIQ/manageiq-providers-openstack/pull/784/files#diff-0e8e6c9331d12cabc425f7bd0135e291136e172135c5f5bc8a3c96e9b9f919aeR12 for an example

Gemfile
Kuldip-Nanda: I was not sure. Let me move it there.

manageiq-providers-ibm_cloud.gemspec
```diff
@@ -18,6 +18,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+
+  spec.add_dependency "sshkey", "~>1.8.0"
```
agrare, line 22: We try to keep these listed alphabetically if you can drop this under the `rest-client` dep

Kuldip-Nanda: I thought this is more a system call. and move it at the top. Just old C habbit. Let me move it down.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -1,4 +1,5 @@
 class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < ManageIQ::Providers::IbmCloud::Inventory::Parser
+  require "sshkey"
```
Fryguy, line 2: @agrare Opinions on the location of the require?  I don't think it should be top-level because it would get required even if a customer never has an PowerVS provider.  I'm thinking either the sshkeys method itself or initialize.

manageiq-providers-ibm_cloud.gemspec
```diff
@@ -18,6 +18,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+
```
agrare, line 21: No need for this extra newline here now

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
agrare: :+1: yeah can move it into the parser method right above https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/349/files#diff-f985d1bd514a044797c4062d1d8d94ff3da33e97f6bfa8d182955661184e7d45R252

manageiq-providers-ibm_cloud.gemspec
Kuldip-Nanda: sure :===(. let me remove it.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Kuldip-Nanda: so anti c, anti pythonic (even though I have done includes to some specific functions). anyways doing it.

Fryguy: @Kuldip-Nanda yeah, requires in Ruby are very different than in C and Python.  They are really just a normal method call that loads some code at that moment.  As such, we don't want it loaded at boot because it incurs a memory cost that may not be needed, and we'd prefer that were delayed until the last moment.

Fryguy (approved):

agrare (approved):

agrare:
This should be merged as part of https://github.com/ManageIQ/manageiq/pull/21759

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/auth_key_pair.rb
```diff
@@ -11,6 +12,11 @@ def self.raw_create_key_pair(ext_management_system, create_options)
...
+      {
+        :name        => create_options['name'],
+        :auth_key    => create_options['public_key'],
```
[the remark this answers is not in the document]
Fryguy: Good eyes

Fryguy:
Gah just realized this didn't have any specs, but they can be added afterwards.

Fryguy:
Backported to `najdorf` in commit c07ff51409b07d22d2db076875f8a07c3e29a8a2.

```text
commit c07ff51409b07d22d2db076875f8a07c3e29a8a2
Date:   Tue Mar 8 14:46:14 2022 -0500

    Merge pull request #349 from Kuldip-Nanda/fingerprint
    
    added Fingerprint calculation as it does not come with the get ssh key rest api for powervs
    
    (cherry picked from commit f0fd4d9f9819add410e481fbc3bb37d1370a1d5f)
```

PR #352, opened 2022-03-10
fix in IBM COS authentication validation

iv1111:
* added missing parameter in 'verify_credentials' 
* implemented 'required_credential_fields' method

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
@@ -215,8 +215,8 @@ class ManageIQ::Providers::IbmCloud::ObjectStorage::StorageManager < ManageIQ::P
 
   def required_credential_fields(type)
     case type.to_s
-    when "default" %i[auth_key]
-    when "bearer"  %i[password]
+    when "default" then %i[auth_key]
+    when "bearer"  then %i[password]
     else           []
     end
   end
--- a/app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
@@ -126,7 +126,7 @@ class ManageIQ::Providers::IbmCloud::ObjectStorage::StorageManager < ManageIQ::P
     self.class.raw_connect(region, endpoint, access_key, secret_key)
   end
 
-  def verify_credentials(_args = {})
+  def verify_credentials(_args = {}, _options = {})
     guid, apikey, region, endpoint, access_key, secret_key = cos_creds
     self.class.verify_bearer(region, endpoint, access_key, secret_key)
     self.class.verify_iam(guid, apikey)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -126,7 +126,7 @@ def connect(_args = {})
     self.class.raw_connect(region, endpoint, access_key, secret_key)
   end
 
-  def verify_credentials(_args = {})
+  def verify_credentials(_args = {}, _options = {})
```
agrare, line 129: The args for this should be `def verify_credentials(_auth_type = nil, _options = {})` (see https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/authentication_mixin.rb#L468)

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -212,4 +212,14 @@ def remove_object(bucket_name, object_name)
...
   end
+
+  def required_credential_fields(type)
+    if type == 'default'
+      [:auth_key]
+    elsif type == 'bearer'
+      [:password]
+    else
+      []
+    end
```
agrare, line 223: Best to `.to_s` the type in case e.g. `:default` is passed in

Also up to you but could be a good use for case/when e.g.
```suggestion
    case type.to_s
    when "default" [:auth_key]
    when "bearer"  [:password]
    else           []
    end
```

agrare: Also for `bearer` it looks like you need `%i[userid password]`

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
iv1111: Corrected, thanks!

agrare (approved):

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -126,7 +126,7 @@ def connect(_args = {})
     self.class.raw_connect(region, endpoint, access_key, secret_key)
   end
 
-  def verify_credentials(_args = {})
+  def verify_credentials(_auth_type = nil, _options = {}
```
agrare, line 129: Ah lost a `)` here
```suggestion
  def verify_credentials(_auth_type = nil, _options = {})
```

agrare: Approved too quickly ;)

iv1111: yeah, I did fix now )

iv1111:
I will retest now and remove WIP.

iv1111:
Tested by running MIQ via `rake evm:start`. 

Added a new COS instance, after the `refresh` I see the auth. status as a green checkmark:

image

Running the `authentication_check` from the console returns:

image


@agrare if code is ok, then let's merge it.

Fryguy:
Backported to `najdorf` in commit 7d9cb3fc065dc56ab8df18680fc4710db07bc325.

```text
commit 7d9cb3fc065dc56ab8df18680fc4710db07bc325
Date:   Thu Mar 10 12:33:53 2022 -0500

    Merge pull request #352 from iv1111/auth_fix
    
    fix in IBM COS authentication validation
    
    (cherry picked from commit 8ec2db725b3c54ea7603e4c59f013dbcf8a8730d)
```

PR #356, opened 2022-03-11
fix in the image import queuing mechanism

iv1111:
Depends on:

- [x] https://github.com/ManageIQ/manageiq/pull/21787

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
@@ -38,7 +38,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::ImageImp
   end
 
   def start
-    queue_signal(:pre_execute, :msg_timeout => options["timeout"]&.to_i&.hours&.seconds)
+    queue_signal(:pre_execute, :msg_timeout => options[:timeout])
   end
 
   def post_execute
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/image_import_workflow.rb
```diff
@@ -37,6 +37,10 @@ def process_runner_result(result)
     end
   end
 
+  def start
+    queue_signal(:pre_execute, :msg_timeout => options["timeout"]&.to_i&.hours&.seconds)
```
agrare, line 41: @iv1111 what is the format of `options["timeout"]`? Can you give an example?

iv1111: It's integers from 1 to 24 signifying number of hours the user wishes to set as a timeout value for the whole workflow to complete. The UI form sets value 3 by default.

This particular line needs a fix anyway as _timeout_ is supposed to be a symbol and not a string here.

iv1111: I just debugged this and it seems like the `options[:timeout]` is in minutes instead of hours as I assumed. I guess I've missed the fact that it get's transformed somewhere in the base classes. I'll dig in to see.

iv1111: If I get this right the `msg_timout` on the other hand is expected to be in seconds.

iv1111: I just debugged this and it seems like the `options[:timeout]` is in seconds already.

agrare (approved):

PR #358, opened 2022-03-21
Fixed security_groups select bug

GilbertCherrie:
Fixes an issue in the image provision form that does not allow for any security groups to be selected.

Before:
Screen Shot 2022-03-21 at 2 18 54 PM

After:
Screen Shot 2022-03-21 at 2 15 24 PM

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
@@ -40,7 +40,7 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Netw
 
   # Fetch a list of security groups.
   # @param _options [void]
-  # @return [Hash<String, String>] Hash with ems_ref as key and name as value.
+  # @return [Hash<Integer, String>] Hash with id as key and name as value.
   def security_group_to_security_group(_options = {})
     cloud_network = field(:cloud_network)
     return {} if cloud_network.nil?
@@ -48,7 +48,7 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Netw
     ar_security_group = ar_ems.security_groups.select do |security_group|
       security_group.cloud_network.ems_ref == cloud_network
     end
-    string_dropdown(ar_security_group)
+    index_dropdown(ar_security_group)
   rescue => e
     logger(__method__).ui_exception(e)
   end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/payload.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/payload.rb
@@ -22,7 +22,7 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::Provision::Payload
       :primary_network_interface => {
         :name            => 'eth0',
         :subnet          => {:id => get_option(:cloud_subnet)},
-        :security_groups => [{:id => get_option(:security_groups)}]
+        :security_groups => security_groups.map { |sg| {:id => sg.ems_ref} }
       },
       :volume_attachments        => cloud_volumes,
       :boot_volume_attachment    => {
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
```diff
@@ -40,15 +40,15 @@ def cloud_subnets(_options = {})
 
   # Fetch a list of security groups.
   # @param _options [void]
-  # @return [Hash<String, String>] Hash with ems_ref as key and name as value.
+  # @return [Hash<Integer, String>] Hash with ems_ref as key and name as value.
```
agrare, line 43: > Hash with ems_ref as key and name as value.

Assume this has to change to say "with ID as the key...."

GilbertCherrie:
@agrare Are you able to help verify if these are valid changes. This pr fixes the drop down field but I was not sure if changing the values of the dropdown options from ems_ref to security group ids would break anything from a back end perspective. After talking to @kavyanekkalapu we were not able to figure out if this is a valid change and were hoping you can take a look.
Not sure if this means anything but the request page does seem to know what security groups we selected by just passing in the ids.

Screen Shot 2022-03-21 at 3 19 50 PM (1)

agrare:
This method is only used by that provision dialog, looks good to me
@GilbertCherrie we need to find out how long this has been broken for so we know if we need to backport to najdorf/morphy/etc...

agrare (approved):

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/network.rb
```diff
@@ -40,15 +40,15 @@ def cloud_subnets(_options = {})
...
+  # @return [Hash<Integer, String>] Hash with id as key and name as value.
   def security_group_to_security_group(_options = {})
     cloud_network = field(:cloud_network)
     return {} if cloud_network.nil?
 
     ar_security_group = ar_ems.security_groups.select do |security_group|
       security_group.cloud_network.ems_ref == cloud_network
     end
-    string_dropdown(ar_security_group)
+    index_dropdown(ar_security_group)
```
agrare, line 51: @GilbertCherrie do you have an example of what this looks like when it makes it to the payload? https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision/payload.rb#L25
I think you need this to be ems_ref not id

agrare: It looks like currently we only support one being selected, if you want to support multiple it'd have to be a string array and we'd have to change the payload to do something like `security_groups => get_option(:security_groups).map { |ems_ref| {:id => ems_ref} }`

agrare (approved):
I think we can refactor this to use the more standard allowed_security_groups from core but that is out of scope of this fix

GilbertCherrie:
> @GilbertCherrie we still need to find out if this was broken in morphy and/or najdorf (possible it was always broken since the feature was introduced)

I'll check on morphy and najdorf but I am assuming its been broken since the code was introduced as I haven't seen any changes to this code in the history.

agrare:
:+1: thanks, wasn't sure if it was something in the e.g. UI that changed causing the failure

GilbertCherrie:
@agrare I couldn't get my morphy running locally but its broken on Najdorf and no changes were made to any of this code since it was introduced so it is probably broken on morphy as well.

Fryguy:
Backported to `najdorf` in commit 608b10076a90e7ea88e1e5d7584ff4d689c73a1b.

```text
commit 608b10076a90e7ea88e1e5d7584ff4d689c73a1b
Date:   Tue Mar 22 14:48:50 2022 -0400

    Merge pull request #358 from GilbertCherrie/fix_instance_provision
    
    Fixed security_groups select bug
    
    (cherry picked from commit 9219e5f96e9b12950a64983154e1ff0fa1096bda)
```

Fryguy:
Backported to `morphy` in commit 044f96f6a28fc4c67eec5b78dd7a9fc8cf3c3bca.

```text
commit 044f96f6a28fc4c67eec5b78dd7a9fc8cf3c3bca
Date:   Tue Mar 22 14:48:50 2022 -0400

    Merge pull request #358 from GilbertCherrie/fix_instance_provision
    
    Fixed security_groups select bug
    
    (cherry picked from commit 9219e5f96e9b12950a64983154e1ff0fa1096bda)
```

PR #361, opened 2022-03-25
Add specs for VPC provision workflow

agrare:
I'm looking to refactor some of these methods and found there were no specs covering the IBM VPC Provision Workflow at all

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
@@ -20,7 +20,8 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Comm
   # Required method to display the provision workflow in UI.
   # @param _message [String]
   # @return [void]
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
```diff
@@ -20,7 +20,8 @@ def ar_ems
   # Required method to display the provision workflow in UI.
   # @param _message [String]
   # @return [void]
-  def dialog_name_from_automate(_message = 'get_dialog_name')
+  def dialog_name_from_automate(message = 'get_dialog_name')
+    super(message, {'platform' => 'ibm_vpc'})
```
agrare, line 24: NOTE this was not implemented and was causing a spec which is common in other provider provision_workflow specs to fail :point_down: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/361/files#diff-ccaa497a4444856f0d0ba4146b3aa3102527657226abcfcfcfc1d57c21cd41e7R15-R19

For consistency with the rest of the provider's provision_workflow I decided to implemented this

nasark (approved):
LGTM cc @Fryguy

Issue #342, opened 2022-02-10
[Oparin] IBM Cloud VPC Provider Enhancements


agrare:
Items completed in Najdorf: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/263

PR #362, opened 2022-03-30
Add support for VPC catalog provisioning 

nasark:
Continuing from https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/301

Ref:
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/342

@miq-bot assign @agrare

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
@@ -6,15 +6,10 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Comm
   # @raise [MiqException::MiqProvisionError] Unable to get a new object from server.
   # @return [ManageIQ::Providers::IbmCloud::VPC::CloudManager]
   def ar_ems
-    return @ar_ems unless @ar_ems.nil?
-
     rui = resources_for_ui[:ems]
     ems = load_ar_obj(rui) if rui
-    raise 'VPC EMS could not be found. Raising an exception.' if ems.nil?
 
-    @ar_ems = ems
-  rescue => e
-    logger(__method__).log_backtrace(e, :context_msg => 'Fetching ems for VPC cloud.')
+    ems
   end
 
   # Required method to display the provision workflow in UI.
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
@@ -6,7 +6,10 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Gene
   # @param _options [void]
   # @return [Hash] Hash with ems_ref as key and name as value.
   def provision_type_to_profile(_options = {})
-    @provision_type_to_profile ||= index_dropdown(ar_ems.flavors)
+    @provision_type_to_profile ||= begin
+      flavors = resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all
+      index_dropdown(flavors)
+    end
   rescue => e
     logger(__method__).ui_exception(e)
   end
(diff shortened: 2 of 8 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/common.rb
```diff
@@ -6,15 +6,10 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Comm
...
   def ar_ems
-    return @ar_ems unless @ar_ems.nil?
-
     rui = resources_for_ui[:ems]
     ems = load_ar_obj(rui) if rui
-    raise 'VPC EMS could not be found. Raising an exception.' if ems.nil?
 
-    @ar_ems = ems
-  rescue => e
-    logger(__method__).log_backtrace(e, :context_msg => 'Fetching ems for VPC cloud.')
```
nasark, line 17: The `raise` here was counterintuitive because in the case that `ar_ems` is `nil` we want to return an empty hash instead of raising an exception

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
```diff
@@ -6,7 +6,12 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Gene
   # @param _options [void]
   # @return [Hash] Hash with ems_ref as key and name as value.
   def provision_type_to_profile(_options = {})
-    @provision_type_to_profile ||= index_dropdown(ar_ems.flavors)
+    return {} if ar_ems.nil?
+
+    @provision_type_to_profile ||= begin
+      flavors = resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all
```
agrare, line 12: Is it possible to ever make it to `ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all` if we're returning if `ar_ems.nil?` above?

agrare:
Hey @nasark I'm not sure where exactly `resources_for_ui[:ems]` is set, but if we have a `src_vm_id` now is it possible to keep the exception if the ar_ems is nil?

nasark:
> Hey @nasark I'm not sure where exactly `resources_for_ui[:ems]` is set, but if we have a `src_vm_id` now is it possible to keep the exception if the ar_ems is nil?

@agrare Here are my observations:

When the exception is thrown, the following message is displayed on the UI and the user is unable to proceed to adding a catalog item. It seems like `ar_ems`/`resources_for_ui[:ems]` is `nil` when landing on the page and so returning an empty hash `{}` is necessary to proceed. Raising an exception would block the execution. Once the page has loaded then `ar_ems`/`resources_for_ui[:ems]` are no longer `nil` and populated with appropriate values

Screen Shot 2022-03-30 at 2 31 01 PM

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
nasark: Doesn't seem like it's possible. Should I remove the ternary operator and similar occurrences in other methods?

agrare:
Oh okay so it is being evaluated before a template has been selected?  If that's the case then we might not need to do the `resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all` because by the time you are selecting a flavor the ems should have already been selected (?)

nasark:
Yeah

> Oh okay so it is being evaluated before a template has been selected? If that's the case then we might not need to do the `resources_for_ui[:ems] ? ar_ems.flavors : ManageIQ::Providers::IbmCloud::VPC::CloudManager::Flavor.all` because by the time you are selecting a flavor the ems should have already been selected (?)

@agrare Yeah seems like it https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/362#discussion_r838861420. I will push the change

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow/general.rb
```diff
@@ -6,7 +6,11 @@ module ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow::Gene
   # @param _options [void]
   # @return [Hash] Hash with ems_ref as key and name as value.
   def provision_type_to_profile(_options = {})
-    @provision_type_to_profile ||= index_dropdown(ar_ems.flavors)
+    return {} if ar_ems.nil?
+
+    @provision_type_to_profile ||= begin
+      index_dropdown(ar_ems.flavors)
+    end
```
agrare, line 13: If these are one line now this can stay as `@provision_type_to_profile ||= index_dropdown(ar_ems.flavors)`

nasark:
@agrare Ready for review

nasark:
@agrare Ready for review

agrare (approved):

Fryguy:
Backported to `najdorf` in commit 9035bc55aacaa0853b8d16a51293db2cb74c58f6.

```text
commit 9035bc55aacaa0853b8d16a51293db2cb74c58f6
Date:   Thu Mar 31 14:16:25 2022 -0400

    Merge pull request #362 from nasark/add_support_for_catalog_provisioning
    
    Add support for VPC catalog provisioning
    
    (cherry picked from commit 0a09c19d381999477723d464b07d99202afd76f0)
```

PR #368, opened 2022-04-12
Add params for create/update cloud databases on VPC

nasark:
- The API isn't implemented yet and so not able to test this on the UI for the time being

@miq-bot add_labels enhancement
@miq-bot assign @agrare, @GilbertCherrie

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/cloud_database.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/cloud_database.rb
@@ -3,6 +3,60 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::CloudDatabase < ::CloudD
   supports :delete
   supports :update
 
+  def self.params_for_create(ems)
+    {
+      :fields => [
+        {
+          :component => 'text-field',
+          :id        => 'name',
+          :name      => 'name',
+          :label     => _('Cloud Database'),
+        },
+        {
+          :component    => 'select',
+          :name         => 'resource_group_name',
+          :id           => 'resource_group_name',
+          :label        => _('Resource Group'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :options      => ems.resource_groups.map do |rg|
+            {
+              :label => rg.name,
+              :value => rg.name,
+            }
+          end,
+        },
+        {
+          :component    => 'select',
+          :name         => 'database',
+          :id           => 'database',
+          :label        => _('Cloud Database'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :options      => ["postgresql", "edb", "mysql", "datastax", "mongodb", "elasticsearch", "redis", "etcd"].map do |db|
+            {
+              :label => db,
+              :value => db,
+            }
+          end,
+        }
+      ],
+    }
+  end
+
+  def params_for_update
+    {
+      :fields => [
+        {
+          :component => 'text-field',
+          :id        => 'name',
+          :name      => 'name',
+          :label     => _('Rename Cloud Database'),
+        }
+      ],
+    }
+  end
+
   def self.raw_create_cloud_database(ext_management_system, options)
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/cloud_database.rb
```diff
@@ -3,6 +3,54 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::CloudDatabase < ::CloudD
...
+          end,
+        },
+        {
+          :component    => 'select',
+          :name         => 'database',
+          :id           => 'database',
+          :label        => _('Cloud Database'),
+          :includeEmpty => true,
+          :isRequired   => true,
+          :options      => ["postgresql", "edb", "mysql", "datastax", "mongodb", "elasticsearch", "redis", "etcd"].map do |db|
```
nasark, line 30: I only have access to deploy postgresql cloud databases so not able to confirm for other databases however, the assumption is that they will follow the same format for the `resource_plan_id` https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/cloud_database.rb#L13
Screen Shot 2022-04-12 at 11 46 03 AM

agrare:
@GilbertCherrie can you review?

GilbertCherrie (commented):
@nasark @agrare Didn't test this but the code looks fine. Only question I have is why the params for update has a name field but params for create does?

agrare:
> Only question I have is why the params for update has a name field but params for create does not?

That's a good question, @nasark is the name not able to be set during creation?

nasark:
I was under the impression that the `name` field was already there by default, as seen in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/storage_manager/cloud_volume.rb and https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/network_manager/cloud_network.rb where there is no `name` under `params_for_create` but `options[:name]` is still used. Also on the UI, the `name` field on the forms for those resources is still present.

@GilbertCherrie @agrare  can you guys confirm or explain why this may be the case?

GilbertCherrie:
> I was under the impression that the `name` field was already there by default, as seen in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/storage_manager/cloud_volume.rb and https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/vpc/network_manager/cloud_network.rb where there is no `name` under `params_for_create` but `options[:name]` is still used. Also on the UI, the `name` field on the forms for those resources is still present.
> 
> @GilbertCherrie @agrare can you guys confirm or explain why this may be the case?

Some forms have name as a common field between providers so it is implemented in the UI code. However, I assumed it is better to include the name as part of the provider fields just so that the only field implemented on the UI side is usually just the EMS select. This is how I have been creating my prs for params for create/edit

nasark:
Thanks @GilbertCherrie that makes sense. Since the UI portion for cloud databases isn't implemented yet I'll add in the `name` field 👍

GilbertCherrie (approved):

agrare (approved):

Fryguy:
Backported to `najdorf` in commit b3a9854820a09915cefa0027e699604e8d259b2c.

```text
commit b3a9854820a09915cefa0027e699604e8d259b2c
Date:   Tue Apr 12 15:12:05 2022 -0400

    Merge pull request #368 from nasark/add_cloud_db_params_for_create_update
    
    Add params for create/update cloud databases on VPC
    
    (cherry picked from commit 2eb998fc130cc77b521d6bf02e3a8ad9a449a4ca)
```

PR #374, opened 2022-05-04
Adding support for clone-to-template on an instance

miyamotoh:
Adding logic/code path to support `clone_to_template` provisioning operation. Happy path mostly works at this point.

To start, you'd be in the details page of a PowerVS instance, then go to Lifecycle > Publish to a Template. The following page opens;
image

Just give it a unique name and Submit. The page would switch to Services > Requests, and the status updates for the image-capture operation would come in as Notifications like below;
image

To control which provider/instance should get this menu action shown, it needs the below PR to disable `:publish` feature in the base class and have yours enable it;
https://github.com/ManageIQ/manageiq/pull/21872

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -4,6 +4,90 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
   end
 
   def prepare_for_clone_task
+    request_type == 'clone_to_template' ? prepare_for_clone_to_template : prepare_for_clone
+  end
+
+  def start_clone(clone_options)
+    begin
+      if request_type == 'clone_to_template'
+        source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
+          vm = Vm.find(get_option(:src_vm_id))
+          body = IbmCloudPower::PVMInstanceCapture.new(clone_options)
+          response = api.pcloud_pvminstances_capture_post(cloud_instance_id, vm.uid_ems, body)
+          response[:taskID]
+        end
+      elsif sap_image?
+        source.with_provider_connection(:service => "PCloudSAPApi") do |api|
+          body = IbmCloudPower::SAPCreate.new(clone_options)
+          response = api.pcloud_sap_post(cloud_instance_id, body)
+          response&.first&.pvm_instance_id
+        end
+      else
+        source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
+          body = IbmCloudPower::PVMInstanceCreate.new(clone_options)
+          response = api.pcloud_pvminstances_post(cloud_instance_id, body)
+          response&.first&.pvm_instance_id
+        end
+      end
+    rescue RestClient::ExceptionWithResponse => e
+      raise MiqException::MiqProvisionError, e.response.to_s
+    end
+  rescue RestClient::ExceptionWithResponse => e
+    raise MiqException::MiqProvisionError, e.response.to_s
+  end
+
+  def do_clone_task_check(clone_task_ref)
+    if request_type == 'clone_to_template'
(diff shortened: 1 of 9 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -4,57 +4,71 @@ def log_clone_options(clone_options)
   end
 
   def prepare_for_clone_task
-    specs = {
-      'image_id'   => get_option_last(:src_vm_id),
-      'pin_policy' => get_option_last(:pin_policy),
-    }
+    if request_type == 'clone_to_template'
```
agrare, line 7: @miyamotoh wonder if we'd be better off splitting these out into two methods rather than having 90% of the code under a conditional?
Assuming we validate the request_type is clone_to_template or clone_to_vm we could even `send("prepare_for_#{request_type}_task")`

miyamotoh: Agreed, @agrare . I'll try to make it that way.

iv1111:
You've added a log file in your PR :-) That'd explain 6K changes.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
miyamotoh: @agrare please take a look at my attempt in the latest push. Basically I tried to contain what you suggest in my provider code, as I wasn't sure if I wanted to change it in the core/parent, without worrying about all the providers I don't touch. The git diff looks a bit big, but basically I'm moving the logic to `private` funcs at the bottom of the file. Give me a guidance for a better implementation.

agrare:
Kicking tests after core PR merged

miyamotoh:
@agrare What kind of new tests do you see to add for this? I was thinking of just capturing a VM via Terraform and making sure we see it as a template in the DB. Thoughts?

agrare:
@miyamotoh there is an existing `spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow_spec.rb` spec, I would move that into a "clone" context and add a new "clone_to_template" context and test the workflow/dialogs/anything else that changed

agrare (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -4,6 +4,90 @@ def log_clone_options(clone_options)
...
+          response = api.pcloud_sap_post(cloud_instance_id, body)
+          response&.first&.pvm_instance_id
+        end
+      else
+        source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
+          body = IbmCloudPower::PVMInstanceCreate.new(clone_options)
+          response = api.pcloud_pvminstances_post(cloud_instance_id, body)
+          response&.first&.pvm_instance_id
+        end
+      end
```
agrare, line 31: NOTE I do think that we can clean this up by splitting into methods rather than all in this one conditional (basically what codeclimate is complaining about) but can be a follow-up

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -4,6 +4,90 @@ def log_clone_options(clone_options)
...
+          status = "The server has been provisioned.; #{stop ? 'Server description available.' : 'Waiting for server description.'}"
+        when 'ERROR'
+          raise MiqException::MiqProvisionError, _("An error occurred while provisioning the instance.")
+        else
+          status = "Unknown server state received from the cloud API: '#{instance_state}'"
+          _log.warn(status)
+        end
+
+        return stop, status
+      end
```
agrare, line 67: Same comment here, we essentially have two totally different methods in one here

PR #379, opened 2022-05-16
Fix powervs vm provision ipv4 field

jaywcarman:
I believe this has been broken since 815d06d when I changed the order of assigning `specs['networks']` and `specs['networks'][0]['ipAddress']`.

Also, the field to set the ipv4 address for a network is _not_ required.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -28,10 +28,6 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
       specs['sys_type']      = get_option_last(:sys_type)
     end
 
-    # TODO: support multiple values
-    ip_addr = get_option_last(:ip_addr)
-    specs['networks'][0]['ipAddress'] = ip_addr unless !ip_addr || ip_addr.strip.blank?
-
     user_script_text = options[:user_script_text]
     user_script_text64 = Base64.encode64(user_script_text) unless user_script_text.nil?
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -49,6 +45,10 @@ def prepare_for_clone_task
     attached_networks.concat(phase_context[:new_networks]).compact!
     specs['networks'] = attached_networks
 
+    # TODO: support multiple values
+    ip_addr = get_option_last(:ip_addr)
+    specs['networks'][0]['ipAddress'] = ip_addr unless !ip_addr || ip_addr.strip.blank?
```
agrare, line 50: Realize this is just a move, but pairing `unless` with `!` and a second boolean condition always causes my brain throws a fault

Could you achieve the same thing with
```suggestion
    specs['networks'][0]['ipAddress'] = ip_addr if ip_addr.present?
```

jaywcarman: > always causes my brain throws a fault

😆. Fixed

iv1111 (approved):

agrare (approved):

Fryguy:
Backported to `najdorf` in commit fec7a007036aa08662c7de82ed70e2edbf61190d.

```text
commit fec7a007036aa08662c7de82ed70e2edbf61190d
Date:   Tue May 17 08:39:18 2022 -0400

    Merge pull request #379 from jaywcarman/fix_powervs_vm_provision_ipv4_field
    
    Fix powervs vm provision ipv4 field
    
    (cherry picked from commit 94b4327f42da9a2bd523d7e469bd30b6d94019cd)
```

PR #382, opened 2022-05-17
terraform files

Kuldip-Nanda:
terraform files


lib/tasks_private/terraform/common.rb
```diff
@@ -0,0 +1,83 @@
...
+      :power_instance_id   => config[:secret]["cloud_instance_id"],
+      :ibmcloud_region     => config[:secret]["ibmcloud_region"],
+      :vm_name             => config[:setting]["vm_name"],
+      :volume_name         => config[:setting]["volume_name"],
+      :key_pair_name       => config[:setting]["key_pair_name"],
+      :image_name          => config[:setting]["image_name"],
+      :sys_type            => config[:setting]["sys_type"],
+      :public_network_name => config[:setting]["public_network_name"],
+      :power_network_name  => config[:setting]["power_network_name"]
+    }
```
agrare, line 47: Codeclimate has a good point these appear to be duplicated, could we extract this to a method like `plan_vars` and then we can do `:vars => plan_vars`?

lib/tasks_private/terraform/common.rb
```diff
@@ -0,0 +1,77 @@
...
+    :ibmcloud_region     => config[:secret]["ibmcloud_region"],
+    :vm_name             => config[:setting]["vm_name"],
+    :volume_name         => config[:setting]["volume_name"],
+    :key_pair_name       => config[:setting]["key_pair_name"],
+    :image_name          => config[:setting]["image_name"],
+    :sys_type            => config[:setting]["sys_type"],
+    :public_network_name => config[:setting]["public_network_name"],
+    :power_network_name  => config[:setting]["power_network_name"]
+  }
+  settings
```
agrare, line 44: Hey you actually don't need to set a local variable if you just want to return the thing, ruby will return the last return anyway and declaring the hash will return itself:
```suggestion
  {
    :ibmcloud_api_key    => config[:secret]["api_key"],
    :power_instance_id   => config[:secret]["cloud_instance_id"],
    :ibmcloud_region     => config[:secret]["ibmcloud_region"],
    :vm_name             => config[:setting]["vm_name"],
    :volume_name         => config[:setting]["volume_name"],
    :key_pair_name       => config[:setting]["key_pair_name"],
    :image_name          => config[:setting]["image_name"],
    :sys_type            => config[:setting]["sys_type"],
    :public_network_name => config[:setting]["public_network_name"],
    :power_network_name  => config[:setting]["power_network_name"]
  }
```

lib/tasks_private/terraform/placement_group/main.tf
```diff
@@ -0,0 +1,51 @@
...
+    pi_proc_type          = "${var.proc_type}"
+    pi_sys_type           = "${var.sys_type}"
+    pi_cloud_instance_id  = "${var.power_instance_id}"
+    #pi_cloud_instance_id = "888f9bf2-ab12-48c4-9dc5-8da65de951a8"
+    pi_image_id           = "${data.ibm_pi_image.power_images.id}"
+    pi_pin_policy         = "${var.pin_policy}"
+    pi_key_pair_name      = "${var.key_pair_name}"
+    pi_storage_type       = "tier3"
+    pi_volume_ids         = ibm_pi_volume.power_volume.*.volume_id
+    pi_health_status      = "WARNING"
```
agrare, line 43: Sorry if I'm missing something, but this looks like you're intentionally creating an instance with a health_status of warning?

lib/tasks_private/terraform/placement_group/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+}
+
+variable "vm_name" {
+    description = "Name of the VM"
+    type = string
+}
+
+variable "power_instance_id" {
+    description = "Power Virtual Server instance ID associated with your IBM Cloud account (note that this is NOT the API key)"
+    default =  "888f9bf2-ab12-48c4-9dc5-8da65de951a8"
```
agrare, line 18: This isn't going to be true for someone with different credentials trying to run this right?  Looks like something that should be part of the config settings/secrets?

lib/tasks_private/terraform/provision/outputs.tf
```diff
@@ -0,0 +1,11 @@
+output "status" {
+    value = "${ibm_pi_instance.pvminstance.status}"
+}
+
+#output "ip_address" {
+#    value = "${ibm_pi_instance.pvminstance.addresses}"
+#}
```
agrare, line 7: Is there a reason this is left commented out?  Should this just be deleted?

lib/tasks_private/terraform/provision/variables.tf
```diff
@@ -16,8 +15,7 @@ variable "vm_name" {
 
 variable "power_instance_id" {
     description = "Power Virtual Server instance ID associated with your IBM Cloud account (note that this is NOT the API key)"
-    #default = "8fa27c40-827c-4568-8813-79b398e9cd27"
-    default = "749e3492-1ff4-4d45-b43c-513674930661"
+    default =  "888f9bf2-ab12-48c4-9dc5-8da65de951a8"
```
agrare, line 18: Same comment re: should this be hard-coded to this

lib/tasks_private/terraform/common.rb
agrare: Ha I beat rubocop to it :wink: https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/382#issuecomment-1130235110

lib/tasks_private/terraform/provision/variables.tf
Kuldip-Nanda: I was building it without the rake wrapper. These hardcoding are left overs. I can. remove them.

lib/tasks_private/terraform/placement_group/main.tf
Kuldip-Nanda: Yes, that is the established procedure. I will later post the URL which refers to it.

agrare (approved):
LGTM
@Kuldip-Nanda can you add a high-level description of what you're changing here and why just so it is more obvious what this change is for when we come back to it in the future?

Issue #384, opened 2022-05-26
Manual refresh of the VPC NetworkManager fails


agrare:
```
MIQ(ManageIQ::Providers::BaseManager::Refresher#refresh) Refreshing all targets...
MIQ(ManageIQ::Providers::BaseManager::Refresher#refresh) EMS: [cmh-vpc Network Manager], id: [12] Refreshing targets for EMS..."}
MIQ(ManageIQ::Providers::BaseManager::Refresher#refresh) EMS: [cmh-vpc Network Manager], id: [12]   ManageIQ::Providers::IbmCloud::VPC::NetworkManager [cmh-vpc Network Manager] id [12]"}
MIQ(ManageIQ::Providers::BaseManager::Refresher#refresh) EMS: [cmh-vpc Network Manager], id: [12] Refresh failed
[NoMethodError]: undefined method `storage_manager' for #<ManageIQ::Providers::IbmCloud::VPC::NetworkManager:0x000055c2a10355c0>
Did you mean?  storage_ids  Method:[block (2 levels) in <class:LogProxy>]
/opt/IBM/infrastructure-management-gemset/gems/activemodel-6.0.4.6/lib/active_model/attribute_methods.rb:432:in `method_missing'
/var/www/miq/vmdb/app/models/manageiq/providers/inventory/persister.rb:89:in `storage_manager'
/var/www/miq/vmdb/app/models/manageiq/providers/inventory/persister/builder/persister_helper.rb:138:in `add_collection_for_manager'
/var/www/miq/vmdb/app/models/manageiq/providers/inventory/persister/builder/persister_helper.rb:48:in `add_storage_collection'
/opt/IBM/infrastructure-management-gemset/bundler/gems/bluecf-providers-ibm_cloud-7d9cb3fc065d/app/models/manageiq/providers/ibm_cloud/inventory/persister/vpc.rb:54:in `initialize_storage_inventory_collections'
/opt/IBM/infrastructure-management-gemset/bundler/gems/bluecf-providers-ibm_cloud-7d9cb3fc065d/app/models/manageiq/providers/ibm_cloud/inventory/persister/vpc.rb:15:in `initialize_inventory_collections'
/var/www/miq/vmdb/app/models/manageiq/providers/inventory/persister.rb:18:in `initialize'
/var/www/miq/vmdb/app/models/manageiq/providers/inventory.rb:12:in `new'
[...]

PR #385, opened 2022-05-26
Add spec to show network_manager.refresh failure

agrare:
VPC (and most of the other providers) refresh all child managers together.  Until all are converted over we have to play some games like this: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister.rb#L88-L90 to get an inventory collection's manager.

If all inventory is refreshed together is is more straightforward to simply delegate refresh to the parent manager.

Will review other providers but if we like this pattern think it can apply to all that do combined refresh.

Fixes https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/384

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/network_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/network_manager.rb
@@ -28,6 +28,8 @@ class ManageIQ::Providers::IbmCloud::VPC::NetworkManager < ManageIQ::Providers::
            :hostname,
            :default_endpoint,
            :endpoints,
+           :refresh,
+           :refresh_ems,
            :to        => :parent_manager,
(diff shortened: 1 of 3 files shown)


Fryguy (approved):

Fryguy:
Backported to `najdorf` in commit 1f04d34cb76b150a89897d7dbb3a051c66750be7.

```text
commit 1f04d34cb76b150a89897d7dbb3a051c66750be7
Date:   Thu May 26 14:53:28 2022 -0400

    Merge pull request #385 from agrare/fix_network_manager_full_refresh_failure
    
    Add spec to show network_manager.refresh failure
    
    (cherry picked from commit cd0688dcc60f7a6f14d25bbefd2ee91e693e597f)
```

PR #387, opened 2022-06-03
Following up for better code readability after #374

miyamotoh:
Attempting to make codeclimate happier after #374 ...

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -10,24 +10,11 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
   def start_clone(clone_options)
     begin
       if request_type == 'clone_to_template'
-        source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
-          vm = Vm.find(get_option(:src_vm_id))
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -37,34 +24,9 @@ def start_clone(clone_options)
...
-        else
-          status = "Unknown server state received from the cloud API: '#{instance_state}'"
-          _log.warn(status)
-        end
-
-        return stop, status
-      end
+    case request_type
+    when 'clone_to_template' then check_task_clone_to_template
+    else                          check_task_clone(clone_task_ref)
```
agrare, line 29: This `case/when` is a little overkill for what could be an `if / else` or ternary operator.  Recommend the way it is done in `prepare_for_clone_task` above on line 7
```suggestion
    request_type == 'clone_to_template' ? check_task_clone_to_template : check_task_clone(clone_task_ref)
```

agrare:
Just that one comment then this LGTM

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
miyamotoh: Agreed, as I was thinking of more `when`'s to come, but am not sure if they will come. Force-pushed.

agrare (approved):

PR #388, opened 2022-06-07
Added volume affinity parametes for affinityPVMInstance & antiAffinityPVMInstances

vinisawa:
Added volume affinity parametes for affinityPVMInstance & antiAffinityPVMInstances

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -6,13 +6,13 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
   end
   supports_not :snapshot_create
   supports_not :update
-  supports :attach do
-    unsupported_reason_add(:attach, _("the volume is not connected to an active Provider")) unless ext_management_system
-    unsupported_reason_add(:attach, _("cannot attach non-shareable volume that is in use.")) if status == "in-use" && !multi_attachment
+  supports :attach_volume do
+    unsupported_reason_add(:attach_volume, _("the volume is not connected to an active Provider")) unless ext_management_system
+    unsupported_reason_add(:attach_volume, _("cannot attach non-shareable volume that is in use.")) if status == "in-use" && !multi_attachment
   end
-  supports :detach do
-    unsupported_reason_add(:detach, _("the volume is not connected to an active Provider")) unless ext_management_system
-    unsupported_reason_add(:detach, _("the volume status is '%{status}' but should be 'in-use'") % {:status => status}) unless status == "in-use"
+  supports :detach_volume do
+    unsupported_reason_add(:detach_volume, _("the volume is not connected to an active Provider")) unless ext_management_system
+    unsupported_reason_add(:detach_volume, _("the volume status is '%{status}' but should be 'in-use'") % {:status => status}) unless status == "in-use"
   end
 
   def available_vms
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -6,13 +6,13 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudV
...
+  supports :attach_volume do
+    unsupported_reason_add(:attach_volume, _("the volume is not connected to an active Provider")) unless ext_management_system
+    unsupported_reason_add(:attach_volume, _("cannot attach non-shareable volume that is in use.")) if status == "in-use" && !multi_attachment
   end
-  supports :detach do
-    unsupported_reason_add(:detach, _("the volume is not connected to an active Provider")) unless ext_management_system
-    unsupported_reason_add(:detach, _("the volume status is '%{status}' but should be 'in-use'") % {:status => status}) unless status == "in-use"
+  supports :detach_volume do
+    unsupported_reason_add(:detach_volume, _("the volume is not connected to an active Provider")) unless ext_management_system
+    unsupported_reason_add(:detach_volume, _("the volume status is '%{status}' but should be 'in-use'") % {:status => status}) unless status == "in-use"
```
jaywcarman, line 15: The attach / detach features no long need the `_volume` suffix. See #373 for details

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -101,6 +101,42 @@ def self.params_for_create(ems)
...
+          :name         => 'affinityPVMInstance',
+          :id           => 'affinityPVMInstance',
+          :label        => _('Cloud Volume Type'),  
+          :validate     => [{:type => 'required'}],
+          :condition    => {
+            :when    => 'affinity_policy',
+            :pattern => '^Off$',
+          },
+          :includeEmpty => true,
+          :options      => ems.vms.map do |cvt|
```
jaywcarman, line 115: The `ems` here is actually a `StorageManager` and VMs are owned by a `CloudManager` so this won't work. You'll have to do something like this:
```ruby
          :options      => ems.parent_manager.vms.map do |vm|
            {                             
              :label => vm.name,                
              :value => vm.ems_ref,
            }
          end,
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -101,6 +101,42 @@ def self.params_for_create(ems)
...
+        {
+          :component    => 'select',
+          :name         => 'affinityPVMInstance',
+          :id           => 'affinityPVMInstance',
+          :label        => _('Cloud Volume Type'),  
+          :validate     => [{:type => 'required'}],
+          :condition    => {
+            :when    => 'affinity_policy',
+            :pattern => '^Off$',
+          },
```
jaywcarman, line 113: There needs to be an additional field to select affinity type: Volume or PVM Instance. Then the volume and VM selection fields can be conditional on the affinity type selected.

jaywcarman (commented):
One more review comment. Once the form is updated the [`raw_create_volume` class method](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb#L108-L130) still needs to be updated to consume the VM/Volume affinity parameters.

jaywcarman:
Opened #407 to finish this change.

PR #389, opened 2022-06-07
Improve PowerVS "Native Console" URL

jaywcarman:
The IBM Cloud console URL for a specific PowerVS service instance can be
constructed using its CRN. This is much more convenient than dropping
the user into the default cloud.ibm.com interface where they'll have to
find and select the service instance they're looking for.

I verified this URL works with or without an existing cloud.ibm.com session (in the without case, the user will have go through the standard auth check before arriving at the PowerVS service web console).

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
@@ -54,7 +54,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
   supports_not :volume_availability_zones
 
   def console_url
-    "https://cloud.ibm.com"
+    "https://cloud.ibm.com/services/power-iaas/#{CGI.escape(pcloud_crn.values.join(":"))}"
   end
 
   def image_name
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb
```diff
@@ -54,7 +54,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager < ManageI
   supports_not :volume_availability_zones
 
   def console_url
-    "https://cloud.ibm.com"
+    "https://cloud.ibm.com/services/power-iaas/" + self.pcloud_crn.values.join("%3A").gsub("/","%2F")
```
agrare, line 57: Couple of things,

1. String interpolation is preferred over string addition so `"https://cloud.ibm.com/services/power-iaas/#{"https://cloud.ibm.com/services/power-iaas/}" (there should be a rubocop warning about this but I didn't check)
2. The extra `self.` is usually redundant so just `pcloud_crn.values` should work (the most common exception here is assigning an instance variable so as to differentiate from a local variable)
3. Rather than using the URI escape chars I think it'd be more readable (and safer if there are any other than `:` and `/`) to use the real characters then CGI.escape the path

So putting it all together something like:
```suggestion
    "https://cloud.ibm.com/services/power-iaas/#{CGI.escape(pcloud_crn.values.join(":"))}"
```

agrare (approved):

PR #390, opened 2022-06-08
Add cloud volume feature

jdenver23:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21923

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
@@ -1,5 +1,6 @@
 class ManageIQ::Providers::IbmCloud::PowerVirtualServers::StorageManager::CloudVolume < ::CloudVolume
   supports :create
+  supports :clone
   supports :delete do
     unsupported_reason_add(:delete, _("the volume is not connected to an active Provider")) unless ext_management_system
     unsupported_reason_add(:delete, _("cannot delete volume that is in use.")) if status == "in-use"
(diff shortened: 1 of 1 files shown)


agrare:
Hey @jdenver23 it looks like you merged master into this branch rather than rebase this branch onto master.  Recommend using `git config --global --add pull.rebase true` so that `git pull upstream master` will rebase your branch onto master without any extra merge commits.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -105,6 +106,20 @@ def self.params_for_create(ems)
...
   end
 
+  def params_for_clone
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Volume Name'),
```
jaywcarman, line 116: I would prefer to call this `_('Volume Base Name')` and then add some "helperText" so the user can determine what the cloned volume name will be.
```ruby
:helperText => _("Base name of the new cloned volume(s). Cloned Volume names will be prefixed with 'clone-' and suffixed with '-#####' (where ##### is a 5 digit random number)")
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -161,4 +176,22 @@ def raw_detach_volume(vm_ems_ref)
...
+      )
+      api.pcloud_v2_volumes_clone_post(
+        cloud_instance_id,
+        clone_volume_params
+      )
+    end
+  rescue => e
+    _log.error("volume=[#{name}], error: #{e}")
+    # Raise Correct Error
+    raise MiqException::MiqVolumeDeleteError, e.to_s, e.backtrace
```
jaywcarman, line 195: Either need to define a `MiqException::MiqVolumeCloneError` class or raise something more generic.

jaywcarman: I think this is where you would add it:
https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-exception.rb#L119

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/storage_manager/cloud_volume.rb
```diff
@@ -105,6 +106,21 @@ def self.params_for_create(ems)
...
+  def params_for_clone
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Volume Base Name'),
+          :isRequired => true,
+          :helperText => _("Base name of the new cloned volume(s). Cloned Volume names will be prefixed with 'clone-' and suffixed with '-#####' (where ##### is a 5 digit random number)"),
```
jaywcarman, line 118: Minor, but through this interface we can only clone a single volume so it would be clearer to keep the help text singular. (in my earlier review I lazily copy/pasted the text from the [Power Cloud API docs](https://cloud.ibm.com/apidocs/power-cloud#pcloud-v2-volumes-clone-post)).
```suggestion
          :helperText => _("Base name of the new cloned volume. The cloned Volume name will be prefixed with 'clone-' and suffixed with '-#####' (where ##### is a 5 digit random number)"),
```

jaywcarman (approved):
LGTM! I'm good to merge after the remaining core PRs are merged:
- https://github.com/ManageIQ/manageiq-api/pull/1164
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8301
- https://github.com/ManageIQ/manageiq-gems-pending/pull/545

agrare (approved):

PR #393, opened 2022-06-13
Fix task reference in VM provision state machine

jaywcarman:
When calling 'check_task_clone' to check the status of a newly
provisioned VM, the VM ID can be accessed through
'phase_context[:clone_task_mor]'. I observed that 'clone_task_ref' is
nil. This was working in the past - not sure if it has been broken for a
while, was broken by the most recent change, 0732f69, or if something
changed in one of the core repos.

Found this after noticing that a PowerVS VM provision request resulted in an error even though the PowerVS service showed the VM had been successfully created. EVM logs showed:
```
[----] E, [2022-06-13T16:24:01.106516 #266244:aaa0] ERROR -- evm: MIQ(ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provision#provision_error) [[ArgumentError]: Missing the required parameter 'pvm_instance_id' when calling            PCloudPVMInstancesApi.pcloud_pvminstances_get] encountered during phase [poll_clone_complete]                                                                                                                                                         
[----] E, [2022-06-13T16:24:01.106648 #266244:aaa0] ERROR -- evm: /home/jwcarman/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/ibm_cloud_power-1.1.1/lib/ibm_cloud_power/api/p_cloud_pvm_instances_api.rb:404:in `pcloud_pvminstances_get_with_http_info'
```

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
@@ -42,10 +42,10 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
 
   def start_clone_task
     update_and_notify_parent(:message => "Starting Clone of #{clone_direction}")
-
     clone_options = prepare_for_clone_task
     log_clone_options(clone_options)
-    phase_context[:clone_task_mor] = start_clone(clone_options)
+    phase_context[:clone_task_ref] = start_clone(clone_options)
+    phase_context.delete(:clone_options)
     signal :poll_clone_complete
   end
 end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -8,13 +8,13 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
   end
 
   def start_clone(clone_options)
-      if request_type == 'clone_to_template'
-        make_request_clone_to_template(clone_options)
-      elsif sap_image?
-        make_request_clone_sap_vm(clone_options)
-      else
-        make_request_clone(clone_options)
-      end
+    if request_type == 'clone_to_template'
+      make_request_clone_to_template(clone_options)
+    elsif sap_image?
+      make_request_clone_sap_vm(clone_options)
+    else
+      make_request_clone(clone_options)
+    end
   rescue IbmCloudPower::ApiError => err
     error_message = JSON.parse(err.response_body)["description"] || err.message
     _log.error("VM start_clone error: #{error_message}")
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -22,7 +22,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
   end
 
   def do_clone_task_check(clone_task_ref)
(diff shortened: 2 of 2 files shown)


agrare:
Interesting, `phase_context[:clone_task_ref]` is set to the return of `start_clone`, I don't see `clone_task_mor` set by any provider other than VMware Infra

jaywcarman:
Thanks for taking a look @agrare. Yeah I'm not sure where it's coming from. I noticed that @miyamotoh used it in the template provisioning [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/commit/0732f69981770d1baf3464f59517cce3a14ad24b#diff-ad8b9c47e7a1a709c3685c42f1771ff4169a4fd28bae7e2ec38c4f1e40577a3eR125). @miyamotoh - any ideas?

For what it's worth, I have tested this patch and it works 🤷🏻‍♂️

agrare:
Ohh, I had an out of date local ibm_cloud repo.  The problem is this override https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/d5eadd0f656c9f62654213ca9254e134baebe726/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb#L48

This looks like mostly a duplicate of the core method I think the answer is to delete that method override completely

jaywcarman:
> This looks like mostly a duplicate of the core method I think the answer is to delete that method override completely

Nice find! I'll update this PR (have to log off now, so won't push until later). Also glad to see this isn't broken in Najdorf.

jaywcarman:
From the 0e0a7f8 commit message:

Use ':clone_task_ref' in provision state machine

Following the `ManageIQ::Providers::CloudManager::Provision::StateMachine` base module, use `:clone_task_ref` instead of `:clone_task_mor`. The `start_clone_task` override could be removed completely if it weren't for the custom `clone_options = prepare_for_clone_task` portion.

jaywcarman:
fyi @agrare this one was sitting in WIP for a _long_ time but it now ready for review.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
```diff
@@ -42,10 +42,10 @@ def determine_placement
 
   def start_clone_task
     update_and_notify_parent(:message => "Starting Clone of #{clone_direction}")
-
     clone_options = prepare_for_clone_task
     log_clone_options(clone_options)
-    phase_context[:clone_task_mor] = start_clone(clone_options)
+    phase_context[:clone_task_ref] = start_clone(clone_options)
+    phase_context.delete(:clone_options)
```
agrare, line 48: Why delete the clone options here?

agrare (approved):
One question but nothing serious, it just looked like we could delete the whole method if it was a dup of core

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
jaywcarman: Just following what's in the base module:
https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/provision/state_machine.rb#L38

jaywcarman:
> One question but nothing serious, it just looked like we could delete the whole method if it was a dup of core

It's not an example dup of core, this is custom:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb#L45

Fryguy:
Backported to `oparin` in commit 0300b7ab9ef54269b34beb0de4249234248aeed8.

```text
commit 0300b7ab9ef54269b34beb0de4249234248aeed8
Date:   Wed Sep 7 15:39:03 2022 -0400

    Merge pull request #393 from jaywcarman/fix_powervs_check_task_clone
    
    Fix task reference in VM provision state machine
    
    (cherry picked from commit c6f0cbde54130c6a6f3cf63702bdad4f7714b634)
```

PR #395, opened 2022-06-14
Adding HTML5 Console support for cloud instance with noVNC

miyamotoh:
This adds Access > VM Console menu action which would launch the "noVNC" console window/tab upon click. Not supported if VM is not powered on, is archived, or orphaned, for which specs are added.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -25,6 +25,14 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
     unsupported_reason_add(:publish, reason) if reason
   end
 
+  supports :html5_console do
+    reason   = _("VM Console not supported because VM is not powered on") unless current_state == "on"
+    reason ||= _("VM Console not supported because VM is orphaned")       if orphaned?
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -92,6 +100,10 @@ def self.calculate_power_state(raw_power_state)
     end
   end
 
+  def console_supported?(type)
+    return true if type.upcase == 'VNC'
```
agrare, line 104: You can actually just do `type.upcase == 'VNC'`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -50,4 +50,32 @@ def raw_remove_all_snapshots
...
+      :instance_id => id,
+      :class_name  => self.class.name,
+      :method_name => 'remote_console_acquire_ticket',
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => 'ems_operations',
+      :args        => [protocol]
+    }
+
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
```
agrare, line 69: Follow-up but I think we should move this up to the base class since this method appears to be duplicated in a number of providers.

agrare (approved):

PR #396, opened 2022-06-16
Adding "Management Console" access to open the VM's page on IBM Cloud

miyamotoh:
Adding Access > Native Console action for a PowerVS VM. Very hacky (and thus WIP and likely change), but can open the VM's page on IBM Cloud.

UI PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/8319

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -33,6 +33,12 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   end
   supports :launch_html5_console
 
+  supports :native_console do
+    reason ||= _("VM Console not supported because VM is orphaned") if orphaned?
+    reason ||= _("VM Console not supported because VM is archived") if archived?
+    unsupported_reason_add(:native_console, reason) if reason
+  end
+
   def cloud_instance_id
     ext_management_system.uid_ems
   end
@@ -102,6 +108,12 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
 
   def console_supported?(type)
     return true if type.upcase == 'VNC'
+  end
+
+  def console_url
+    crn = ERB::Util.url_encode(ext_management_system.pcloud_crn.values.join(":"))
+    params = URI.encode_www_form(:paneId => "manageiq", :crn => crn)
+    URI::HTTPS.build(:host => "cloud.ibm.com", :path => "/services/power-iaas/#{crn}/server/#{uid_ems}", :query => params)
   end
 
   private
--- a/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm_spec.rb
+++ b/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm_spec.rb
@@ -60,4 +60,26 @@ describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm do
       expect(vm.supports?(:html5_console)).to be_falsey
     end
   end
+
+  context "supports management console access?" do
+    it 'supports console access if powered on' do
+      vm.update(:raw_power_state => power_state_on)
+      expect(vm.supports?(:native_console)).to be_truthy
+    end
+
+    it 'no console access if powered off' do
+      vm.update(:raw_power_state => power_state_suspended)
+      expect(vm.supports?(:native_console)).to be_truthy
+    end
+
+    it 'no console access if orphaned' do
+      vm.update(:ems_id => nil)
+      expect(vm.supports?(:native_console)).to be_falsey
+    end
+
+    it 'no console access if archived' do
+      vm.update(:ems_id => nil, :storage_id => nil)
+      expect(vm.supports?(:native_console)).to be_falsey
+    end
+  end
 end


agrare:
@miyamotoh are you trying to open the VM's remote console or the details/dashboard page?

miyamotoh:
@agrare What I'm trying to open is a page on IBM Cloud (outside ManageIQ), showing the meta/details of the VM.

agrare:
@miyamotoh yeah so the methods you're implementing are intended for "Remote Console" no the dashboard/details page.

For the VM dashboard you just need:
```
supports :native_console

def console_url
  "https://cloud.ibm.com/services/power-iaas/#{crn}/server/#{uid_ems}/?paneId=manage&crn=#{crn}"
end
```

I didn't realize until very recently that Ovirt Remote Console had methods _also_ called native_console when we were doing the PowerVS "open the native provider dashboard" work.

miyamotoh:
@agrare if that's all I need on the provider side, that's great, but what about the UI side? I wouldn't get the "Native Console" menu item nor any standalone button on the page. How does it get picked up by UI?

agrare:
@miyamotoh correct but this is the provider PR :wink: I'll comment in the UI PR about what I think needs to happen there

Fryguy:
Updated the OP to cross-link the PRs

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -104,6 +107,15 @@ def console_supported?(type)
     return true if type.upcase == 'VNC'
   end
 
+  def console_url
+    crn = CGI.escape(ext_management_system.pcloud_crn.values.join(":"))
```
Fryguy, line 111: As annoying as this is, `CGI.escape` should not be used because it doesn't handle things correctly...prefer `ERB::Util.url_encode`.

Aside, I'm tempted to put this in more_core_extensions so we don't have to think about this as much in the future.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -104,6 +107,15 @@ def console_supported?(type)
...
+    "https://cloud.ibm.com/services/power-iaas/#{crn}/server/#{uid_ems}/?paneId=manage&crn=#{crn}"
```
Fryguy, line 112: Actually, I just realized you are using this for building the query part of the URL...in that case, I think you should use `URI.encode_www_form`:

```suggestion
    params = URI.encode_www_form(:paneId => "manageiq", :crn => ext_management_system.pcloud_crn.values.join(":"))
    "https://cloud.ibm.com/services/power-iaas/#{crn}/server/#{uid_ems}/?params"
```

miyamotoh: Feedback incorporated, force-push'd. PTAL.

Fryguy:
Looks nice - just need to make sure the caller of this can handle a URI object or String being returned (they can just do `.to_s`)

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -33,6 +33,13 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   end
   supports :launch_html5_console
 
+  supports :native_console do
+    reason ||= _("VM Console not supported because VM is orphaned") if orphaned?
+    reason ||= _("VM Console not supported because VM is archived") if archived?
+    unsupported_reason_add(:native_console, reason) if reason
+  end
```
agrare, line 40: I don't believe this is correct, when you're opening the "management console" of a VM it can be powered off

agrare, line 40:
```suggestion
  supports :native_console
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -33,6 +33,13 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
...
+  supports :launch_native_console
```
agrare, line 41: I don't believe this feature is what you're adding here (aka vm remote console)
```suggestion
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -104,6 +111,16 @@ def console_supported?(type)
...
 
+  def console_url
+    crn = ERB::Util.url_encode(ext_management_system.pcloud_crn.values.join(":"))
+    params = URI.encode_www_form(:paneId => "manageiq", :crn => crn)
+    URI::HTTPS.build(:host => "cloud.ibm.com", :path => "/services/power-iaas/#{crn}/server/#{uid_ems}", :query => params)
+  end
+
+  def validate_native_console_support
+    true
+  end
```
agrare, line 122: This is also related to the "remote console" type of support and is not necessary
```suggestion
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
miyamotoh: Powered off VM can still open Management Console, it's `supported` by the above, no?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
jaywcarman: I recently used `CGI.escape` [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager.rb#L57). Sounds like it should be avoided in general?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
miyamotoh: What I try to do here is to hide the menu if we lost track of the VM's EMS or otherwise considered "archived" (=offline?) in ManageIQ?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
miyamotoh: I was somehow under the impression that's how it decides to show or route the menu, but you are right. It works without it.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
miyamotoh: Good catch. Dropping...

PR #409, opened 2022-07-28
Fix PowerVS inventory refresh when volumes are missing

jaywcarman:
I encountered a PowerVS service instance that had VMs with references to volumes that no longer existed. The VMs were in fail state, so perhaps the resource relationships were stale. In any case, this caused inventory refresh to fail completely.

We've seen similar circumstances, so this PR is targeting consistency across all other resources in the PowerVS collection.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
@@ -74,6 +74,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Parser::PowerVirtualServers < Ma
       # saving instance disk information
       instance.volume_i_ds.to_a.each do |vol_id|
         volume = collector.volume(vol_id)
+        next if volume.nil?
 
         persister.disks.build(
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -75,15 +75,17 @@ def pvm_instances
...
-        persister.disks.build(
-          :hardware        => ps_hw,
-          :device_name     => volume.name,
-          :device_type     => volume.disk_type,
-          :controller_type => "ibm",
-          :backing         => persister.cloud_volumes.find(vol_id),
-          :location        => vol_id,
-          :size            => volume.size&.gigabytes
-        )
+        unless volume.nil?
```
jaywcarman, line 78: Since @miq-bot doesn't approve of this, instead I could do:
```suggestion
        next if volume.nil?
```

agrare: :+1: yes that is the preferred style

agrare (approved):

PR #363, opened 2022-03-30
Server Placement group

Kuldip-Nanda:
Show the placement group under `pinning` in the customization tab dialog during provisioning a powervs instance.
This is a dependent PR for adding and enabling Server Placement Group as Menu in the Core

1. Manageiq schema changes.
    https://github.com/ManageIQ/manageiq-schema/pull/644
2. Server Placement Group in ManageIQ UI Changes.
    https://github.com/ManageIQ/manageiq-ui-classic/pull/8214
3. PowerVS changes.
    https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/363
4. ManageIQ Core changes
    https://github.com/ManageIQ/manageiq/pull/21807

PR #411, opened 2022-08-03
Remove placement groups from PowerVS advanced settings

jaywcarman:
Follow up to #363 

Depends on:
- ManageIQ/manageiq-ui-classic#8214


spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb
```diff
@@ -133,6 +133,7 @@ def assert_specific_vm
         :description      => "PVM Instance",
         :vendor           => "ibm_power_vs",
         :power_state      => "on",
+        :placement_group  => nil,
```
jaywcarman, line 136: Would be nice to also include specs for a VM with an associated placement group.

jaywcarman: Done in cc7f886d

agrare (approved):

PR #413, opened 2022-08-11
Replace persister.cloud_volumes.find with .lazy_find

agrare:
The `#find` method on an inventory_collection introduces an ordering dependency and should be replaced with a `#lazy_find`


jaywcarman (approved):
LGTM, thanks @agrare

PR #420, opened 2022-09-16
Refactoring vm resize option to work with ddf

alizapeikes:
Part of https://github.com/ManageIQ/manageiq-ui-classic/issues/7603

Depends on:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8452

Connected to:
https://github.com/ManageIQ/manageiq-api/pull/1181
https://github.com/ManageIQ/manageiq/pull/22122

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
@@ -130,4 +130,58 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
       $ibm_cloud_log.info("VM instance #{instance.id} state is #{raw_power_state}")
     end
   end
+
+  def params_for_resize
+    {
+      :fields => [
+        {
+          :component  => 'text-field',
+          :name       => 'name',
+          :id         => 'name',
+          :label      => _('Name'),
+          :isRequired => true,
+          :validate   => [
+            {
+              :type => 'required',
+            },
+            {
+              :type    => 'pattern',
+              :pattern => '^[a-zA-Z][a-zA-Z0-9_-]*$',
+              :message => _('Must contain only alphanumeric, hyphen, and underscore characters'),
+            }
+          ],
+        },
+        {
+          :component => 'textarea',
+          :name      => 'description',
+          :id        => 'description',
+          :label     => _('Description'),
+        },
+      ],
+    }
+  end
+
+  def resize_queue(userid, options = {})
+    _log.info("in resize_queue method")
+    $log.info("in resize_queue method")
+    task_opts = {
+      :action => "Resizing vm for #{userid}",
+      :userid => userid
+    }
+
+    queue_opts = {
+      :class_name  => self.class.name,
+      :method_name => 'resize',
+      :instance_id => id,
+      :role        => 'ems_operations',
+      # :queue_name  => ext_management_system.queue_name_for_ems_operations,
+      # :zone        => ext_management_system.my_zone,
+      :args        => [options]
+    }
+    MiqTask.generic_action_with_callback(task_opts, queue_opts)
+  end
+
+  def resize(options = {})
+    raw_resize(options)
+  end 
 end
--- a/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
@@ -112,22 +112,12 @@ class ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm < ManageIQ::Providers
     unsupported_reason_add(:resize, _('The VM is not powered off')) unless current_state == "off"
   end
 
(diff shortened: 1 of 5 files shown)


app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -130,4 +130,58 @@ def sdk_update_status(instance)
       $ibm_cloud_log.info("VM instance #{instance.id} state is #{raw_power_state}")
     end
   end
+
+  def params_for_resize
```
jaywcarman, line 134: Does this need to be a public method?

agrare: Yes, the API controller calls this from the object here https://github.com/ManageIQ/manageiq-api/blob/2a449e0e5e78595a03bcc165c80a896e88841b73/app/controllers/api/base_controller/renderer.rb#L557 not from an internal method

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -112,15 +112,82 @@ def raw_destroy
...
+      ],
+    }
+  end
+
+  def resize_form_options(recordId)
+    # assert_privileges("instance_resize")
+
+    # @request_id = params[:id]
+    # @record = find_record_with_rbac(VmOrTemplate, recordId)
+    @record = ManageIQ::Providers::IbmCloud::VPC::CloudManager::Vm.find_by(:id => 408)
```
agrare, line 149: This is an instance method that is called from a vm object, so you don't need to lookup the record manually.

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -112,15 +112,69 @@ def raw_destroy
...
+          :name       => 'flavor',
+          :id         => 'flavor',
+          :label      => _('Flavor'),
+          :isRequired => true,
+          :options    => resize_form_options.map do |flavor|
+            {
+              :label => flavor[:name],
+              :value => flavor[:id],
+            }
+          end,
```
agrare, line 138: If we're going to build an array of hashes down here https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/420/files#diff-5af23959f8142218f0266befea8e232991173bc7a177a5b6a24154a782cc2c2eR150 we might as well make it in the format that is being expected here:
```suggestion
          :options    => resize_form_options,
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -112,15 +112,69 @@ def raw_destroy
...
+  def resize_form_options
+    flavors = []
+    # include only flavors with root disks at least as big as the instance's current root disk.
+    self.ext_management_system&.flavors&.each do |ems_flavor|  
+      # include only flavors with root disks at least as big as the instance's current root disk.
+      if self.flavor.nil? || ((ems_flavor != self.flavor) && (ems_flavor.root_disk_size >= self.flavor.root_disk_size))
+        flavors << {:name => ems_flavor.name_with_details, :id => ems_flavor.name}
+      end
+    end
+    flavors
```
agrare, line 153: Few things:
1. In ruby it is preferred to use:
```
array.map do |item|
  {:key => item.value}
end
```

Instead of:
```
result = []
array.each do |item|
  result << {:key => item.value}
end
```

2. Since we're building an array of hashes we might as well return it in the format that is expected by `:options => ` above

3. If you use `.map` then the implicit return will work without having to explicitly return `flavors`

4. The `self.ext_management_system` is redundant (surprised miq-bot hasn't commented yet about this)

5. It looks like the logic might be a little cleaner if we invert the conditional below
```suggestion
    # include only flavors with root disks at least as big as the instance's current root disk.
    ext_management_system.flavors.map do |ems_flavor|
      # include only flavors with root disks at least as big as the instance's current root disk.
      next if flavor && (ems_flavor == flavor || ems_flavor.root_disk_size < flavor.root_disk_size)

      {:label => ems_flavor.name_with_details, :id => ems_flavor.name}
    end
```

app/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/vm.rb
```diff
@@ -112,15 +112,69 @@ def raw_destroy
     unsupported_reason_add(:resize, _('The VM is not powered off')) unless current_state == "off"
```
agrare, line 112: We should probably add `unsupported_reason_add(:resize, _('The VM is not connected to a provider')) unless ext_management_system` so that you don't have to worry about the `self.ext_management_system` being `nil` below (which I assume is why you had `self.ext_management_system&.flavors&.each`)

agrare (approved):
LGTM

alizapeikes:
I believe the test is failing because of a change made to manageIQ core which hasn't been merged.

agrare (approved):

PR #423, opened 2022-09-30
[WIP] Power vs resize feature

alizapeikes:
dependent on:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8452

connected to:
https://github.com/ManageIQ/manageiq-gems-pending/pull/549

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -137,13 +137,11 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
           :initialValue      => form_default_values.find_by(:name => 'entitled_processors').value,
           :type              => 'number',
           :min               => 0.25,
-          :max               => 6.25,
           :step              => 0.25,
           :isRequired        => true,
           :validate          => [
             {:type => 'required'},
-            {:type => 'min-number-value', :value => 0.25, :message => _('Size must be greater than or equal to .25')},
-            {:type => 'max-number-value', :value => 6.25, :message => _('Size must be lower than or equal to 6.25')}
+            {:type => 'min-number-value', :value => 0.25, :message => _("Size must be greater than or equal to .25")}
           ],
         },
         {
@@ -155,13 +153,11 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
           :initialValue      => hardware.memory_mb / 1024,
           :type              => 'number',
           :min               => 2,
-          :max               => 807,
           :step              => 1,
           :isRequired        => true,
           :validate          => [
             {:type => 'required'},
-            {:type => 'min-number-value', :value => 2, :message => _('Size must be greater than or equal to 2')},
-            {:type => 'max-number-value', :value => 807, :message => _('Size must be lower than or equal to 807')}
+            {:type => 'min-number-value', :value => 2, :message => _("Size must be greater than or equal to 2")}
           ],
 
         },
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -107,8 +107,8 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
       :fields => [
         {
           :component         => 'select',
-          :name              => 'pinning',
-          :id                => 'pinning',
+          :name              => 'pin_policy',
+          :id                => 'pin_policy',
           :label             => _('Pinning'),
           :initializeOnMount => true,
           :initialValue      => form_default_values.find_by(:name => 'pin_policy').value,
@@ -130,8 +130,8 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
         },
         {
           :component         => 'text-field',
-          :name              => 'cores',
-          :id                => 'cores',
+          :name              => 'processors',
+          :id                => 'processors',
           :label             => _('Cores'),
           :initializeOnMount => true,
           :initialValue      => form_default_values.find_by(:name => 'entitled_processors').value,
@@ -167,8 +167,8 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
         },
         {
           :component         => 'radio',
-          :name              => 'coreType',
-          :id                => 'coreType',
+          :name              => 'proc_type',
+          :id                => 'proc_type',
           :label             => _('Core Type'),
           :initializeOnMount => true,
           :initialValue      => form_default_values.find_by(:name => 'processor_type').value,
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -39,6 +39,11 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
     unsupported_reason_add(:native_console, reason) if reason
   end
 
+  supports :resize do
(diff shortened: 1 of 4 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -97,6 +102,100 @@ def params_for_create_snapshot
...
+            },
+          ]
+        },
+      ],
+    }
+  end
+
+  def form_default_values
+    @form_default_values ||= advanced_settings
+  end
```
agrare, line 197: That's slick

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -97,6 +102,100 @@ def params_for_create_snapshot
...
+        {
+          :component         => 'text-field',
+          :name              => 'cores',
+          :id                => 'cores',
+          :label             => _('Cores'),
+          :initializeOnMount => true,
+          :initialValue      => form_default_values.find_by(:name => 'entitled_processors').value,
+          :type              => 'number',
+          :min               => 0.25,
+          :max               => 6.25,
```
agrare, line 140: Is this documented anywhere or did you just see what you could pick in the PowerVS UI?
This is something we'll have to keep updated as PowerVS might allow e.g. 8.25 tomorrow. (Same for memory below).
We might want to consider moving this to `config/settings.yml` so that customers can change it before we can push a new release

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
```diff
@@ -16,6 +16,19 @@ def raw_create_snapshot(name, desc = nil, _memory = nil)
     raise MiqException::MiqVmSnapshotError, err.to_s
   end
 
+  def raw_resize(options)
+    with_provider_connection(:service => 'PCloudPVMInstancesApi') do |api|
+      body = IbmCloudPower::PVMInstanceUpdate.new("memory"     => options["memory"].to_i,
+                                                  "processors" => options["cores"].to_f,
+                                                  "proc_type"  => options["coreType"],
+                                                  "pin_policy" => options["pinning"])
```
agrare, line 24: Any reason not to name the DDF option the same as the value that `PVMInstanceUpdate` is expecting?  Why name [[this]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/423/files#diff-ee71765cda793a11b1ceeedbb4530ecefd066ddd46c797f14c75a0875f1ac60aR171) `coreType` instead of `proc_type`

agrare:
Looking really good, don't worry about the codeclimate issue DDF forms are always a bit repetitive

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm/operations.rb
alizapeikes: I had named it coreType as in the UI the label is 'Core Type', but I can change the name and id to the value that PVMInstanceUpdate expects if you think that makes more sense?

agrare: Yeah no one is ever going to see the internal name/id (that is what label is for) so we can name it anything we want

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
agrare: > in the PowerVS UI it changes everyday

:scream: that's not great, @jaywcarman what do you think about making this a setting vs not having a maximum and having the user hit a failure if they picked a number over whatever max powervs feels like that day?

jaywcarman: I think we should remove the max and let the request fail if the service doesn't have available capacity (same behavior as our VM provisioning in PowerVS).

As a future enhancement, we would want to use the [PCloudSystemPools API](https://cloud.ibm.com/apidocs/power-cloud#pcloud-systempools-get) to query core and memory capacity on the fly. I'm assuming this is what the IBM Cloud web UI is doing.

alizapeikes: @agrare @jaywcarman 
Any reason not implement this check now. This seems to work:
`  def params_for_resize
    {
      :fields => [
        {
          :component         => 'select',
          :name              => 'pin_policy',
          :id                => 'pin_policy',
          :label             => _('Pinning'),
          :initializeOnMount => true,
          :initialValue      => form_default_values.find_by(:name => 'pin_policy').value,
          :isRequired        => true,
          :options           => [
            {
              :label => 'None',
              :value => 'none',
            },
            {
              :label => 'Hard',
              :value => 'hard',
            },
            {
              :label => 'Soft',
              :value => 'soft',
            },
          ]
        },
        {
          :component         => 'text-field',
          :name              => 'processors',
          :id                => 'processors',
          :label             => _('Cores'),
          :initializeOnMount => true,
          :initialValue      => form_default_values.find_by(:name => 'entitled_processors').value,
          :type              => 'number',
          :min               => 0.25,
          :max               => 6.25,
          :step              => 0.25,
          :isRequired        => true,
          :validate          => [
            {:type => 'required'},
            {:type => 'min-number-value', :value => 0.25, :message => _("Size must be greater than or equal to .25")},
            {:type => 'max-number-value', :value => max_available.cores, :message => _("Size must be lower than or equal to #{max_available.cores}")}
          ],
        },
        {
          :component         => 'text-field',
          :name              => 'memory',
          :id                => 'memory',
          :label             => _('Memory (GiB)'),
          :initializeOnMount => true,
          :initialValue      => hardware.memory_mb / 1024,
          :type              => 'number',
          :min               => 2,
          :max               => 807,
          :step              => 1,
          :isRequired        => true,
          :validate          => [
            {:type => 'required'},
            {:type => 'min-number-value', :value => 2, :message => _("Size must be greater than or equal to 2")},
            {:type => 'max-number-value', :value => max_available.memory, :message => _("Size must be lower than or equal to #{max_available.memory}")}
          ],

        },
        {
          :component         => 'radio',
          :name              => 'proc_type',
          :id                => 'proc_type',
          :label             => _('Core Type'),
          :initializeOnMount => true,
          :initialValue      => form_default_values.find_by(:name => 'processor_type').value,
          :isRequired        => true,
          :options           => [
            {
              :label => 'Shared uncapped',
              :value => 'shared',
            },
            {
              :label => 'Shared capped',
              :value => 'capped',
            },
            {
              :label => 'Dedicated',
              :value => 'dedicated',
            },
          ]
        },
      ],
    }
  end

  def max_available
    @max_available ||= system_pool[flavor.name.to_sym].max_available
  end

  def system_pool
    @system_pool ||= with_provider_connection(:service => 'PCloudSystemPoolsApi') do |api|
      api.pcloud_systempools_get(cloud_instance_id)
    end
  end`

agrare: @alizapeikes the issue with running API calls against the provider from the context of DDF `params_for_*` methods is related to zones and network access restrictions.

MIQ has the concept of zones and roles and customers/users can deploy servers or label nodes in such a way that access to network resources can be controlled/limited.

A real world example might be:
1. All outbound internet access is disabled by default
2. Create an "IBM Cloud Providers" Zone
3. Servers in this zone have cloud.ibm.com allowed, event, inventory, and operations roles enabled for these servers
4. API workers do not have outbound internet access

Under this deployment model (which is actually quite common) this API call will be denied.

The options are either
1. Fetch the values required as part of inventory refresh and persist in the database, this can be accessed from API workers without issue
2. Queue a task which runs this API call in the right zone with the ems_operations role and then fetch the task results.

Agree with @jaywcarman I think for now it is simpler to remove the max validation and allow it to fail (assuming the error message is helpful to the user)

alizapeikes: ok, understood. Will remove the max validation.

alizapeikes:
Merged changes into https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/420

PR #426, opened 2022-10-12
removed placement group field from vm provision workflow

alizapeikes:


jaywcarman:
Nice find @alizapeikes ! So I think #363 was originally going to add placement groups to the PowerVS VM provisioning workflow, but after the PR was created we decided to defer that piece until later (https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/363#discussion_r882910910). Unfortunately we missed this item on the dialog - giving users the impression that they can specify a placement group when in fact it does nothing.

jaywcarman (approved):
LGTM

PR #427, opened 2022-10-14
Add placement group option to powervs vm provisioning

alizapeikes:
connected to https://github.com/ManageIQ/manageiq-ui-classic/pull/8486

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
@@ -122,6 +122,12 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
     Hash[cloud_volumes || {}]
   end
 
+  def allowed_placement_groups(_options = {})
+    return {} if ar_ems.nil?
+
+    ar_ems.placement_groups.to_h { |group| [group.ems_ref, "#{group.policy}: #{group.name}"] }
+  end
+
   def set_request_values(values)
     values[:new_volumes] = parse_new_volumes_fields(values)
     super
@@ -174,6 +180,21 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisio
     end
 
     return _('IP-address field has to be either blank or a valid IPv4 address') unless valid
+  end
+
+  def validate_placement_group(_filed, _values, _dlg, _fld, value)
+    return if value.blank?
+
+    placement_group = ar_ems.placement_groups.find_by!(:ems_ref => value)
+    vms_in_placement_group = placement_group.vms
+
+    # If policy is affinity, check to make sure the new vm has the same flavor as all the members of the group
+    valid = if placement_group[:policy] == 'affinity' && vms_in_placement_group.present?
+              vms_in_placement_group.first.flavor.name == values&.dig(:sys_type, 1)
+            else
+              true
+            end
+    _('Invalid placement group - incompatible colocation policy') unless valid
   end
 
   private
--- a/content/miq_dialogs/miq_provision_ibm_powervs_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_powervs_dialogs_template.yaml
@@ -23,6 +23,15 @@
           :data_type: :string
           :min_length: 8
           :max_length: 16
+        :placement_group:
+          :values_from:
+            :method: :allowed_placement_groups
+          :description: Placement Group
+          :required: false
+          :validation_method: :validate_placement_group
+          :display: :edit
+          :auto_select_single: false
+          :data_type: :string
         :pin_policy:
           :values:
             1: "none"
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -69,6 +69,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
       specs['sys_type']      = get_option_last(:sys_type)
     end
 
+    specs['placement_group'] = get_option(:placement_group) unless get_option(:placement_group).nil?
     user_script_text = options[:user_script_text]
     user_script_text64 = Base64.encode64(user_script_text) unless user_script_text.nil?
     specs['user_data'] = user_script_text64 unless user_script_text64.nil?


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -122,6 +122,14 @@ def allowed_cloud_volumes(_options = {})
     Hash[cloud_volumes || {}]
   end
 
+  def allowed_placement_groups(_options = {})
+    return {} if ar_ems.nil?
+    
+    ar_placement_groups = ar_ems.placement_groups
+    placement_groups = ar_placement_groups&.map { |group| [group['ems_ref'], "#{group['policy']}: #{group['name']}"] }
+    Hash[placement_groups || {}]
```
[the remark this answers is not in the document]
agrare: Actually when would `ar_ems.placement_group` be `nil` ?  If it is an association it will at least be an empty assoc

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -176,6 +184,20 @@ def validate_ip_address(_field, _values, _dlg, _fld, value)
     return _('IP-address field has to be either blank or a valid IPv4 address') unless valid
   end
 
+  def validate_placement_group(_filed, _values, _dlg, _fld, value)
+    return if value.blank?
+
+    placement_group = ar_ems.placement_groups.find_by(:ems_ref => value)
```
agrare, line 190: Recommend either `.find_by!` which will raise an exception if it can't find the record or handling `placement_group` being `nil` if it isn't found by `value`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -176,6 +184,20 @@ def validate_ip_address(_field, _values, _dlg, _fld, value)
...
+    vms_in_placement_group = placement_group.vms
+
+    valid = if placement_group[:policy] == 'affinity' && vms_in_placement_group.present?
+              vms_in_placement_group.first.flavor.name == values&.dig(:sys_type, 1)
```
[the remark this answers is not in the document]
jaywcarman: Since we use flavors to track system/host type in PowerVS, IF the placement group policy is `affinity` all member VMs _must_ have the same flavor.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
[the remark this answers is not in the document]
alizapeikes: ok, so I should just remove the & and ignore the first suggestion?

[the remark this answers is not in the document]
agrare: Yeah you can pass a block to `.to_h` so even better would be:
```
ar_ems.placement_groups.to_h { |group| [group.ems_ref, "#{group.policy}: #{group.name}"] }
```

content/miq_dialogs/miq_provision_ibm_powervs_dialogs_template.yaml
```diff
@@ -17,12 +17,21 @@
       :fields:
         :ip_addr:
           :description: Specify an IPv4 address (applies to first attached network)
-          :required_method: :validate_ip_address
+          :validation_method: :validate_ip_address
```
agrare, line 20: This change looks like it is from another PR, make sure to rebase this branch on master

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -176,6 +182,21 @@ def validate_ip_address(_field, _values, _dlg, _fld, value)
...
+    placement_group = ar_ems.placement_groups.find_by!(:ems_ref => value)
+    vms_in_placement_group = placement_group.vms
+
+    # If policy is affinity, check to make sure the new vm has the same flavor as all the members of the group
+    valid = if placement_group[:policy] == 'affinity' && vms_in_placement_group.present?
+              vms_in_placement_group.first.flavor.name == values&.dig(:sys_type, 1)
+            else
+              true
+            end
+    return _('Invalid placement group - incompatible colocation policy') unless valid
```
agrare, line 197: You actually don't need this redundant `return` here as ruby automatically returns the last statement

```suggestion
    _('Invalid placement group - incompatible colocation policy') unless valid
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
```diff
@@ -122,6 +122,12 @@ def allowed_cloud_volumes(_options = {})
     Hash[cloud_volumes || {}]
   end
 
+  def allowed_placement_groups(_options = {})
+    return {} if ar_ems.nil?
+
```
agrare, line 127: Make sure to remove the trailing whitespace here

agrare:
Just a couple more items then this LGTM

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
alizapeikes: @agrare 
I will remove if you think I should, but just noting that I had originally left the white space in, even with the rubocop warning, to follow the same pattern as similar methods within this file.
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/01085d1c7c5dffd4d18932dd936a2f45a0b7ead0/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb#L57-L124

agrare: All the other methods in this file are wrong then :smile: I'm guessing they were merged while miq-bot was down since github's diff doesn't show it

alizapeikes: ok :)

agrare (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb
agrare: Oh, I think there was a miscommunication.  The issue isn't the extra newline it was the trailing whitespace.

So:
```
def allowed_instance_type(_options = {}) 
  return {} if ar_ems.nil? 

  if sap_image?
```

Is fine, but

```
def allowed_instance_type(_options = {}) 
  return {} if ar_ems.nil? 
  
  if sap_image?
```

Is the issue

agrare:
That looks better, recommend using an editor with a setting to strip trailing whitespace as that is what most of our devs use

PR #430, opened 2022-11-07
Remove newlines from public SSH keys

jaywcarman:
PowerVS allows newlines within public SSH key comments, but ManageIQ newlines are not permitted between key data.


app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -260,10 +260,11 @@ def placement_groups
   def sshkeys
     require "sshkey"
     collector.sshkeys.each do |tkey|
+      public_key = tkey.ssh_key.tr("\n", ' ')
       persister.auth_key_pairs.build(
         :name        => tkey.name,
-        :public_key  => tkey.ssh_key,
-        :fingerprint => SSHKey.sha1_fingerprint(tkey.ssh_key)
+        :public_key  => public_key,
+        :fingerprint => SSHKey.sha1_fingerprint(public_key)
```
Fryguy, line 267: @agrare Unrelated but weren't we just talking about this gem?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
```diff
@@ -260,10 +260,11 @@ def placement_groups
   def sshkeys
     require "sshkey"
     collector.sshkeys.each do |tkey|
+      public_key = tkey.ssh_key.tr("\n", ' ')
```
Fryguy, line 263: I would have expect this to be `''` as opposed to `' '`., which in turn might affect the fingerprint.  Have you verified these keys and fingerprints match what we'd see in Power VS?

jaywcarman: I didn't think comments affect the fingerprint. I purposefully replaced with `' '` to address the case where the newline immediately follows the public key. Replacing with `''` would product a key and comment with no delimiting space.

Always nice to be sure, though. The PowerVS API doesn't provide a fingerprint, so we can't check against that.

Messing around in `irb` seems to confirm that the comment has not affect on the fingerprint:
```ruby
irb(main):001:0> require "sshkey"
=> true
irb(main):002:0> public_key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQCEtw1woHfutLwXTZl+czuLg0qqJhJ0Zz+4n+SBIwhobPc6ewsX+tJw2ZI+ZYj8kObsTRqWqlC9n9XoJphHQHqnfCUSzWwbQ+y05QMfQuM41pbJuf/bzD7FNhe5ybapd5fclLEfOuZoAUex28BBdBgONhVehH8iBsC6er3wISgvC4iiFThMBFPGgE7
c5VVx3G8aAvVa5yiGx9/MSWAZb5eWSOIToegEFHpED4bla6x5nP+EA1qIxcgv771LYTGq2kma3uewsnqLr+MChPY2DaEjtHtBf9RDLpnN41aGmu3NYPg+Dog4cBT/EjYyCgTh7X1F3sREBjHUdK4X3x4hcGS/U+tUZwhINYWUwFZk3GHk42fn9Wlhc0xDwgWwZYoxPZOF5XfarYjGw2jDAzTcTnp7u2IWoPnNH23pZoy7ySJQxYN31b+YT3prL
Ud4Y4uGeA3cheWphMqYzGZS1OTQ5eSW0U2SG6eaeTK385lufOyRMbgZGVrcXYe10vf/zT/hDNc="
=> "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQCEtw1woHfutLwXTZl+czuLg0qqJhJ0Zz+4n+SBIwhobPc6ewsX+tJw2ZI+ZYj8kObsTRqWqlC9n9XoJphHQHqnfCUSzWwbQ+y05QMfQuM41pbJuf/bzD7FNhe5ybapd5fclLEfOuZoAUex28BBdBgONhVehH8iBsC6er3wISgvC4iiFThMBFPGgE7c5VVx3G8aAvVa5yiGx9/MS...
irb(main):003:0> SSHKey.sha1_fingerprint(public_key)
=> "df:71:07:2d:c1:a7:4d:62:5c:aa:4c:11:79:c7:38:83:c1:6d:92:40"
irb(main):004:0> SSHKey.sha1_fingerprint("#{public_key}\nline break!")
        7: from /home/jwcarman/.rbenv/versions/2.7.5/bin/irb:23:in `<main>'
        6: from /home/jwcarman/.rbenv/versions/2.7.5/bin/irb:23:in `load'
        5: from /home/jwcarman/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        4: from (irb):4
        3: from /home/jwcarman/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/sshkey-1.8.0/lib/sshkey.rb:88:in `sha1_fingerprint'
        2: from /home/jwcarman/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/sshkey-1.8.0/lib/sshkey.rb:149:in `decoded_key'
        1: from /home/jwcarman/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/sshkey-1.8.0/lib/sshkey.rb:157:in `parse_ssh_public_key'
SSHKey::PublicKeyError (newlines are not permitted between key data)
irb(main):005:0> SSHKey.sha1_fingerprint("#{public_key}\nline break!".tr("\n", ""))
=> "df:71:07:2d:c1:a7:4d:62:5c:aa:4c:11:79:c7:38:83:c1:6d:92:40"
irb(main):006:0> SSHKey.sha1_fingerprint("#{public_key}\nline break!".tr("\n", " "))
=> "df:71:07:2d:c1:a7:4d:62:5c:aa:4c:11:79:c7:38:83:c1:6d:92:40"
```

jaywcarman: I didn't like how I mixed single and double quotes, so I force pushed an update to fix it.

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
agrare: Actually no this is a different gem (https://github.com/heyvito/ssh-fingerprint)

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
agrare: I've actually never heard of newlines in the ssh public_key comments :shrug:  Usually I see them each on one line in an authorized_keys file.

I also confirmed anything after the public_key does not change the fingerprint.

agrare:
@jaywcarman just curious, what happens in manageiq when a key has a newline in it?

app/models/manageiq/providers/ibm_cloud/inventory/parser/power_virtual_servers.rb
Fryguy: Oh I misunderstood - I thought you were turning a multi line key into a single line.

Fryguy: The problem that concerns me with `.tr` is it's like `gsub` and will change all of the newlines.  Reading it also tells the user to expect mtultiple newlines (or at least that's how I read it, hence my comment). 

Wonder if it makes more sense to do something like `public_key.split("\n").first`

jaywcarman: > just curious, what happens in manageiq when a key has a newline in it? Does save_inventory fail or do we blow up trying to display it?

@agrare - Even though you scratched this question it got me thinking: Is the comment even used in ManageIQ? As far as I can tell it's only visible in reports:
- ✅ Report web UI works
- ✅ Report exported as csv works
- ✅ Report exported as pdf works
- ❌ Report exported as text ends up with a line break that ruins the cool text borders

3/4 isn't bad, and makes me think the best solution is to use `ssh_key.split("\n").first` to generate the fingerprint, but leave the public_key value as-is.

jaywcarman: Just force pushed what I described in the previous comment along with a much better commit message.

agrare (approved):

Fryguy:
Backported to `oparin` in commit c58584dcdc16929c2bb47393fa7e1a0913a6dfe5.

```text
commit c58584dcdc16929c2bb47393fa7e1a0913a6dfe5
Date:   Mon Nov 14 12:15:33 2022 -0500

    Merge pull request #430 from jaywcarman/remove_powervs_ssh_key_newlines
    
    Remove newlines from public SSH keys
    
    (cherry picked from commit e96bc8ee500f144a683e9d8b8e2cb4a92168999e)
```

PR #432, opened 2022-11-08
Check for SAP profiles during inventory collection 

alizapeikes:
Right now the inventory parser for powerVS vm's always uses system_type in order to update flavor, and does not distinguish between regular and SAP profiles.

PR #433, opened 2022-11-08
Remove VM resize support for SAP

alizapeikes:
Removing resize support for VM's with SAP Profile as the SAP resize options are different than regular PowerVS vm resize options.
Eventually we would want to create a special SAP form/special SAP options and support SAP resize.
Depends on #432

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
@@ -42,6 +42,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   supports :resize do
     unsupported_reason_add(:resize, _('The VM is not powered off')) unless current_state == "off"
     unsupported_reason_add(:resize, _('The VM is not connected to a provider')) unless ext_management_system
+    unsupported_reason_add(:resize, _('SAP VM resize not supported')) if flavor.kind_of?(ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile)
   end
 
   def cloud_instance_id


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/vm.rb
```diff
@@ -42,6 +42,7 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Vm < Man
   supports :resize do
     unsupported_reason_add(:resize, _('The VM is not powered off')) unless current_state == "off"
     unsupported_reason_add(:resize, _('The VM is not connected to a provider')) unless ext_management_system
+    unsupported_reason_add(:resize, _('SAP VM resize not supported')) unless flavor.type != 'ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile'
```
agrare, line 45: Prefer to use `kind_of?` rather than an exact `.type !=` check, mainly because it will still work if we subclass `ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile` with another more specific type of SAPProfile, or if we subclass the whole provider.  Also prefer the positive `if ==` instead of `unless !=`

```suggestion
    unsupported_reason_add(:resize, _('SAP VM resize not supported')) if flavor.kind_of?(ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile)
```

agrare (approved):

Fryguy:
Backported to `oparin` in commit 5f83b29c4a4f5641b015048385b673c956f2d4dd.

```text
commit 5f83b29c4a4f5641b015048385b673c956f2d4dd
Date:   Thu Nov 10 14:58:41 2022 -0500

    Merge pull request #433 from alizapeikes/sap-unsupports-resize
    
    Remove VM resize support for SAP
    
    (cherry picked from commit c34ab9effbf7312868fc83efd0572d5c2350caee)
```

PR #431, opened 2022-11-08
Add processor pools as resource pools to inventory collection

alizapeikes:
Depends on:
- https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/pull/70
- https://github.com/ManageIQ/manageiq/pull/22228

PR #435, opened 2022-11-14
Add shared processor pool field to PowerVS vm provision form

alizapeikes:
Depends on:
 - #431 
 - https://github.com/ManageIQ/manageiq-ui-classic/pull/8522


agrare (approved):
Looks good, will merge when https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/431 is merged

PR #436, opened 2022-11-14
Refactor PowerVS refresher VCR recording tasks

jaywcarman:
Refactoring the PowerVS rake task for recording the refresher spec VCR.

I attempted to use the existing Terraform based automation but ran into several issues. I believe we used Terraform to provision resources in order to end up with a '*.tfstate' inventory that could be read by the spec (to eliminate ever having to update spec assertions by hand). This is a cool idea, but in retrospec I don't think it's worth adding the overhead of maintaining Terraform code (that is completely unused anywhere else in this plugin).

This PR will remove all of the Terraform and replace it with the `ibm_cloud_power` SDK called directly in the rake tasks, prioritizing consistency with the rest of this plugin.


jaywcarman:
I'd like to ignoring the remaining miq-bot offenses.

The `puts` in the rake task are very useful when tasks take a while to complete.

The Layout/HashAlignment offenses are all for nested hashes. I think it's more readable as-is.

agrare:
@jaywcarman looks like you brought in two commits from master into this branch rather than rebasing this onto master

Also no issue with ignoring those rubocops

PR #438, opened 2022-11-20
Improve cos validation error messages

jaywcarman:


app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
```diff
@@ -182,9 +191,10 @@ def self.verify_iam(crn, api_key)
 
     begin
       resource_controller_api.get_resource_instance(:id => crn).result
-    rescue IbmCloudResourceController::ApiException => err
-      _log.error("GUID resource lookup failed: #{err.code} #{err.error}")
-      raise MiqException::MiqInvalidCredentialsError, err.error
+    rescue Exception => err
+      error_message = "CRN resource lookup failed: #{err.code} #{err.error}"
```
agrare, line 195: If we are going to rescue the more general `Exception` class then we can't count on having `:code` or `:error`
When this happens what exception is raised?

Fryguy:
If these error messages will bubble up to the user they should probably be i18n'd

app/models/manageiq/providers/ibm_cloud/object_storage/storage_manager.rb
Fryguy: Oh yeah, this is kinda dangerous - is there something more specific we can capture?

jaywcarman: Should have used `IBMCloudSdkCore::ApiException` (force pushed a fix)

agrare (approved):

PR #443, opened 2023-01-31
Processor pools support for PowerVS

miyamotoh:
Aggregated PR with 6 commits from 3 developers, adding support for "shared processor pools" in PowerVS. 
- inventory code written by @alizapeikes 
- VCR cassette updated by @jaywcarman to support the new resources
- processor pool specific code added to specs by @miyamotoh, along with another VCR cassette refresh

This supersedes #431 which has been gating;
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/435
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8522

### DIFF
--- a/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb
@@ -95,7 +95,7 @@ describe ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Refre
 
     def assert_cloud_manager
       expect(ems).to have_attributes(
-        :provider_region => "mon01"
+        :provider_region => "us-east01"
       )
     end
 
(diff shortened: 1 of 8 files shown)


spec/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/refresher_spec.rb
```diff
@@ -158,6 +158,10 @@ def assert_specific_vm
       )
 
       expect(vm.snapshots.first.total_size).to be > 0
+
+      instance_name = "test-instance-rhel-s922-shared-tier3"
```
agrare, line 162: Could we put `test-instance-ibmi-s922-capped-tier1` in a resource_pool so that we don't have to check a second vm?

miyamotoh: I tried once, but if `placement_group` and `spp_placement_group` contradict each other, trying to put the VM on this or that host, then it'd fail, and it did fail on me trying to provision a VM with both specified. I figured that'd be just asking for a trouble at a later time, kinda relying on a luck.

agrare (approved):

PR #446, opened 2023-03-08
convert console_supported? to supports

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/21990

I also got distracted and removed the `unsupported_reason_add` method. we can remove this second commit if we feel it is too much.


kbrock:
udpate:
- fixed extra paren in code (ugh)

kbrock:
- fixed paren
- fixed cop unless/else
(2 different pushes)

PR #456, opened 2023-05-23
VPC: Fix child NetworkManager and StorageManager refresh

jaywcarman:
- Add missing nested classes
- Add network and storage manager persister methods
- Add cloud_database_flavors as a _cloud_ collection
- Subclass Refresher for network and storage manager
- Delegate 'provider_region' to cloud manager

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/persister/vpc.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/persister/vpc.rb
@@ -56,7 +56,7 @@ class ManageIQ::Providers::IbmCloud::Inventory::Persister::VPC < ManageIQ::Provi
   end
 
   def add_cloud_database_flavors(extra_properties = {})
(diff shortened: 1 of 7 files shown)


agrare:
@jaywcarman I think the difference is in how you're running the refresh.  You're passing the target directly to `EmsRefresh.refresh()` where we are delegating the `#refresh` and `#refresh_queue` methods.  If you called `ManageIQ::Providers::IbmCloud::VPC::NetworkManager.first.refresh` I bet it would work.

jaywcarman:
> If you called `ManageIQ::Providers::IbmCloud::VPC::NetworkManager.first.refresh` I bet it would work.

Yes, that works, and I notice that it returns a `{ManageIQ::Providers::IbmCloud::VPC::CloudManager::Refresher=>[#<ManageIQ::Providers::IbmCloud::VPC::CloudManager:0x00000000165c9098>]`, whereas the `EmsRefresh.refresh` will use the `NetworkManager::Refresher` and return a `NetworkManager` object.

PR #457, opened 2023-05-25
Rename placement_type to the standard instance_type

agrare:
There is core code which depends on having an `:instance_type` option
for Cloud-type provision requests.

This should be merged with the schema change: https://github.com/ManageIQ/manageiq-schema/pull/692


jrafanie:
LGTM.  What do you think about a data migration?  I think this is only found in miq_requests table but I can't be sure.

jrafanie (approved):
LGTM

jrafanie:
One last question.  Is this covered by tests?

jrafanie:
> One last question. Is this covered by tests?

Offline discussion.  It's not covered by tests but not an insignificant amount of effort to add.  We'll consider doing this after merging this PR if we find a easier way to test it.

jrafanie (approved):

agrare:
This introduced an issue where the System Profile dropdown is no longer rendered by the UI, resolved by https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/461

PR #460, opened 2023-06-12
Retirement settings upon VM provision in PowerVS

miyamotoh:
Adding "Schedule" tab to Provision VM dialog for PowerVS so retirement settings can be specified. Tested with a PowerVS provider, and verified the VM provisioned with the specified retirement date already set.


jaywcarman (approved):
LGTM

PR #461, opened 2023-06-19
Fix missing System Profile dropdown VPC Provision

agrare:
The VPC Provision Form has to show the "instance_type" selection under the Hardware/Properties tab not the General tab.

I believe this was "working" before with the system type dropdown under "General" because there are other providers (vmware, rhv) which use `:provosion_type` for something else (e.g. ISO vs PXE).  This meant that the UI knew to show a dropdown under `general/provision_type` even if it was for something completely different.

When I renamed this to `instance_type` in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/457 to be in line with all other cloud providers, I didn't know that the UI wouldn't render it without it being under the `hardware` section.

Before:
image

After:
image

With this fix in place I can provision a VPC VM successfully
image

### DIFF
--- a/spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow_spec.rb
+++ b/spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow_spec.rb
@@ -39,6 +39,10 @@ describe ManageIQ::Providers::IbmCloud::VPC::CloudManager::ProvisionWorkflow do
       expect(workflow.security_group_to_security_group).to eq({})
     end
 
+    it "#allowed_instance_types" do
+      expect(workflow.allowed_instance_types).to eq({})
+    end
+
     it "#storage_type_to_profile" do
       expect(workflow.storage_type_to_profile).to eq({})
     end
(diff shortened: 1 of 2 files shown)


spec/models/manageiq/providers/ibm_cloud/vpc/cloud_manager/provision_workflow_spec.rb
agrare, line 1: NOTE these specs wouldn't have caught this issue since it was a problem with the YAML / UI but technically there weren't any specs covering `allowed_instance_types` or the previous method and we should have some.

Fryguy:
Backported to `petrosian` in commit 38aef1efc96ec5f7592c225f2daf165982e9390a.

```text
commit 38aef1efc96ec5f7592c225f2daf165982e9390a
Date:   Mon Jun 19 12:04:39 2023 -0400

    Merge pull request #461 from agrare/fix_vpc_provision_system_profile_dropdown_missing
    
    Fix missing System Profile dropdown VPC Provision
    
    (cherry picked from commit 9a3f1eb3b186cf16159b1208c88dc9c4d8f2cbfd)
```

PR #462, opened 2023-06-22
PowerVS: Attach newly created volumes during VM provision

jaywcarman:
When a user requests new volumes as part of a VM provision, the volumes
need to be created _and_ attached.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -82,7 +82,7 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
 
     attached_volumes = options[:cloud_volumes] || []
     attached_volumes.concat(phase_context[:new_volumes]).compact!
-    specs['volume_ids'] = attached_volumes unless attached_volumes.empty?
+    specs['volume_ids'] = attached_volumes.uniq unless attached_volumes.empty?
 
     attached_networks = case get_option(:vlan)
                         when 'None'
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
@@ -15,9 +15,10 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
     if new_volumes.any?
       source.with_provider_connection(:service => "PCloudVolumesApi") do |api|
         new_volumes.each do |new_volume|
-          api.pcloud_cloudinstances_volumes_post(
+          new_volume = api.pcloud_cloudinstances_volumes_post(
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
```diff
@@ -28,7 +29,7 @@ def prepare_volumes_and_networks
       source.with_provider_connection(:service => "PCloudNetworksApi") do |api|
         new_network_params = IbmCloudPower::NetworkCreate.new(:type => "pub-vlan")
         new_network = api.pcloud_networks_post(cloud_instance_id, new_network_params)
-        phase_context[:new_networks] << {"networkID" => new_network.network_id}
+        phase_context[:new_networks] << {"networkID"=> new_network.network_id}
```
agrare, line 32: Minor but this should still be ` => `

agrare (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -82,7 +82,7 @@ def prepare_for_clone
 
     attached_volumes = options[:cloud_volumes] || []
     attached_volumes.concat(phase_context[:new_volumes]).compact!
-    specs['volume_ids'] = attached_volumes unless attached_volumes.empty?
+    specs['volume_ids'] = attached_volumes.uniq unless attached_volumes.empty?
```
jaywcarman, line 85: For some reason the `prepare_for_clone` is called twice for a single provision. Before I added the `.uniq` method this would cause the newly added volume IDs to be added twice (and the provision request to PowerVS backend fails).

agrare: Interesting could you post the `caller` backtrace on each?

jaywcarman: [caller_1.txt](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/files/11895313/caller_1.txt)
[caller_2.txt](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/files/11895314/caller_2.txt)

The `prepare_for_clone_task` method gets called to set `phase_context[:clone_options]` before the `signal :start_clone_task`.

https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_provision/state_machine.rb?plain=1#L14

agrare: So I haven't checked why this was like this, but the only difference with the overridden start_clone_task method was the addition of an explicit call to `prepare_for_clone_task`:
```
-    clone_options = prepare_for_clone_task
-    log_clone_options(clone_options)
-    phase_context[:clone_task_ref] = start_clone(clone_options)
+    log_clone_options(phase_context[:clone_options])
+    phase_context[:clone_task_ref] = start_clone(phase_context[:clone_options])
```

Since you're signaling `prepare_provision` from your `prepare_volumes_and_networks` state then `prepare_for_clone_task` is already being called.

Can you try deleting the `start_clone_task` method override and see if it works as expected?

PR #469, opened 2023-08-24
Override quota calculation methods

jaywcarman:
Depends on:
https://github.com/ManageIQ/manageiq/pull/22680

PowerVS uses the more typical Flavor style of cpu allocation for SAP
profiles. For non SAP profiles, the Flavor only set the host type and
the user is allowed to request any number of cpus using the
entitled_processors form option.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
@@ -59,6 +59,15 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Template
     _log.error("image=[#{name}], error: #{e}")
   end
 
+  def number_of_cpus_for_request(request, _flavor_id = nil)
+    flavor_obj = Flavor.find_by(:ems_id => ems_id, :name => request.options[:sys_type][1])
+    if flavor_obj.kind_of?(ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::SAPProfile)
+      flavor_obj.try(:cpus)
+    else
+      request.get_option(:entitled_processors).to_f.ceil
+    end
+  end
+
   def self.raw_import_image(ext_management_system, options = {})
     session_id = SecureRandom.uuid
     wrkfl_timeout = options['timeout'].to_i.hours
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/template.rb
```diff
@@ -59,6 +59,15 @@ def raw_delete_image
     _log.error("image=[#{name}], error: #{e}")
   end
 
+  def number_of_cpus_for_request(request)
+    flavor_obj = Flavor.find_by(:name => request.options[:sys_type][1])
```
agrare, line 63: I think you want `request.get_option(:sys_type)` here.  Also we'll need to include the EMS ID so we don't find a flavor from another EMS

jaywcarman: > I think you want `request.get_option(:sys_type)` here.

That wasn't working for me, because we're not using the `flavors` table index as the index for this form.

```ruby
(byebug) request.options[:sys_type]
[1, "s922"]
(byebug) Flavor.find(request.get_option(:sys_type)).name
  Flavor Load (93.2ms)  SELECT "flavors".* FROM "flavors" WHERE "flavors"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Flavor Inst Including Associations (0.2ms - 1rows)
"e980"
(byebug) Flavor.find_by(:ems_id => ems_id, :name => "s922").name
PostgreSQLAdapter#log_after_checkout, connection_pool: size: 5, connections: 2, in use: 2, waiting_in_queue: 0
  Flavor Load (42.8ms)  SELECT "flavors".* FROM "flavors" WHERE "flavors"."ems_id" = $1 AND "flavors"."name" = $2 LIMIT $3  [["ems_id", 3], ["name", "s922"], ["LIMIT", 1]]
PostgreSQLAdapter#log_after_checkout, connection_pool: size: 5, connections: 3, in use: 3, waiting_in_queue: 0
  Flavor Inst Including Associations (0.3ms - 1rows)
  TRANSACTION (20.1ms)  BEGIN
"s922"
(byebug) Flavor.find_by(:ems_id => ems_id, :name => "s922").id
  TRANSACTION (5508.8ms)  BEGIN
  Flavor Load (13.9ms)  SELECT "flavors".* FROM "flavors" WHERE "flavors"."ems_id" = $1 AND "flavors"."name" = $2 LIMIT $3  [["ems_id", 3], ["name", "s922"], ["LIMIT", 1]]
  Flavor Inst Including Associations (0.6ms - 1rows)
2
```

This list is generated [here](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision_workflow.rb#L80C17-L80C91), where it looks like the index is just an enumeration instead of mapping to the Flavor ID.

>  Also we'll need to include the EMS ID so we don't find a flavor from another EMS

Nice catch, I'll fix this

PR #479, opened 2024-01-18
Add targeted refresh specs to PowerVS / Fix targeted VM refresh bug

jaywcarman:
I noticed duplicate entries in the `advanced_settings` table on an internal MIQ instance. @agrare suggested I add some targeted refresh specs to see if it caught the issue. Looks like there is an issue with duplicate snapshots.

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
@@ -65,6 +65,18 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers::
     []
   end
 
+  def snapshots
+    begin
+      @snapshots ||= snapshots_api.pcloud_cloudinstances_snapshots_getall(cloud_instance_id).snapshots.select do |snapshot|
+        references(:vms).include?(snapshot.pvm_instance_id)
+      end
+    rescue IbmCloudPower::ApiError => err
+      error_message = JSON.parse(err.response_body)["description"]
+      _log.debug("Error retrieving snapshots: #{error_message}")
+      nil
+    end.compact
+  end
+
   private
 
   def parse_targets!
--- a/lib/tasks_private/power_virtual_servers.rake
+++ b/lib/tasks_private/power_virtual_servers.rake
@@ -115,7 +115,7 @@ namespace :vcr do
         ),
         IbmCloudPower::PVMInstanceCreate.new(
           "server_name"     => "test-instance-ibmi-s922-capped-tier1",
-          "image_id"        => "IBMi-75-00-2984-1",
+          "image_id"        => "IBMi-75-02-2984-1",
           "sys_type"        => "s922",
           "proc_type"       => "capped",
           "storage_type"    => "tier1",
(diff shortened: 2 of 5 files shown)


app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
```diff
@@ -65,6 +65,18 @@ def sshkeys
     []
   end
 
+  def snapshots
+    @snapshots ||= references(:vms).map do |ems_ref|
+      snapshots_api.pcloud_cloudinstances_snapshots_getall(cloud_instance_id).snapshots.select do |snapshot|
+        snapshot.pvm_instance_id == ems_ref
+      end
```
agrare, line 72: @jaywcarman it looks like this is getting a list of all snapshots for each vm target, if there are >1 targets this seems inefficient.  Could we get the list of snapshots first then loop through the `references(:vms)` ?  Maybe
```suggestion
    @snapshots ||= snapshots_api.pcloud_cloudinstances_snapshots_getall(cloud_instance_id).snapshots.select do |snapshot|
      references(:vms).include?(snapshot.pvm_instance_id)
    end
```

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
```diff
@@ -65,6 +65,18 @@ def sshkeys
...
+    rescue IbmCloudPower::ApiError => err
+      error_message = JSON.parse(err.response_body)["description"]
+      _log.debug("SnapshotID not found: #{error_message}")
+      nil
+    end.flatten.compact
```
agrare, line 77: Super minor and I think we don't need this, but you can use `iterator.flat_map {}` instead of `iterator.map {}.flatten`

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
```diff
@@ -65,6 +65,18 @@ def sshkeys
...
   end
 
+  def snapshots
+    @snapshots ||= references(:vms).map do |ems_ref|
+      snapshots_api.pcloud_cloudinstances_snapshots_getall(cloud_instance_id).snapshots.select do |snapshot|
+        snapshot.pvm_instance_id == ems_ref
+      end
+    rescue IbmCloudPower::ApiError => err
+      error_message = JSON.parse(err.response_body)["description"]
+      _log.debug("SnapshotID not found: #{error_message}")
```
agrare, line 75: Would the issue here be that the snapshot_id isn't found?  Only the `cloud_instance_id` is actually passed in as a parameter.

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
jaywcarman: nice, there's always a new Ruby trick to learn 😄

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers/target_collection.rb
jaywcarman: Good catch. I updated the message to indicate it could be any error from the get snapshots api call.

agrare (approved):

Fryguy:
Backported to `quinteros` in commit a1be39427584aeb46dd66ebc0bbd105d700316b7.

```text
commit a1be39427584aeb46dd66ebc0bbd105d700316b7
Date:   Tue Jan 23 09:28:44 2024 -0500

    Merge pull request #479 from jaywcarman/powervs_targeted_vm_refresh_specs
    
    Add targeted refresh specs to PowerVS / Fix targeted VM refresh bug
    
    (cherry picked from commit 14a1b1dc33c14ed445c4afa574e07b5c33bfb718)
```

PR #481, opened 2024-01-25
[WIP] Fix duplicate advanced settings entries

jaywcarman:
Each PowerVS VM should have 3 corresponding advanced settings entries (entitled proc count, proc type and pin policy). The current refresher VCR cassette has 6 VMs so we should expect to see 18 advanced settings records.

Refreshing either the PowerVS NetworkManager or StorageManager directly adds duplicate advanced settings entries. Once the duplicate entries are in place, refreshing the CloudManager will attempt to destroy them but since we set `:read_only => true` they remain in place (even though logs say they were destroyed! See ManageIQ/inventory_refresh#120).

TODO: Figure out where the `NetworkManager` and `StorageManager` refresh scope is wrong.


jaywcarman:
Question: Why are we setting `:read_only => true` for advance settings? Is there some interface where users can change them if they're not read only?

Fryguy:
@agrare Please review.

jaywcarman:
Meant to mark this as WIP - planning to add commits to address the new spec failure.

jaywcarman:
Replaced by #483

PR #483, opened 2024-02-01
Fix duplicate advanced settings

agrare:
Built off of https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/481

We should delegate the refresh from the child managers to the parent manager.  Similar fix as https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/385

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager.rb
@@ -25,7 +25,8 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::NetworkManager < Manag
            :hostname,
            :default_endpoint,
            :endpoints,
-           :snapshots,
+           :refresh,
+           :refresh_ems,
            :to        => :parent_manager,
            :allow_nil => true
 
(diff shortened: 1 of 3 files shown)


agrare:
cc @jaywcarman

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager.rb
```diff
@@ -25,7 +25,8 @@ class ManageIQ::Providers::IbmCloud::PowerVirtualServers::NetworkManager < Manag
            :hostname,
            :default_endpoint,
            :endpoints,
-           :snapshots,
```
jaywcarman, line 28: I bet we noticed this issue back when snapshots were added, but didn't realize the scope was bigger than just snapshots.

jaywcarman (approved):

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/network_manager.rb
agrare: :+1: yeah I think you're right about that

Fryguy:
Backported to `quinteros` in commit 13ddf030c64231be402d73e1f3a9f1960f50d6d4.

```text
commit 13ddf030c64231be402d73e1f3a9f1960f50d6d4
Author: Jay Carman <jwcarman@us.ibm.com>
Date:   Thu Feb 1 11:12:26 2024 -0600

    Merge pull request #483 from agrare/fix_duplicate_advanced_settings
    
    Fix duplicate advanced settings
    
    (cherry picked from commit fbf6db240d9979a59ee2e2cf651d7406a7197578)
```

PR #509, opened 2024-11-20
[WIP] [DO NOT MERGE] Diagnose cache size

Fryguy:


Fryguy:
Closing as we've determined the cause, which is that bundler doesn't seem to git gc, and the git repositories grow continually. While all versions are affected, newer bundler versions mitigate the need since they are doing a shallow clone, so this only truly affects Ruby 3.1 with the default bundler 3.2

PR #510, opened 2024-11-21
Force bundler to the latest version on Ruby 3.1

Fryguy:
We noticed that our cache sizes were growing enormously, namely in the git repositories, where the git directory is not git gc'd. The bundler shipped with Ruby 3.1, bundler 2.3.27, exhibits the problem, whereas newer bundlers > 2.4.0 seem to have mitigated the problem because of the underlying change to how git respositories are shallow clones.

@agrare Please review.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,6 +37,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: "${{ matrix.ruby-version }}"
+        bundler: "${{ matrix.ruby-version == '3.1' && 'latest' || 'default' }}"
         bundler-cache: true
       timeout-minutes: 30
     - name: Prepare tests


.github/workflows/ci.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: "${{ matrix.ruby-version }}"
+        bundler: "${{ matrix.ruby-version == '3.1' && 'latest' || 'default' }}"
```
Fryguy, line 40: Alternatively, I could just set it to `latest` for everything, but I worry that we won't follow up with a removal when we drop 3.1.  This way we can't miss it.  Thoughts?

Fryguy:
Going to close this for now - we can deal with the caches until we drop 3.1, which will hopefully be soon.

PR #520, opened 2025-04-24
Add schedule tab with "Time until Retirement"

agrare:
This was missing and thus wasn't easy to set a time to retire the instance

image

### DIFF
--- a/content/miq_dialogs/miq_provision_ibm_vpc_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_vpc_dialogs_template.yaml
@@ -14,8 +14,61 @@
 
     :schedule:
       :description: Schedule
-      :fields: {}
-      :display: :hide
+      :fields:
+        :schedule_type:
+          :values:
+            schedule: Schedule
+            immediately: Immediately on Approval
+          :description: When to Provision
+          :required: false
+          :display: :edit
+          :default: immediately
+          :data_type: :string
+        :vm_auto_start:
+          :values:
+            false: 0
+            true: 1
+          :description: Power on virtual machines after creation
+          :required: false
+          :display: :edit
+          :default: true
+          :data_type: :boolean
+        :schedule_time:
+          :values_from:
+            :options:
+              :offset: 1.day
+            :method: :default_schedule_time
+          :description: Provision on
+          :required: false
+          :display: :edit
+          :data_type: :time
+        :retirement:
+          :values:
+            0: Indefinite
+            1.month: 1 Month
+            3.months: 3 Months
+            6.months: 6 Months
+          :description: Time until Retirement
+          :required: false
+          :display: :edit
+          :default: 0
+          :data_type: :integer
+        :retirement_warn:
+          :values_from:
+            :options:
+              :values:
+                1.week: 1 Week
+                2.weeks: 2 Weeks
+                30.days: 30 Days
+              :include_equals: false
+              :field: :retirement
+            :method: :values_less_then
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

Fryguy:
> Retirement Warn isn't showing up
Does this actually run through retirement or does it just create the UI and option, which are then later ignored?

agrare:
> Does this actually run through retirement or does it just create the UI and option, which are then later ignored? I saw the TODO, but wasn't sure if that was the same as what I'm asking

The retirement warn is another UI dropdown where the user can select when they are warned that the vm will be retired.  I wonder if there is another issue with that though because locally I don't see it for VMware either, where looking at a released version we did see it.

I haven't tested VPC retirement after setting this, it should work like any other VM retirement at that point since it isn't the provision dialogs doing the retirement (good to test that there isn't something else also needed though)

agrare:
Okay there is something else going on with the retirement_warn dropdown, I don't see it on the VMware schedule tab on master locally, on a -devel appliance, and on a radjabov-1 appliance so I don't want to hold this up for that.

agrare:
Ah I'm an idiot, I didn't realize the retirement warning only shows up if you select a ` Time until Retirement ` that isn't indefinite.

agrare:
Okay I ordered a VPC VM via the service catalog, and selected 1 month retires_on and 1 week retirement_warn,
```
>> vm.retires_on
=> Sun, 01 Jun 2025 02:37:20.367762000 UTC +00:00
>> vm.retirement_warn
=> 7
```
I did a "Retire this VM" to test that the retire process works for VPC.  I got an error but I'm not sure if it is because of VPC or just retire in general:
```
[----] I, [2025-05-01T12:16:36.462750#893853:8a34]  INFO -- evm: MIQ(MiqQueue#deliver) Message id: [8546], Delivering...
[----] I, [2025-05-01T12:16:36.481554#893853:8a34]  INFO -- evm: Delivering {"request"=>"vm_retire", "VmOrTemplate::vm"=>1546} for object [VmRetireTask.40] with state [CheckRemovedFromProvider] to Automate
[----] I, [2025-05-01T12:16:36.488341#893853:8a34]  INFO -- evm: MIQ(VmRetireTask#after_ae_delivery) ae_result="error"
[----] E, [2025-05-01T12:16:36.516775#893853:8a34] ERROR -- evm: MIQ(MiqQueue#deliver) Message id: [8546], Error: [Tried to load unspecified class: MiqAeEngine::StateVarHash]
[----] E, [2025-05-01T12:16:36.517030#893853:8a34] ERROR -- evm: [Psych::DisallowedClass]: Tried to load unspecified class: MiqAeEngine::StateVarHash  Method:[block (2 levels) in <class:LogProxy>]
[----] E, [2025-05-01T12:16:36.517205#893853:8a34] ERROR -- evm: /home/grare/adam/.gem/ruby/3.3.0/gems/psych-4.0.6/lib/psych/class_loader.rb:99:in `find'
/home/grare/adam/.gem/ruby/3.3.0/gems/psych-4.0.6/lib/psych/class_loader.rb:28:in `load'
/home/grare/adam/.gem/ruby/3.3.0/gems/psych-4.0.6/lib/psych/visitors/to_ruby.rb:426:in `resolve_cl
```

Going to try to retire a VMware VM and see if I get the same issue

agrare:
Retire VM bug fixed by https://github.com/ManageIQ/manageiq-automation_engine/pull/572

agrare:
And a successful VPC VM retire after that PR is applied locally
image

PR #538, opened 2025-07-01
Switch to use VcrSecrets helper from core

Fryguy:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23497


spec/spec_helper.rb
```diff
@@ -68,19 +68,8 @@ def header_sanitizer(response_header, transient_headers)
   config.ignore_hosts('codeclimate.com') if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
 
-  config.before_record do |i|
-    replace_token_contents(i) if i.request.uri == "https://iam.cloud.ibm.com/identity/token"
-    vpc_sanitizer(i) if i.request.uri.match?('iaas.cloud.ibm') || i.request.uri.match?('tags.global-search-tagging')
-  end
```
Fryguy, line 74: The tests passed without this, so I'm not sure why it's needed, unless it's during recording?

spec/spec_helper.rb
```diff
@@ -59,28 +59,18 @@ def header_sanitizer(response_header, transient_headers)
...
 VCR.configure do |config|
-  # config.debug_logger = $stdout # Keep for debugging tests.
-
-  # Configure VCR to use rspec metadata.
-  config.hook_into(:webmock)
-  config.configure_rspec_metadata!
-
   config.ignore_hosts('codeclimate.com') if ENV['CI']
   config.cassette_library_dir = File.join(ManageIQ::Providers::IbmCloud::Engine.root, 'spec/vcr_cassettes')
+  config.hook_into(:webmock)
```
Fryguy, line 64: This is defined by core, so it's possible we don't need this.

jrafanie (approved):

agrare (approved):

spec/spec_helper.rb
agrare: Yeah we can probably drop it, can merge as-is since it isn't hurting anything

PR #557, opened 2026-04-28
Include stock images

shashank-gowda:
When populating image details, include stock images as well, but ignore any images marked as stock within the custom image list.

@miq-bot add-label bug

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -41,7 +41,13 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
   end
 
   def images_by_id
-    @images_by_id ||= images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images.index_by(&:image_id)
+    @images_by_id ||= begin
+      custom_images = images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images
+      stock_images = images_api.pcloud_cloudinstances_stockimages_getall(cloud_instance_id).images
+      # Filter out custom images where imageType contains 'stock'
+      filtered_custom_images = custom_images.reject { |image| image.specifications&.image_type&.include?('stock') }
+      (filtered_custom_images + stock_images).index_by(&:image_id)
+    end
   end
 
   def images
--- a/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
+++ b/app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
@@ -45,8 +45,8 @@ class ManageIQ::Providers::IbmCloud::Inventory::Collector::PowerVirtualServers <
       custom_images = images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images
       stock_images = images_api.pcloud_cloudinstances_stockimages_getall(cloud_instance_id).images
       # Filter out custom images where imageType contains 'stock'
-      filtered_custom_images = custom_images.reject { |image| image.specifications&.image_type&.include?('stock') }
-      (filtered_custom_images + stock_images).index_by(&:image_id)
+      custom_images.reject! { |image| image.specifications&.image_type&.include?('stock') }
+      (custom_images + stock_images).index_by(&:image_id)
     end
   end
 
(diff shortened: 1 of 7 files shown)


agrare:
@shashank-gowda can anything about an image be changed if it is in the custom images list?  Could you copy a "stock" image but change something about it?  How does a stock image make it to the custom images list?

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
```diff
@@ -41,7 +41,13 @@ def image(image_id)
   end
 
   def images_by_id
-    @images_by_id ||= images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images.index_by(&:image_id)
+    @images_by_id ||= begin
+      custom_images = images_api.pcloud_cloudinstances_images_getall(cloud_instance_id).images
+      stock_images = images_api.pcloud_cloudinstances_stockimages_getall(cloud_instance_id).images
+      # Filter out custom images where imageType contains 'stock'
+      filtered_custom_images = custom_images.reject { |image| image.specifications&.image_type&.include?('stock') }
```
agrare: You can use `reject!` to apply the results of the block to the current variable, saves introducing a new local variable when the old one has data that we don't care about

```suggestion
      custom_images.reject! { |image| image.specifications&.image_type&.include?('stock') }
```

agrare: > `image.specifications&.image_type&.include?('stock')`

Is there a list of image types?  I'm wondering if it would be better to use a regular expression `match?` with more concrete rules for what a stock image is.  (NOTE from the current VCR it looks like they all are `"imageType":"stock"`)

And just to confirm this can't be set by the user correct?

agrare:
@shashank-gowda with the new API call we'll need to re-record the VCR cassette

app/models/manageiq/providers/ibm_cloud/inventory/collector/power_virtual_servers.rb
Fryguy: > You can use `reject!` to apply the results of the block to the current variable, saves introducing a new local variable when the old one has data that we don't care about

As long as `.images` isn't a cached copy inside the images_api

Fryguy:
Can you describe what stock images are? I can't understand how an image can be both "custom" and "stock" based on the casual definition of stock.

shashank-gowda:
@Fryguy @agrare 
Stock images are built-in images. Custom images include imported images, as well as images used when VMs are created using stock images.

Fryguy:
Do you mean a customer can take a stock image, deploy it, modify it, then create an image from that VM? Is that both "stock" and "custom"?

agrare:
@shashank-gowda just to confirm, do you mean:
1. Initially the custom images list is empty
2. A customer can upload their own images to custom images
3. If a customer deploys a vm from a stock image, that stock image gets added to the custom images list? (weird but okay)

shashank-gowda:
@agrare 

Yes, if a VM is deployed from a stock image, that image is also added to the custom image list.

Now the issue is, When a new PowerVS workspace without any custom images is added to ManageIQ, the image list is empty, so users cannot deploy VMs. With this fix user will be able to deploy the VM using the stock images.

Fryguy:
Ok that is very strange, but it is what it is. As such, I'm good with this PR.

However, this leads to a couple more questions.

Is there an image id in the VM metadata referring to which image it came from? If so, which image does it point to?

Asking because we use this for VM genealogy (i.e. understanding/displaying the VM<->Image history). My concern is this PR throws away the custom image as opposed to associating it with the corresponding stock image and doing a genealogy tree of `stock-image -> custom-stock-image -> VM`

agrare:
@shashank-gowda if a vm is provisioned from a stock image and subsequently deleted, is that "stock" image in the custom images list also deleted or do all images stay indefinitely if they have ever been used?
Can a user delete a stock image from the custom images list, even if it is currently in use by an active vm?

shashank-gowda:
@agrare 

If a vm is provisioned from a stock image and subsequently deleted, is that "stock" image in the custom images list also deleted or do all images stay indefinitely if they have ever been used?

--- When VM is deleted, the "stock" image in the custom images list will not be deleted it stays.

Can a user delete a stock image from the custom images list, even if it is currently in use by an active vm?

--- Yes, the stock image can be deleted from the custom images list, even when it is currently in use by an active vm.

agrare:
>  Yes, the stock image can be deleted from the custom images list, even when it is currently in use by an active vm.

Interesting I wonder then to @Fryguy's point about geneology if we link vms that point to "custom" stock images to the actual stock image instead so it can't be deleted by the end user

Fryguy:
Yeah I don't know if we have genealogy currently (if not, we should, because it's easy enough to add). I just don't want to paint ourselves into a corner where the images we need no longer exist.

Fryguy:
> Merge commit https://github.com/ManageIQ/manageiq-providers-ibm_cloud/commit/fc6e1bb6a9ce868c14cec24f77894221b75f804b detected. Consider rebasing.

@shashank-gowda Please use rebase instead of merge when pull in upstream master.

agrare:
@shashank-gowda this looks like it has gotten left behind, what do you want to do here?

PR #568, opened 2026-07-08
Bump ibm_cloud_iam to v2.0

agrare:
https://github.com/IBM-Cloud/ibm-cloud-sdk-ruby/pull/86


app/models/manageiq/providers/ibm_cloud/container_manager.rb
```diff
@@ -35,7 +35,7 @@ def connect_options(options = {})
 
   def self.kubernetes_auth_options(options)
     {
-      :bearer_token => get_token(options[:api_key]).id_token
+      :bearer_token => get_token(options[:api_key]).access_token
```
agrare, line 38: I want to do a live test of this to ensure this is still accurate.

agrare:
@Fryguy can you review?

PR #572, opened 2026-08-11
add support for Power VS multi-vm provisioning

Nikhil-Desai-IBM:
Add support for multi-vm provisioning, a capability native to Power VS. This is done using the **replicants** field.
As the **number_of_vms** goes beyond 1, its value is copied into the hidden **replicants** field and the former is reset to 1 to avoid core fan-out of ProvisionTasks.
Visibility of certain fields that are no longer relevant are toggled off while other fields that are, are rendered.

Related:
* https://github.com/ManageIQ/manageiq-ui-classic/pull/10227

### DIFF
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -73,8 +73,13 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
       specs['sys_type']      = get_option_last(:sys_type)
     end
 
-    specs['placement_group'] = get_option(:placement_group) unless get_option(:placement_group).nil?
-    specs['shared_processor_pool'] = get_option(:shared_processor_pool) unless get_option(:shared_processor_pool).nil?
+    if get_option_last(:number_of_vms).to_i > 1
+      policy = get_option(:replicant_affinity_policy)
+      specs['placementGroupPolicy'] = policy if policy.present? && policy != 'none'
+    else
+      specs['placement_group'] = get_option(:placement_group) unless get_option(:placement_group).nil?
+      specs['shared_processor_pool'] = get_option(:shared_processor_pool) unless get_option(:shared_processor_pool).nil?
+    end
     user_script_text = options[:user_script_text]
     user_script_text64 = Base64.encode64(user_script_text) unless user_script_text.nil?
     specs['user_data'] = user_script_text64 unless user_script_text64.nil?
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -169,13 +169,14 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
 
       phase_context[:cloud_api_completion_time] = Time.zone.now.utc if all_done
 
-      if building.any?
-        status = "#{building.length} of #{ids.length} instance(s) still provisioning."
-      elsif all_done
-        status = "All #{ids.length} instance(s) provisioned and active."
-      else
-        status = "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
-      end
+      status = if building.any?
+                "#{building.length} of #{ids.length} instance(s) still provisioning."
+               elsif all_done
+                "All #{ids.length} instance(s) provisioned and active."
+               else
+                _log.warn("Unknown server state received from the cloud API: '#{instance_state}'")
+                "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
+               end
 
       return all_done, status
     end
--- a/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
@@ -50,6 +50,8 @@ module ManageIQ::Providers::IbmCloud::PowerVirtualServers::CloudManager::Provisi
   end
 
   def prepare_for_clone
+    replicants = get_option(:replicants)
+
     specs = {
       'image_id'   => get_option_last(:src_vm_id),
       'pin_policy' => get_option_last(:pin_policy),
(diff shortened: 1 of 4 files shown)


Nikhil-Desai-IBM:
Please refer the API endpoint spec here https://cloud.ibm.com/docs/apis/power-cloud#pcloud-pvminstances-post
Also attached are the native Power VS UI screenshots.
Screenshot 2026-08-10 at 9 34 56 PM
Screenshot 2026-08-10 at 9 35 16 PM

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -120,7 +130,8 @@ def make_request_clone(clone_options)
     source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
       body = IbmCloudPower::PVMInstanceCreate.new(clone_options)
       response = api.pcloud_pvminstances_post(cloud_instance_id, body)
-      response&.first&.pvm_instance_id
+      ids = response&.map(&:pvm_instance_id)&.compact
+      ids.length == 1 ? ids.first : ids
```
agrare: If you're going to `Array()` this result anyway [[ref]](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/572/changes#diff-ad8b9c47e7a1a709c3685c42f1771ff4169a4fd28bae7e2ec38c4f1e40577a3eR149) what if you just always return an array of IDs here even if it is an array of 1?

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/state_machine.rb
```diff
@@ -9,6 +9,46 @@ def create_destination
...
+
+      if manager.allow_targeted_refresh?
+        ids.each do |id|
+          target = InventoryRefresh::Target.new(
+            :manager     => manager,
+            :association => :vms,
+            :manager_ref => {:ems_ref => id}
+          )
+          EmsRefresh.queue_refresh(target)
+        end
```
agrare: It is more efficient to queue a single refresh with multiple targets.  You can build up a `TargetCollection`:

(Untested)
```suggestion
        target_collection = InventoryRefresh::TargetCollection.new(:manager => manager)
        ids.each { |id| target_collection.add_target(:association => :vms, :manager_ref => {:ems_ref => id}) }
        EmsRefresh.queue_refresh(target_collection)
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
Nikhil-Desai-IBM: you're right! I had kept it for the sake of convention. Homogenized the return type now.

Nikhil-Desai-IBM: Thanks @agrare for taking the time to review it!
I have incorporated both of your suggestions. I wasn't aware of the TargetCollection.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -135,26 +145,39 @@ def check_task_clone_to_template(clone_task_ref)
...
-      when 'ACTIVE'
-        stop = (instance.processors.to_f > 0) && (instance.memory.to_f > 0)
-        phase_context[:cloud_api_completion_time] = Time.zone.now.utc if stop
-        status = "The server has been provisioned.; #{stop ? 'Server description available.' : 'Waiting for server description.'}"
-      when 'ERROR'
-        raise MiqException::MiqProvisionError, _("An error occurred while provisioning the instance.")
+      statuses = ids.map do |id|
+        begin
+          instance = api.pcloud_pvminstances_get(cloud_instance_id, id)
+          [id, instance.status, instance.processors.to_f, instance.memory.to_f]
```
agrare: Really minor, might be something you can do in a refactoring follow-up PR but it seems like returning the stats, procs, and memory to just check `state == ACTIVE && cpus > 0 && mem > 0` later could be simplified.  Maybe even build up a hash like `{"ERROR" => [], "BUILD" => [], "ACTIVE" => []}` so make the process of checking below simpler.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -135,26 +145,39 @@ def check_task_clone_to_template(clone_task_ref)
...
+
+      if building.any?
+        status = "#{building.length} of #{ids.length} instance(s) still provisioning."
+      elsif all_done
+        status = "All #{ids.length} instance(s) provisioned and active."
       else
-        status = "Unknown server state received from the cloud API: '#{instance_state}'"
-        _log.warn(status)
+        status = "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
       end
```
agrare: I'm actually surprised miq_bot/rubocop didn't warn about this (I think the bot might be down), but if you're setting a local variable in every branch of a conditional it is preferred to do:

```suggestion
      status = if building.any?
                 "#{building.length} of #{ids.length} instance(s) still provisioning."
               elsif all_done
                 "All #{ids.length} instance(s) provisioned and active."
               else
                 _log.warn("Unknown server state received from the cloud API: '#{instance_state}'")
                 "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
               end
```

agrare:
Overall this looks good to me just a minor nit-pick

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
Nikhil-Desai-IBM: @agrare, agree; status is now being set using the if expression idiom.

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -135,26 +145,40 @@ def check_task_clone_to_template(clone_task_ref)
...
+      phase_context[:cloud_api_completion_time] = Time.zone.now.utc if all_done
+
+      status = if building.any?
+                "#{building.length} of #{ids.length} instance(s) still provisioning."
+               elsif all_done
+                "All #{ids.length} instance(s) provisioned and active."
+               else
+                _log.warn("Unknown server state received from the cloud API: '#{instance_state}'")
+                "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
+               end
```
agrare, line 176:
```suggestion
      status = if building.any?
                 "#{building.length} of #{ids.length} instance(s) still provisioning."
               elsif all_done
                 "All #{ids.length} instance(s) provisioned and active."
               else
                 _log.warn("Unknown server state received from the cloud API: '#{instance_state}'")
                 "#{active.length} of #{ids.length} instance(s) active, waiting for full description."
               end
```

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -120,7 +130,7 @@ def make_request_clone(clone_options)
     source.with_provider_connection(:service => "PCloudPVMInstancesApi") do |api|
       body = IbmCloudPower::PVMInstanceCreate.new(clone_options)
       response = api.pcloud_pvminstances_post(cloud_instance_id, body)
-      response&.first&.pvm_instance_id
+      response&.map(&:pvm_instance_id)&.compact
```
agrare: rubocop is right `response&.filter_map` is better than `.map.compact`

app/models/manageiq/providers/ibm_cloud/power_virtual_servers/cloud_manager/provision/cloning.rb
```diff
@@ -135,26 +145,40 @@ def check_task_clone_to_template(clone_task_ref)
...
+        begin
+          instance = api.pcloud_pvminstances_get(cloud_instance_id, id)
+          [id, instance.status, instance.processors.to_f, instance.memory.to_f]
+        rescue IbmCloudPower::ApiError => e
+          # The instance may not be queryable immediately after creation (transient
+          # 500s are common in the first few seconds).  Treat it as still building.
+          _log.warn("Transient error polling instance #{id}: #{e.message}. Will retry.")
+          [id, 'BUILD', 0, 0]
+        end
       end
```
agrare, line 159: Try without the explicit begin/end if you're going to wrap an entire block in a rescue:
```suggestion
      statuses = ids.map do |id|
        instance = api.pcloud_pvminstances_get(cloud_instance_id, id)
        [id, instance.status, instance.processors.to_f, instance.memory.to_f]
      rescue IbmCloudPower::ApiError => e
        # The instance may not be queryable immediately after creation (transient
        # 500s are common in the first few seconds).  Treat it as still building.
        _log.warn("Transient error polling instance #{id}: #{e.message}. Will retry.")
        [id, 'BUILD', 0, 0]
     end
```

agrare (approved):

